PR #6, opened 2017-09-25
Minor fix in the installation snippet

peteruhnak:


btamayo:
Thanks! Good catch :)

PR #17, opened 2017-10-02
Transfer bats#163 - Avoid undocumented usage of the head utility.

agent-0028:
all instances of head appear to be using -n variant now. :thumbsup:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2,7 +2,7 @@
 set -e
 
 resolve_link() {
-  $(type -p greadlink readlink | head -1) "$1"
+  $(type -p greadlink readlink | head -n1) "$1"
 }
 
 abs_dirname() {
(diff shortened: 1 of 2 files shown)


mbland:
Note that `resolve_link()` in `libexec/bats` has been rewritten to eliminate the command substitution (part of #8). You could copy that version into `install.sh` to resolve the issue.

jasonkarns:
Closes sstephenson/bats#163

jasonkarns:
I think it should be safe to leave the "simplified" readlink code in the install script, since there really won't be any performance concerns for the install script. (There's value in having the implementations be consistent but meh?)

mbland:
Yeah, that's what I was thinking when I went ahead and merged it, that the install script doesn't need to be optimized per se.

PR #19, opened 2017-10-02
Transfer bats#142 - This commit corrects the problem of incorrect TAP format for skipped tests

agent-0028:
todo
- [x] validate the skip reason is included in the formatted "spec" output

This transfers sstephenson/bats#142 to bats-core and makes "skipped" test output compliant to TAP spec

### DIFF
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -57,8 +57,8 @@ fixtures bats
   [ $status -eq 0 ]
   [ "${lines[0]}" = "1..3" ]
   [ "${lines[1]}" = "ok 1 a passing test" ]
-  [ "${lines[2]}" = "ok 2 a skipping test # skip" ]
-  [ "${lines[3]}" = "ok 3 skip test with a reason # skip for a really good reason" ]
+  [ "${lines[2]}" = "ok 2 a skipped test with no reason # skip" ]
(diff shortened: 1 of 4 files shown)


test/bats.bats
```diff
@@ -49,7 +49,16 @@ fixtures bats
 @test "summary passing and skipping tests" {
   run filter_control_sequences bats -p $FIXTURE_ROOT/passing_and_skipping.bats
   [ $status -eq 0 ]
-  [ "${lines[2]}" = "2 tests, 0 failures, 1 skipped" ]
+  [ "${lines[3]}" = "3 tests, 0 failures, 2 skipped" ]
+}
+
+@test "tap passing and skipping tests" {
+  run filter_control_sequences bats --tap $FIXTURE_ROOT/passing_and_skipping.bats
```
mbland: Add double quotes around the `$FIXTURE_ROOT/passing_and-skipping.bats` path here and on line 77. Otherwise a `$FIXTURE_ROOT` with spaces in it will break the test. See ebb192e036284b5006ecac08343c72be6e93e19b from #8.

jasonkarns: @agent-0028 You good to make this change?

agent-0028: Indeed. I'll take care of it.

PR #21, opened 2017-10-03
test/bats: Add has_tput helper to fix Alpine tests

mbland:
The test cases that use the `-p` or `--pretty` flag would fail if `tput` wasn't available (installed as part of the `ncurses` package). On Alpine, `ncurses` isn't installed by default. This change ensures the tests validate the expected behavior regardless of whether `tput` is available.

Alternatives might be to fail the whole test suite and insist the user install `ncurses` first, or to `skip` tests using `-p` or `--pretty` when `tput` is missing.

cc: @jasonkarns

### DIFF
--- a/test/test_helper.bash
+++ b/test/test_helper.bash
@@ -11,6 +11,13 @@ filter_control_sequences() {
   "$@" | sed $'s,\x1b\\[[0-9;]*[a-zA-Z],,g'
 }
 
+if ! command -v tput >/dev/null; then
+  tput() {
+    printf '1000\n'
+  }
+  export -f tput
+fi
+
 teardown() {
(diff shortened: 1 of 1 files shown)


jasonkarns:
I like the idea of just skipping the tests. Would honestly like to move all the --pretty tests into a separate test file dedicated to validating the --pretty formatter.

At some point in the future, we could also leverage bats-mock to mock tput when it doesn't exist and the tests would no longer need to be skipped.

mbland:
@jasonkarns Your comment gave me a better idea: Now `test_helper` will define a fake `tput` function if the command is missing.

btamayo:
Merged #21.

Issue #20, opened 2017-10-02
1.0 bug fixes


jasonkarns:
(transferring bugs list from sstephenson/bats#196)

- [ ] Empty output lines are missing from `${lines[@]}`.
    sstephenson/bats#89
    sstephenson/bats#90
    sstephenson/bats#92

    sstephenson/bats#90 has already been merged to master
- [x] Use POSIX compliant options with head for better compatibility (e.g. Busybox support).
    sstephenson/bats#163

    closed by #17
- [x] Incorrect TAP format for skipped tests.
    sstephenson/bats#141
    sstephenson/bats#142

    will be closed by #19 
- [x] `set -o nounset` breaks `bats-exec-test`.
    sstephenson/bats#171
- [x] `bats-preprocess` breaks when translating a `@test` line that uses tabs for white space, instead of plain spaces.
    sstephenson/bats#98
    sstephenson/bats#122
- [x] Bats failing on OS X 10.11.2.
    sstephenson/bats#140

mbland:
Corresponding to each of the above:

* Actually, empty `output` lines _are_ still missing from `lines`. I've got a [split_bats_output_into_lines helper](https://github.com/mbland/go-script-bash/blob/v1.6.0/lib/bats/helpers#L282-L291) (and its [__split_bats_output_into_lines helper](https://github.com/mbland/go-script-bash/blob/v1.6.0/lib/bats/helpers#L415-L423) for speed reasons) that would be trivial to apply—but it would be a breaking change for a lot of tests. I'd vote for that to go into v1.0, with a corresponding `bats_strip_blank_lines` helper or other mechanism to help people migrate from the old behavior to the new.

* I've noticed test breakages on Alpine, which I believe would fall into the Busybox compatibility bucket, although #17 did resolve the one issue. I'll try taking a look at them momentarily.

* I'll be reviewing #19 shortly.

* Haven't looked at the `set -o nounset` issue closely yet. Some definite subtleties there.

* The whitespace solution from sstephenson/bats#122 should now apply to the definition of `BATS_TEST_PATTERN` in `libexec/bats`, but otherwise should just work.

[...]

PR #23, opened 2017-10-07
Support tabs in @test definition

mbland:
Part of #20. The base of this is ported from sstephenson/bats#122 by @BenoitLefebvre to address sstephenson/bats#98. Added a couple of refinements on top of that original.

One of those refinements is to emit a warning if a test name is empty. If folks would rather, I can split that commit off into another PR.


test/fixtures/bats/empty_name.bats
```diff
@@ -0,0 +1,3 @@
+@test  { true; }
+@test '' { true; }
+@test '	 	 	 ' { true; }
```
sublimino: Perhaps move "tabs and spaces" to individual tests and annotate/comment them for clarity?

test/fixtures/bats/single_line.bats
```diff
@@ -1,9 +1,9 @@
-@test "empty" { }
+@test "empty" {}
 
-@test "passing" { true; }
+@test "passing" {true;}
 
-@test "input redirection" { diff - <( echo hello ); } <<EOS
```
sublimino: Any reason for removing spaces here and not in `empty_name.bats`? I prefer the readability of spaced, but it's another styleguide question to which only conformity is important!

mbland:
FYI @sublimino points taken; will address later this evening.

test/fixtures/bats/empty_name.bats
jasonkarns: If I can speak for @mbland here, I think this test is still just ensuring that a whitespace-only test name is still treated as "empty", so I think it still belongs in this set of fixtures.

test/fixtures/bats/single_line.bats
jasonkarns: Agree here. I'd rather the `tabs.bats` be changed to `whitespace.bats` and cover all the whitespace variants:
- no whitespace around single-line
- spaces around single-line
- tabs around single-line

- tabs before @test 
- tabs before test name
- tabs before test body
- spaces before @test 
- spaces before test name
- spaces before test body

jasonkarns:
@mbland I do think the warning might be better off in a separate PR. reason being... I'm beginning to see a theme around bats having its own "strict mode" so to speak. consider:

- warning about empty test names
- warning about duplicate test names
- bats itself being in or out-of UnofficialBashStrictMode
- bats "fixing" known bash gotchas (like `[[`)

Perhaps we should have a more comprehensive flag that enables bats "strict" mode (or maybe even better, defaults to strict mode and offer a --loose alternative). In this theoretical strict mode, empty and duplicate test names automatically fail; bats itself is placed in strict mode before the execution of every test (to ensure the sourced test doesn't disable it); and we automatically fix any bash gotchas that we know about?

test/fixtures/bats/single_line.bats
mbland: I've renamed the fixture to `whitespace.bats` and added a few extra cases, but didn't add extra cases for tabs vs. spaces, since the `[:space:]` character class matches both.

What _did_ happen is the new "extra whitespace around single-line test" case exposed a bug where the extra spaces after the closing quote were getting matched as part of the test case name. As a result, Bash would exit with a parse error when executing the preprocessed file, since the test name included the closing quote and its trailing leading to an unterminated quote parse error.

I've since updated `BATS_TEST_PATTERN` so that the name/description capture matches a non-`[:space:]` character last instead of first.

mbland:
@jasonkarns and @sublimino I've rebased the branch, renamed `tabs.bats` to `whitespace.bats`, added new test cases, fixed a bug by updating `BATS_TEST_PATTERN` (see my reply to the review comment from earlier, or the commit message for a18c8f87e693ce079a9e40f49a22a15cbef52164), and peeled off the empty-name check for a future PR. PTAL.

Also, I like the idea of enabling a strict mode by default with a `--loose` or `BATS_DISABLE_STRICT_MODE` option for v1.0.0.

libexec/bats
```diff
@@ -77,7 +77,7 @@ abs_dirname '.' 'BATS_CWD'
 
 export BATS_PREFIX
 export BATS_CWD
-export BATS_TEST_PATTERN='^ *@test +(.+) +\{ *(.*)$'
+export BATS_TEST_PATTERN="^[[:space:]]*@test[[:space:]]+(.*[^[:space:]])[[:space:]]+\{(.*)\$"
```
jasonkarns: Aren't the extra wrapping brackets around `[:space:]` unnecessary? (Except in the single negation instance?)

jasonkarns: Apparently they _are_ required. TIL (I thought they operated just like the shortcut forms.)

mbland: They're all required. Just tried locally to make sure, but the outside brackets just announce a set of characters to match, and `[:space:]` is special syntax announcing a character class within those outer brackets.

mbland:
@jasonkarns TIL about `[:blank:]`. How'd I miss that all these years? I've pushed a new commit with that change. PTAL?

Also, I don't think the Bash version would have an impact on the complexity of the regex at all. At least, off the top of my head, I can't think of why it would. Have you ideas?

jasonkarns:
@mbland 

> How'd I miss that all these years?

No judgement here. I just learned this morning when I was reading about the double brackets.



> complexity of the regex

I was mainly just thinking of character class shorthands, but I guess those aren't even in ERE...

mbland:
@jasonkarns Yeah, I think all the regex parsing is done by the standard C lib.

Just waiting for Travis to clear its macOS backlog before merging...

PR #24, opened 2017-10-07
Emit ` || return 1` suffix for double-bracket tests

### LINKED ISSUE #18: Workaround for [[ not triggering test failure on some Bash versions
Corresponds to sstephenson/bats#49. I believe we can work around this during the preprocessing step by matching lines against `\]\][[:space:]]*$` and appending ` || false` to them, at least when `[[ "${BASH_VERSION:0:1}" -lt '4' ]]`. But of course I'd need to test it!

mbland:
Closes #18. Note: I had originally used ` || false`, but then realized ` || return 1` was even better.

Corresponds to sstephenson/bats#49. For a negligible performance cost, assertions using `[[ ... ]]` will now behave as expected even on Bash versions prior to 4.1-alpha. Most notably, they will work with Bash 3.2.57(1)-release that ships with macOS.

The nice thing is, the way `bats_print_failed_command` already works, the `|| return 1` suffix doesn't appear in the failure output. Also, since the cost is negligible, there's no need to make the suffix insertion dependent on a version check.

Tested using both 3.2.57(1)-release and Bash 4.4.12(1)-release.

### DIFF
--- a/libexec/bats-preprocess
+++ b/libexec/bats-preprocess
@@ -45,7 +45,7 @@ while IFS= read -r line; do
     tests["${#tests[@]}"]="$encoded_name"
     echo "${encoded_name}() { bats_test_begin \"${name}\" ${index}; ${body}"
   elif [[ "$line" =~ \]\][[:space:]]*$ ]]; then
-    printf "%s || false\n" "$line"
+    printf "%s || return 1\n" "$line"
   else
     printf "%s\n" "$line"
   fi
--- a/libexec/bats-preprocess
+++ b/libexec/bats-preprocess
@@ -44,6 +44,8 @@ while IFS= read -r line; do
     encode_name "$name" 'encoded_name'
     tests["${#tests[@]}"]="$encoded_name"
     echo "${encoded_name}() { bats_test_begin \"${name}\" ${index}; ${body}"
+  elif [[ "$line" =~ \]\][[:space:]]*$ ]]; then
+    printf "%s || false\n" "$line"
   else
     printf "%s\n" "$line"
   fi
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -338,3 +338,11 @@ END_OF_ERR_MSG
     emit_debug_output && return 1
   fi
 }
+
+@test "ensure double bracket expressions fail as necessary" {
+  run bats "$FIXTURE_ROOT/double_brackets.bats"
+  [ "$status" -eq '1' ]
+  [ "${lines[1]}" == 'not ok 1 double bracket returns false' ]
+  [ "${lines[2]}" == "# (in test file $RELATIVE_FIXTURE_ROOT/double_brackets.bats, line 4)" ]
+  [ "${lines[3]}" == "#   \`[[ \"\$value\" == 'false' ]]' failed" ]
+}
--- a/test/fixtures/bats/double_brackets.bats
+++ b/test/fixtures/bats/double_brackets.bats
@@ -1,0 +1,5 @@
+@test "double bracket returns false" {
+  local value='true'
+  [[ "$value" == 'true' ]]
+  [[ "$value" == 'false' ]]
+}


mbland:
FYI, chapter and verse from https://tiswww.case.edu/php/chet/bash/CHANGES indicating that this was another issue resolved in Bash 4.1-alpha:

```
This document details the changes between this version, bash-4.1-alpha,
and the previous version, bash-4.0-release.

3.  New Features in Bash

j.  The [[ and (( commands are now subject to the setting of `set -e' and the
    ERR trap.
```

sublimino:
Simple changes but it took me a while to grok this (perhaps because I don't have a mac!). There's [a note on that issue](https://github.com/sstephenson/bats/issues/49#issuecomment-38510279):

> We don't have tutorials on writing actual tests (this is supposed to be left as an exercise to the reader)

I'd like this to change (especially Mac vs Linux, BASH 3 vs 4), what do you think @mbland? Happy to contribute to these.

On this PR, testing this end on `GNU bash, version 4.4.12(1)-release (x86_64-pc-linux-gnu)` all these are fine, [the builds pass for BASH 3 and 4](https://travis-ci.org/bats-core/bats-core/jobs/284742184), and the tests cover the new functionality.
 
LGTM

mbland:
Ah hell yeah! I'm all about producing living, evolving documentation artifacts to communicate what makes for good style and good practice given the best ideas available over time.

BTW, as occurred to me only this morning, this change won't deal with double bracket tests that are followed by a semicolon, such as `@test foo { [[ -n "$bar" ]]; }`. But since I doubt that's a common idiom or high-demand feature, the official, published recommendation may be "don't do that then!"

sublimino:
A fine point re: trailing `;`. I suppose that space could also be a tab, and there could be multiple spaces, tabs, and semicolons, as well as comments (I presume?). Maybe we're into actual parser territory and should just throw and error :confused:

jasonkarns:
@sublimino Would you mind creating a wiki page for examples in the interim? We could transition those to either an /examples suite of sample tests, or even a full blown demo repo at some point.

@mbland I'm waffling a bit on this feature. I'm intentionally calling it a feature and not a bug because the underlying issue is not a bug with bats at all, but a bug with bash itself. As a feature, I'm generally in favor of this change as it will improve the lives of bats test authors. However, the current fix has two quirks: the semicolon for one, as well as the fact this won't be fixed if the double-bracket is used in the single-line form. (Which technically makes it a special case of the semi-colon quirk, though addressing it would necessitate a bigger change to this implementation than merely improving the regex.)

So, since this fix doesn't _uniformly_ catch the double-bracket problem... And since the underlying issue is with bash itself, not bats... I would much rather we (for now) just print a warning message (which could be more robust in its implementation, catching the semicolon and singleline variants).

Also, this will be the first time that bats is fundamentally altering our users' tests. I feel like this should be given a bit more pause. Perhaps put this "helpful" alteration (and others to come?) behind an option? I feel queasy fundamentally modifying a user's tests without the ability to disable it. (regardless whether it's opt-in or opt-out)

Further, this whole thing really directs us to a discussion around bash version support. We already have #28, and I'd like to get some sort of consensus on versions of Bash we officially support. From there, we can determine a policy for handling known bash "bugs" (real or perceived), and whether that includes ignoring them, warning about them, or attempting to gracefully fix them.

jasonkarns:
(unclassified this as a bug as per above)

sublimino:
@jasonkarns sure thing, have put up skeleton and will flesh out and add some executable examples to core https://github.com/bats-core/bats-core/wiki/Usage-Examples-(WIP)

mbland:
@jasonkarns Running short on time yesterday and today—I _will_ address this and other PRs in full—but I have an some idea how to address all double-brackets with the option to disable. Will post more tomorrow for certain.

mbland:
After thinking through it a bit, @sublimino may be right: a full solution may be tantamount to fully parsing Bash. Aside from the present implementation not catching semicolons or single-line tests, there's the matter of double-bracket expressions appearing in `if` and `while` statements, so determining whether `\]\][[:space:]]*;` comes before a `do` or `then` or whether `\]\][[:space]]*$` is followed by a `do` or `then` on the next line could be a problem.

What we maybe could do instead is match `^[[:space:]]*\[\[` or `;[[:space:]]*\[\[` and issue a warning that such assertions won't trigger test failures for Bash < 4.1, and that they should be wrapped in an `if` block. Chapter and verse from https://tiswww.case.edu/php/chet/bash/CHANGES:

```
This document details the changes between this version, bash-4.1-alpha,
and the previous version, bash-4.0-release.

3.  New Features in Bash

j.  The [[ and (( commands are now subject to the setting of `set -e' and the
    ERR trap.
```

Of course, this compatibility issue and a desire to see values producing a failure is why I started writing `lib/bats/assertions` in mbland/go-script-bash. 😛

mbland:
Another thought: We _could_ also add a feature whereby the user can set `BATS_USING_DOUBLE_BRACKET_ASSERTIONS`, effectively pledging that all `\]\][[:space:]]*$` patterns _are_ assertions, and we can inject the `|| return 1` in that case.

sublimino:
Based on @jasonkarns observation [above](https://github.com/bats-core/bats-core/pull/24#issuecomment-336240982)

> this will be the first time that bats is fundamentally altering our users' tests

is there enough appetite and dependent test coverage to ship this? 

My preference is to err towards uber-caution, and defer this until we have a comprehensive test suite for BATS consumers, so removing it from 0.5.0 https://github.com/bats-core/bats-core/milestone/1

tterranigma:
@sublimino This task is still listed as blocking 0.5.0. Is it really?

sublimino:
@tterranigma I'd personally like to remove it from 0.5.0 to give it more attention as long as everybody's happy with that.

mbland:
Yeah, I'd remove it for now.

sublimino:
Removed this issue from [0.5.0 milestones](https://github.com/bats-core/bats-core/milestone/1)

martin-schulze-vireso:
The essence of the above discussion seems to be: Doing the detection reliable is hard/amounts to having a bash parser. Lets get back to the drawing board in #18, since the original author won't drive this further (AFAIK).

PR #26, opened 2017-10-07
Support "unoffical Bash strict mode"

mbland:
Part of #20. Addresses sstephenson/bats#171 filed by @felixSchl.

*Note: I'd originally called this "unofficial **Bats** strict mode" after frying my brain from debugging for a couple hours.*

Bats is now capable of executing test cases that conform to the "unoffical Bash strict mode" described by:

- http://redsymbol.net/articles/unofficial-bash-strict-mode/

While the test fixture is straightforward, it took trial and error to realize a few things. First, `set -u` changed in Bash 4.1-alpha. Chapter and verse from https://tiswww.case.edu/php/chet/bash/CHANGES:

```
  This document details the changes between this version,
  bash-4.1-alpha, and the previous version, bash-4.0-release.

  n.  Fixed the behavior of `set -u' to conform to the latest Posix
      interpretation: every expansion of an unset variable except $@ and
      $* will cause the shell to exit.
```

This almost certainly accounts for the fact that referencing `"${FOO[@]}"` when `FOO=()` is defined is OK under the latest Bash 4.4.12(1)-release, but breaks under the Bash 3.2.57(1)-release that ships with macOS. The fix was to do an array length check before each such operation.

It probably also accounts for the fact that under Bash 3.2.57, no error output is generated at all for certain unset variable accesses. The workaround there is to run the test using the latest version of Bash.

The biggest surprise was realizing that no error output is visible when the unset variable access happens inside a DEBUG trap handler. All versions of Bash seem susceptible to this.

As a result of these debugging subtleties, the test case contains a thorough error message to suggest how to proceed when the fixture fails.

### DIFF
--- a/libexec/bats
+++ b/libexec/bats
@@ -100,34 +100,38 @@ for arg in "$@"; do
 done
 
 unset count_flag pretty
+count_flag=''
+pretty=''
 [ -t 0 ] && [ -t 1 ] && pretty="1"
-[ -n "$CI" ] && pretty=""
+[ -n "${CI:-}" ] && pretty=""
 
-for option in "${options[@]}"; do
-  case "$option" in
-  "h" | "help" )
-    help
-    exit 0
-    ;;
-  "v" | "version" )
-    version
-    exit 0
-    ;;
-  "c" | "count" )
-    count_flag="-c"
-    ;;
-  "t" | "tap" )
-    pretty=""
-    ;;
-  "p" | "pretty" )
-    pretty="1"
-    ;;
-  * )
-    usage >&2
-    exit 1
-    ;;
-  esac
-done
+if [[ "${#options[@]}" -ne '0' ]]; then
+  for option in "${options[@]}"; do
+    case "$option" in
+    "h" | "help" )
+      help
+      exit 0
+      ;;
+    "v" | "version" )
+      version
+      exit 0
+      ;;
+    "c" | "count" )
+      count_flag="-c"
+      ;;
+    "t" | "tap" )
+      pretty=""
+      ;;
+    "p" | "pretty" )
+      pretty="1"
+      ;;
+    * )
+      usage >&2
+      exit 1
+      ;;
+    esac
+  done
+fi
 
 if [ "${#arguments[@]}" -eq 0 ]; then
   usage >&2
--- a/libexec/bats-exec-test
+++ b/libexec/bats-exec-test
@@ -73,6 +73,7 @@ teardown() {
   true
 }
 
+BATS_TEST_SKIPPED=''
 skip() {
   BATS_TEST_SKIPPED=${1:-1}
   BATS_TEST_COMPLETED=1
@@ -89,11 +90,16 @@ bats_test_begin() {
 
 bats_test_function() {
   local test_name="$1"
-  BATS_TEST_NAMES["${#BATS_TEST_NAMES[@]}"]="$test_name"
+  BATS_TEST_NAMES+=("$test_name")
 }
 
+BATS_CURRENT_STACK_TRACE=()
+BATS_PREVIOUS_STACK_TRACE=()
+
 bats_capture_stack_trace() {
-  BATS_PREVIOUS_STACK_TRACE=( "${BATS_CURRENT_STACK_TRACE[@]}" )
+  if [[ "${#BATS_CURRENT_STACK_TRACE[@]}" -ne '0' ]]; then
+    BATS_PREVIOUS_STACK_TRACE=("${BATS_CURRENT_STACK_TRACE[@]}")
+  fi
   BATS_CURRENT_STACK_TRACE=()
 
   local test_pattern=" $BATS_TEST_NAME $BATS_TEST_SOURCE"
@@ -254,7 +260,7 @@ bats_teardown_trap() {
 
 bats_exit_trap() {
   local status
-  local skipped
+  local skipped=''
   trap - err exit
 
   if [ -n "$BATS_TEST_SKIPPED" ]; then
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -316,7 +316,7 @@ $(< "$FIXTURE_ROOT/unofficial_bash_strict_mode.bash")
 --------
 
 See:
-- https://github.com/sstephenson/bats/issues/171'
+- https://github.com/sstephenson/bats/issues/171
 - http://redsymbol.net/articles/unofficial-bash-strict-mode/
 
 If there is no error output from the test fixture, run the following to
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -85,8 +85,7 @@ fixtures bats
 @test "one failing test" {
   run bats "$FIXTURE_ROOT/failing.bats"
   [ $status -eq 1 ]
-  printf 'lines:\n' >&2
-  printf '%s\n' "${lines[@]}" >&2
+  emit_debug_output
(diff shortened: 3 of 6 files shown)


test/bats.bats
```diff
@@ -300,3 +297,44 @@ fixtures bats
...
+
+This test failed because the Bats internals are violating one of the
+constraints imposed by:
+
+--------
+$(< "$FIXTURE_ROOT/unofficial_bash_strict_mode.bash")
+--------
+
+See:
+- https://github.com/sstephenson/bats/issues/171'
```
sublimino: Trailing `'`

libexec/bats-exec-test
```diff
@@ -89,11 +90,16 @@ bats_test_begin() {
...
-  BATS_TEST_NAMES["${#BATS_TEST_NAMES[@]}"]="$test_name"
+  BATS_TEST_NAMES+=("$test_name")
 }
 
+BATS_CURRENT_STACK_TRACE=()
+BATS_PREVIOUS_STACK_TRACE=()
+
 bats_capture_stack_trace() {
-  BATS_PREVIOUS_STACK_TRACE=( "${BATS_CURRENT_STACK_TRACE[@]}" )
+  if [[ "${#BATS_CURRENT_STACK_TRACE[@]}" -ne '0' ]]; then
```
sublimino: Nit: not that it makes a difference to parsing, but `'0'` -> integer `0` ?

mbland: It's a preference thing: I've been bitten by unquoted values so many times, that rather than try to remember when to quote/when not to, I literally quote _everything_. (Except for variables used as the pattern in a `=~` expression, which you never want to quote.)

Granted, it may seem like overkill for integer comparisons, and I'd eventually relent on this point if the pressure mounts; but for now I'd just as soon leave it.

sublimino: No pressure at all, nonstandard but for good reason.

Perhaps we could have a project styleguide? I default to https://google.github.io/styleguide/shell.xml but anything would do. Will open an issue if there's nothing already.

test/bats.bats
mbland: Fixed, thanks!

And yes, I'm a huge proponent of fixing even the tiniest details such as this whenever we notice them. The less sloppiness, the less cruft, the less friction, the better experience for everybody!

mbland:
Thanks, @sublimino!

@btamayo and @jasonkarns, not sure if you were notified since I added bats-core as a reviewer. Will hold off on merging until I get an ack from at least one of you.

libexec/bats-exec-test
sublimino: Your detailed styleguide at https://github.com/mbland/go-script-bash/blob/master/CONTRIBUTING.md#formatting is great @mbland!

mbland: Aww shucks, thanks! Feel free to open an issue and maybe reference mbland/go-script-bash#29. I've been wanting to do something like that for a long time.

Not to oversell my wares, but I was considering using [mbland/guides-template](https://github.com/mbland/guides-template) to produce it in the near future, as I'm presently dusting it off for use at my new company. I have a whole theory behind why I want to use it; my very rough [Documenation as organization hacking tools](https://mike-bland.com/docs-as-hacking-tools) deck contains a bit about it.

Also, regarding the Google style guide: Full disclosure, I'm an ex-Googler. We didn't have the Bash guide when I was there (2005-2011), I don't think—but when I came across it a few weeks ago, I was surprised/not surprised to find I wrote most of mbland/go-script-bash in conformance with it already! 😛 

This speaks to the power of good conventions that transcend individual languages, and guides are critical to developing and disseminating them. This said, I'm a proponent of using existing work as a _template_ (see what I did there?) and fitting it to the project/organization as needed. If innovations emerge, then they can be proposed for inclusion upstream. (And if only we had a tool that supported such a distributed editing model...) I find it provides a nice balance between leveraging prior art and allowing for shared ownership rather than slavish insistence on conformity for conformity's sake.

Enough philosophy for one code review comment? 😛

sublimino: > slavish insistence on conformity for conformity's sake
Ah yes, oft observed.

Agree strongly on the contents of that slide deck, and support using https://github.com/mbland/guides-template - thanks for the time and thought you're contributing to this project/ecosystem :smile: Bring on the philosophy!

mbland:
Pinging @btamayo and @jasonkarns... I'm not in a huge hurry, but a heads-up on when I can expect one/both of y'all to take a look at this and my other PRs would be appreciated, even if it isn't for some time. Or if you're OK with me just merging. I'm excited to keep the momentum going!

libexec/bats
```diff
@@ -100,34 +100,38 @@ for arg in "$@"; do
 done
 
 unset count_flag pretty
```
jasonkarns: We can kill the unset now, right? Since we're forcefully setting to `""`?

libexec/bats-exec-test
jasonkarns: I'm also having a visceral reaction to the quoted integer... :)

If we are going to quote it (personally, I'm still 👎  on quoting literals, especially integers), can we at least stick with doublequotes?

libexec/bats
```diff
@@ -100,34 +100,38 @@ for arg in "$@"; do
...
+count_flag=''
```
jasonkarns: As we don't have a styleguide... is there a strong objection to using `""` for all the variable (re)sets?

jasonkarns:
Tangentially related... Should we consider making bats itself set unofficial strict mode?

libexec/bats
mbland: My general approach is single quotes for anything that doesn't have a variable expansion or a `'` character. But it isn't a strong guideline.

mbland:
@jasonkarns Thanks! Will go ahead and merge. Will rebase the other PR branches tomorrow, unless I steal a few minutes today (they should all conflict because all add a new test case at the end of `test/bats.bats`).

mbland:
Oh, one more thing @jasonkarns: Adding a feature to enforce this mode is an interesting idea. Wanna file it for v1.0.0 (or v1.1.0)?

jasonkarns:
I'd say definitely post-1.0 since it'd be a "big" thing. (likely incorporating other bits of strictness/helpfullness as discussed in https://github.com/bats-core/bats-core/pull/23#issuecomment-336245484

jasonkarns:
Curious why this PR hasn't been moved to "Merged to master" on the project board...

mbland:
@jasonkarns Done. Can you not move them, and merge PRs, or are you preferring not to? (I was just assuming that, as maintainers, any of us could just do stuff like that.)

jasonkarns:
I never tried, I assumed they would move automatically. (Are they all moved manually?)

TBH, I haven't used GH Projects much, as I've always used Waffle (which _does_ move them automatically)... just learning :D

mbland:
Yeah, I've always moved them manually in GH projects, and I don't see a setting to set any kind of triggers on merged PRs.

btamayo:
@jasonkarns @mbland https://waffle.io/bats-core/bats-core/join :)

btamayo:
> Oh, one more thing @jasonkarns: Adding a feature to enforce this mode is an interesting idea. Wanna file it for v1.0.0 (or v1.1.0)?

> https://github.com/bats-core/bats-core/pull/23#issuecomment-336245484

Created: https://github.com/bats-core/bats-core/issues/36 :)

PR #27, opened 2017-10-09
Allow sourcing of helper files from BATS_LIB_PATH 

ntnn:
The existing functionality of `load()` is preserved:
    1. Load helper files in the same directory the current testfile
       resides in
    2. Load helper files by absolute path

Additionally an environment variable BATS_LIB_PATH can be defined to
be source helper files from.

If BATS_LIB_PATH is empty the following locations are used:
    1. $HOME/.bats/lib
       Allows users to manage helper libraries themselves, similar to
       python/ruby/...
    2. /usr/lib/bats
       Allows to install helper libraries via a package manager.

Example usage:

* bats installed via package manager
* [bats-support](https://github.com/ztombol/bats-support) files installed via package manager to `/usr/lib/bats`

```sh
$ tree /usr/lib/bats
/usr/lib/bats
├── bats-support/
│   ├── error.bash
│   ├── lang.bash
│   └── output.bash
└── bats-support.bash

1 directory, 4 files
$ cat /usr/lib/bats/bats-support
source "$(dirname "${BASH_SOURCE[0]}")/bats-support/output.bash"
source "$(dirname "${BASH_SOURCE[0]}")/bats-support/error.bash"
source "$(dirname "${BASH_SOURCE[0]}")/bats-support/lang.bash"
$ cat example.bats
#!/usr/bin/env bats

load bats-support
@test "This will fail" {
    fail "This is an error message with error.bash loaded from /usr/lib/bats/bats-support/error.bash"
}
$ bats example.bats
1..1
not ok 1 This will fail
# (from function `fail' in file /usr/lib/bats/bats-support/error.bash, line 40,
#  in test file example.bats, line 6)
#   `fail "This is an error message with error.bash loaded from /usr/lib/bats/error.bash"' failed
# This is an error message with error.bash loaded from /usr/lib/bats/error.bash
```

If no loading file exists all files in a directory are sourced.

As a maintainer and active user of bats in project I'd very much like to see a proper library path implemented.


mbland:
Interesting concept. This will have to wait for [v1.1.0](https://github.com/bats-core/bats-core/milestone/2), as we're focused on getting [v1.0.0](https://github.com/bats-core/bats-core/milestone/1) out the door, so it will be a little while before we merge. See the discussion in #16 for details.

That said, I've added this PR to the v1.1.0 milestone to keep track of it, and I'm going to propose an idea in #16 to see if we can't expedite contributions such as this.

ntnn:
I've rebased the PR onto the current master.

ntnn:
@sublimino I've rebased the PR again.

I've forgone using `PATH` as mentioned in ztombol/bats-docs#15 in favor of my original proposal to introduce `BATS_LIB_PATH`. `PATH` is usually already overly full (on my own systems alone it already has 20-oddish entries, at work I usually exceed 40).

Using a separate variable also prevents mixups with existing executables and libraries (e.g. a library that provides bats-functions for xvfb-run and the binary xvfb-run itself).

As in my original PR I've preserved existing functionality. Relative import paths still work as they're expanded from `BATS_TEST_DIRNAME`.

The only difference to my original PR is that it now supports the `load` and `load.bash` files for libraries - my original PR only considered placing initializion scripts outside of the directory with the same name (e.g. `/path/to/example.bash` is the loading file for `/path/to/example/`).

That also solves the problem with the `load.bash` idiom @dimo414 noted in https://github.com/ztombol/bats-docs/issues/15.

If the changes are OK I'll fix the tests and add new ones for the tests - but I don't want to invest more time than necessary into a PR that has been ignored for three years now.

henning-schild:
Did not even care to read all this, how come a PR stays open for so long and not being worked on?

https://github.com/gentoo/gentoo/pull/12280
https://github.com/gentoo/gentoo/pull/16409

sublimino:
@ntnn thanks for the rebase. Code looks good and conceptually this sounds good to me, can @dimo414 and @martin-schulze-vireso see what they think of this proposal too?

If we're in agreement I'll look to target `1.3.0-dev` after `1.2.1` is released. 

As on the other thread, apologies for the delay with `bats-core` PRs. I tend to direct my efforts to "big features" at the cost of smaller ones so the project maintains some sort of velocity, this change seems worthwhile and happy to work with you to get it merged : ]

ntnn:
The breaking test is fixed, I've added two more for llibrary loading with and without a `load.bash` and updated the docs.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,88 @@
...
-  source "${file}"
+  # Replace spaces with __SPACE__ to turn back into spaces after
+  # splitting in the for loop.
+  # Splitting on spaces is required to prevent eating input. However
+  # users may have places paths in BATS_LIB_PATH with spaces - those
+  # paths would get split as well; hence the substitution and
+  # replacement.
+  bats_lib_path="${BATS_TEST_DIRNAME// /__SPACE__}:${bats_lib_path// /__SPACE__}"
+
+  for part in ${bats_lib_path//:/ }; do
```
rico-chet, line 78: Note how the code breaks if there is `__SPACE__` in  `$bats_lib_path`, oops.
I tried `mapfile` and it worked:

```bash
  bats_lib_path="${BATS_TEST_DIRNAME}:${bats_lib_path}"                          
                                                                                 
  local -a parts                                                                 
  mapfile -d : -t parts <<< "${bats_lib_path}:"                                  
  for part in "${parts[@]}"; do                                                  
```

test/bats.bats
```diff
@@ -221,14 +221,41 @@ fixtures bats
   [ $status -eq 0 ]
 }
 
-@test "load supports scripts on the PATH" {
+@test "load supports scripts on the BATS_LIB_PATH" {
   path_dir="$BATS_TMPNAME/path"
```
rico-chet, line 225: Add a couple spaces to `path_dir` to test the splitting code. Repeat with the tests following.

test/bats.bats
```diff
@@ -221,14 +221,41 @@ fixtures bats
...
   mkdir -p "$path_dir"
   cp "${FIXTURE_ROOT}/test_helper.bash" "${path_dir}/on_path"
-  PATH="${path_dir}:$PATH"  HELPER_NAME="on_path" run bats "$FIXTURE_ROOT/load.bats"
+  BATS_LIB_PATH="${path_dir}"  HELPER_NAME="on_path" run bats "$FIXTURE_ROOT/load.bats"
   [ $status -eq 0 ]
```
rico-chet, line 229: This test is too thin, IMHO. Here, I'd expect to see a successful call to a function defined through the `load` call (e.g. `help_me`), so that the value proposal is fulfilled.

lib/bats-core/test_functions.bash
ntnn: Thanks, I didn't know about mapfile - I originally used readarray, but that was only added in ~4.3.

test/bats.bats
ntnn: That is exactly what the fixture `load.bats` does.

test/bats.bats
ntnn: Good idea, I've added a test specifically for whitespace splitting.

ntnn:
Mhm, `mapfile` doesn't support `-d` before 4.3, same problem as with `readarray`. Instead we're now using `read` and set the IFS.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,83 @@
...
+    bats_lib_path="$HOME/.bats/lib:/usr/lib/bats"
   fi
+  bats_lib_path="$BATS_TEST_DIRNAME:$bats_lib_path"
+
+  local -a parts
+  IFS=: read -ra parts <<< "$bats_lib_path"
+
+  for part in "${parts[@]}"; do
+    if _load "$part/$file"; then
+      # _load finished without error, file/library was sourced, return
```
rico-chet, line 76: I prefer it shorter:

> First successful file/library load wins.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,83 @@
...
+    for opt in "${opts[@]}"; do
+        if [[ -f "$opt" ]]; then
+            source "$opt"
+            return
+        fi
+    done
+
+    if [[ -d "$file" ]]; then
+        for contained in "$file"/*.bash; do
+            source "$contained"
```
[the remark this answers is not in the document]
ntnn: Thanks, I fixed that.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,83 @@
...
 
-# Shorthand for source-ing files relative to the BATS_TEST_DIRNAME,
-# optionally with a .bash suffix appended. If the argument doesn't
-# resolve relative to BATS_TEST_DIRNAME it is sourced as-is.
+# _load expects an absolute path to a file or directory to source.
+# If the argument is a file the file will be sourced or the same path
+# with the added suffix `.bash` exists it will be sourced..
+# If the argument is a directory _load checks of the file `load` or
+# `load.bash` inside - if it exists this file is loaded. If it doesn't
+# exist all files with the `.bash` suffix are sourced.
```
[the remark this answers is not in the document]
ntnn: Then no files are sourced. That is implicit. If we start documenting everything `bats` doesn't do we would never finish writing documentation.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,83 @@
...
+
+    for opt in "${opts[@]}"; do
+        if [[ -f "$opt" ]]; then
+            source "$opt"
+            return
+        fi
+    done
+
+    if [[ -d "$file" ]]; then
+        for contained in "$file"/*.bash; do
```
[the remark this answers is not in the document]
ntnn: That is not correct - passing a directory to `source` does not produce a non-zero exit status.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,83 @@
...
+#  - /path/to/example.bash
+#  - /path/to/example
+#  - /path/to/example/load.bash
+#  - /path/to/example/load
+#  - /path/to/example/*.bash
+_load() {
+    local file="${1:?}"
+
+    if [[ "${file:0:1}" != "/" ]]; then
+        printf "Received argument with a relative path, expected absolute: %s\n", "$file"
```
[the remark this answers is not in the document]
ntnn: Noticed that as well, and fixed that with an earlier commit.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,83 @@
...
-  # TODO consider flipping the order here; it would be more consistent
-  # and less surprising to look for an exact-match first.
-  if [[ -f "${BATS_TEST_DIRNAME}/${file}.bash" ]]; then
-    file="${BATS_TEST_DIRNAME}/${file}.bash"
-  elif [[ -f "${BATS_TEST_DIRNAME}/${file}" ]]; then
-    file="${BATS_TEST_DIRNAME}/${file}"
+  # Check if target is absolute
+  if [[ "${file:0:1}" == "/" ]]; then
+    if ! _load "$file"; then
+      printf "Failed to load file '%s'\n" "$file" >&2
```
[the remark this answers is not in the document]
ntnn: Thanks, fixed.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,83 @@
...
+      return
+    fi
+  done
 
-  # Dynamically loaded user file provided outside of Bats.
-  # Note: 'source "$file" || exit' doesn't work on bash3.2.
-  # shellcheck disable=SC1090
-  source "${file}"
+  printf "Failed to load file or library based on argument '%s'\n" "$file" >&2
+  exit 1
```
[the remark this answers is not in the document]
ntnn: I came from the viewpoint that I include a library I want to use it - but I see being `return 1` more useful - e.g. to then try sourcing another library if one fails (similar to how in python the switch from 2 to 3 was largely handled).
I'm updating it - thanks!

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,83 @@
...
+      # _load finished without error, file/library was sourced, return
+      return
+    fi
+  done
 
-  # Dynamically loaded user file provided outside of Bats.
-  # Note: 'source "$file" || exit' doesn't work on bash3.2.
-  # shellcheck disable=SC1090
-  source "${file}"
+  printf "Failed to load file or library based on argument '%s'\n" "$file" >&2
```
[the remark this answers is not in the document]
ntnn: No shell and no application I know of lists its configured lookup paths when it fails to load something.

lib/bats-core/test_functions.bash
ntnn: Only the first found wins - `_load` does not go through the entire path and _tries_ any targets. The first match it finds it sources and then returns.

lib/bats-core/test_functions.bash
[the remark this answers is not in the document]
ntnn: Disregard my previous comment - in zsh it doesn't produce a non-zero return code; in bash it does. I'll push a fix.

lib/bats-core/test_functions.bash
ntnn: Yeah, with the changes you suggested to use `return` instead of `exit` the first successfull library wins. Imho that would be wrong though - the first matching library should be sourced and if that fails an error should be reported. I'll update that accordingly.

ntnn:
@sublimino: @rico-chet requested that the `load` function shouldn't call `exit` - which makes sense to me, to allow to e.g. load a different library instead.

However - is it even possible with `bats`' parser to have conditionals around `load` statements in `bats` files?

If it is and that is possible I'll rewrite the logic to split up discovery and load attempts - otherwise the code would get too messy and lower priority libraries would shadow failing higher priority libraries.
Otherwise we can leave it at a call to `exit`.

test/bats.bats
```diff
@@ -221,11 +221,11 @@ fixtures bats
...
 }
 
-@test "load supports scripts on the PATH" {
+@test "load supports scripts on the BATS_LIB_PATH" {
```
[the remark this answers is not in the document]
ntnn: No, loading scripts from PATH was never my intention and I've stated that in both the original PR and the comment the the recent rebase.
If the user wants to load scripts from PATH they can add it to BATS_LIB_PATH if they desire.

test/bats.bats
```diff
@@ -265,6 +265,52 @@ fixtures bats
...
+  for lib in liba libb libc; do
+      mkdir -p "$path_dir/$lib"
+      cp "${FIXTURE_ROOT}/exit1.bash" "$path_dir/$lib/load.bash"
+  done
+  mkdir -p "$path_dir/test_helper"
+  cp "${FIXTURE_ROOT}/test_helper.bash" "$path_dir/test_helper/load.bash"
+  BATS_LIB_PATH="$path_dir" HELPER_NAME="test_helper" run bats "$FIXTURE_ROOT/load.bats"
+}
+
+@test "bats  when a library errors while sourcing" {
```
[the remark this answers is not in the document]
ntnn: Yep, I've fixed that locally when I started to look into returning instead of exiting, but haven't pushed that commit yet.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,89 @@
...
-    file="${BATS_TEST_DIRNAME}/${file}.bash"
-  elif [[ -f "${BATS_TEST_DIRNAME}/${file}" ]]; then
-    file="${BATS_TEST_DIRNAME}/${file}"
+  # Check if target is absolute
+  if [[ "${file:0:1}" == "/" ]]; then
+    if ! _load "$file"; then
+      printf "Failed to load file or library '%s'\n" "$file" >&2
+      return 1
+    fi
+    return
```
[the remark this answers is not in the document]
ntnn: Why should this be a fatal error in _load?
If _load returns a non-zero returncode (or exits) the conditional will print a message and return.
_load returns a non-zero returncode when it can't find the targeted file/library.
Otherwise the file at the absolute path has been sourced correctly and load can return.

ntnn:
> 
> 
> I had never tried handling `load` errors before, but it looks like a hard fail:
> 
> ```shell
> $ cat test.bats 
> if ! load absent_lib; then
>   echo no load
> fi
> 
> @test "load" {
>   true
> }
> ```
> 
> Then
> 
> ```shell
> bin/bats test.bats
> bats: absent_lib does not exist
>  ✗ 
>    (from function `source' in file test.bats, line 1,
>     from function `bats_run_setup_file' in file libexec/bats-core/bats-exec-file, line 81,
>     from function `main' in test file libexec/bats-core/bats-exec-file, line 207)
>      `bats_run_setup_file' failed
> 
> 1 test, 1 failure
> ```
> 
> This is from the existing behaviour in `lib/bats-core/test_functions.bash`:
> 
> ```
> 21    if [[ ! -f "$file" ]] && ! type -P "$file" >/dev/null; then
>   1     printf 'bats: %s does not exist\n' "$file" >&2           
>   2     exit 1                                                   
>   3   fi                                                         
> ```

I'd say that behaviour should be kept, otherwise we'd see `set -e` or `||die` sprinkled in bats files.
But I could see adding a `load_safe` method that doesn't call exit.

ntnn:
I've rewritten the `load` with more helper functions, that makes the flow a bit clearer.
Now there's also a `load_safe` (I don't really like the name, please do suggest alternatives) to allow for conditional imports.

ntnn:
@rico-chet: I can't re-request your review, but please do have a look.

test/bats.bats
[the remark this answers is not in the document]
sublimino: Ah, that wasn't clear to me from `The existing functionality of load() is preserved` in this PR. 

I can't say I use `PATH` to load libraries so didn't pick it up, so to be clear: this is an intentionally breaking change for users that do use PATH?

lib/bats-core/test_functions.bash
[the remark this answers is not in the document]
sublimino: IIRC in a previous patchset there was an `exit 1` around this line https://github.com/bats-core/bats-core/pull/27/files#diff-223055a1b11f29286d5c612d214344c7R159

Apologies if that was a mis-remember. Current patchset looks correct WRT return codes, will test again.

test/bats.bats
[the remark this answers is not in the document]
ntnn: Well, PATH wasn't in the picture for `load` before the last release.

If you want to maintain that `PATH` is targeted by default that can be added. But I'd advice strongly against that, I've noted previously that this could cause confusion when a BATS library is created for an existing application.
The example I've noted previously is `xvfb-run`, where both the BATS library and the application itself would be called `xvfb-run`, confusing `load`.

That `PATH` can be added to `BATS_LIB_PATH` can of course be documented in the readme's `load` section.

rico-chet:
I will try to get to it this weekend.

README.md
```diff
@@ -362,18 +364,87 @@ load test_helper.bash
...
+- `/var/lib/bats/test_helper` (directory)
+
+If the found load path is a file it is sourced.
+If the found load path is a directory all files ending `.bash` inside of
+the directory are sourced in alphanumerical order.
+
+If `source` returns with an error on any of the source files `load`
+prints a message and Bats exits.
+
+__Note:__ `BATS_LIB_PATH` defaults to `$HOME/.bats/lib:/usr/lib/bats` if
```
martin-schulze-vireso, line 438: The install path of bats is actually `$PREFIX/lib/bats-core` I think we should harmonize that.

README.md
```diff
@@ -362,18 +364,87 @@ load test_helper.bash
...
+- `~/.bats/libs/test_helper/load`
+- `~/.bats/libs/test_helper` (directory)
+- `/var/lib/bats/test_helper.bash`
+- `/var/lib/bats/test_helper`
+- `/var/lib/bats/test_helper/load.bash`
+- `/var/lib/bats/test_helper/load`
+- `/var/lib/bats/test_helper` (directory)
+
+If the found load path is a file it is sourced.
+If the found load path is a directory all files ending `.bash` inside of
```
martin-schulze-vireso, line 432: I think that behavior can be somewhat surprising. Why do we need this? Can't we just use a custom `load.bash` to achieve the same result?

If we really need it, we should make it explicit by requiring users to write `load test_helper/` (note the trailing slash).

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,179 @@
...
+# If no library load path can be found load_safe prints an error message
+# and returns 1.
+load_safe() {
+  local slug="${1:?}"
+
+  # Check if slug is an absolute path
+  if [[ "${slug:0:1}" == / ]]; then
+
+    # Check for library load paths
+    local library_path="$(find_library_load_path "$slug")"
```
martin-schulze-vireso, line 151: Won't `find_library_load_path` try to prepend prefixes instead of using this absolute path as is?

Maybe we should have additional tests for that.

test/bats.bats
```diff
@@ -191,6 +191,105 @@ fixtures bats
   [ "${lines[2]}" = "# (in test file $FIXTURE_ROOT/failing.bats, line 4)" ]
 }
 
+@test "find_library_load_path finds single-file libraries with the suffix .bash" {
```
martin-schulze-vireso, line 194: I think with the amount of tests here, an own file for test of load and affiliated functions would be in order.

test/bats.bats
```diff
@@ -191,6 +191,105 @@ fixtures bats
...
+  lib_dir="$BATS_TMPNAME/find_library_path/single_file_suffix"
```
martin-schulze-vireso, line 195: This PR has been overtaken by some others that might help now: We should use the new `BATS_*_TMPDIR` variables for this.

lib/bats-core/test_functions.bash
```diff
@@ -3,30 +3,179 @@
...
+#
+# Library load paths are recognized using find_library_load_path.
+#
+# If no library is found find_in_bats_lib_path returns 1.
+find_in_bats_lib_path() {
+  local library_name="${1:?}"
+
+  local bats_lib_path="$BATS_LIB_PATH"
+  if [[ -z "$bats_lib_path" ]]; then
+    bats_lib_path="$HOME/.bats/lib:/usr/lib/bats"
```
martin-schulze-vireso, line 58: Maybe we should put that default into `libexec/bats-core/bats` instead and let users override later? This would allow them to prepend/append their own paths without removing the default paths, which is closer to the usual PATH handling.

test/bats.bats
```diff
@@ -221,11 +320,11 @@ fixtures bats
   [ $status -eq 0 ]
 }
 
-@test "load supports scripts on the PATH" {
+@test "load loads scripts on the BATS_LIB_PATH" {
```
martin-schulze-vireso, line 323: I am wary of changing the default. I don't know how many projects out there rely on loading from PATH right now. I saw your discussion for avoiding PATH but I'd like to avoid breaking existing code.

I would say we should deprecate loading from PATH and fully disable it with the next major release. Maybe we could add a switch for using PATH or BATS_LIB_PATH to ease the transition.

test/bats.bats
```diff
@@ -265,6 +364,59 @@ fixtures bats
...
+  [ $status -eq 1 ]
+}
+
+@test "bats skips directories when sourcing .bash files in library" {
+  path_dir="$BATS_TMPNAME/libraries_skip_dir/"
+  mkdir -p "$path_dir/target_lib/adir.bash"
+  cp "${FIXTURE_ROOT}/test_helper.bash" "$path_dir/target_lib/test_helper.bash"
+  BATS_LIB_PATH="$path_dir" HELPER_NAME="target_lib" run bats "$FIXTURE_ROOT/load.bats"
+}
+
```
martin-schulze-vireso, line 419: Please add a test for loading absolute paths.

martin-schulze-vireso (changes requested):
I am sorry for the delay. Let me know if you need help with rebasing this onto current master.

test/bats.bats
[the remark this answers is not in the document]
martin-schulze-vireso: It is really unfortunate, that we got into this situation because this solution is much cleaner. Nevertheless, we have to deal with the fact that some users might rely on PATH now. I added a comment for this in my review so we can continue the discussion over there. The gist is: We should preserve the current behavior but transition to the suggestion by @ntnn without hard breaks for existing users.

test/bats.bats
martin-schulze-vireso: We already have tests that load absolute paths (e.g. "load supports plain symbols"). When using BATS_LIB_PATH, an absolute path does not make sense.

lib/bats-core/test_functions.bash
martin-schulze-vireso: I mistook that for `find_in_bats_lib_path`. `find_library_load_path` just tries the common extensions as expected. Also, there is already a test ("load supports plain symbols") that loads absolute paths.

README.md
martin-schulze-vireso: I have been mulling this over for some time now and am decided that I am against it. I will also slash down the other additional load paths that are tried to get back to some minimalism:

We already have hard to understand behavior with `load` as is and the case table just exploded. I don't see the headaches of debugging the complex automatic loading mechanism worth it to just shave some `.bash` or `load` off the end of the `load` command. If users are really annoyed by this, they could simply put up a wrapper `load_lib` that appends `load.bash` to their string and be done.
Loading an extensive list of `.bash` files from a directory might be a more tedious case but I'd first like to see some examples of this to decide whether we should support that.

martin-schulze-vireso: My final decision is to keep `load` separate from the newly added `bats_load_library`. The old `load` should remain as close as possible to the original implementation and the new function should just deal with loading libraries from the defined load paths. It allows for loading libraries with only the library name (e.g. `bats_load_library 'bats-assert'`) or with the specific file to load (`bats_load_library 'bats-detik/lib/detik.bash'`).

For minimal source changes I am thinking about a feature flag to let `load` fall back to `bats_load_library` but the migration path does not really seem promising.

README.md
martin-schulze-vireso: I decided to keep the internal lib/bats-core separate from the helper libs, so we will stay with `/usr/lib/bats`

lib/bats-core/test_functions.bash
martin-schulze-vireso: I decided to have no default value at all and to produce an error message when `bats_load_library` is used without setting `BATS_LIBRARY_PATH`. While the documentation encourages package maintainers to use `/usr/lib/bats` as the lib path, it might not be possible with package managers like `npm` or `brew`. This tries to strike a balance where we use the "natural flow" of the used package manager but unify where it is possible.

martin-schulze-vireso:
I rebased this once again and implemented the changes discussed above in #548, so I am going to close this issue. Thanks for your contribution @ntnn.

martin-schulze-vireso:
With the amount of changes compared to this PR, I'd like to invite all subscribers to this issue to leave a review over at #548. I'll let it sit for a little while for feedback and if there are no showstoppers I will merge it to get the new release out the door ... finally.

PR #32, opened 2017-10-21
Checking permission and symbolic link file

xnum:
If I download the zip file and unzip it on Windows, then upload to my workstation. 
The permission of executable files and symbolic link file are broken.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -34,4 +34,14 @@ cp -R "$BATS_ROOT"/libexec/* "$PREFIX"/libexec
 cp "$BATS_ROOT"/man/bats.1 "$PREFIX"/share/man/man1
 cp "$BATS_ROOT"/man/bats.7 "$PREFIX"/share/man/man7
 
+# fix broken symbolic link file
+if [ ! -L "$PREFIX"/bin/bats ]; then
+    rm -f "$PREFIX"/bin/bats
+    ln -s "$PREFIX"/libexec/bats "$PREFIX"/bin/bats
+fi
+
+# fix file permission
+chmod a+x "$PREFIX"/bin/*
+chmod a+x "$PREFIX"/libexec/*
+
 echo "Installed Bats to $PREFIX/bin/bats"


mbland:
Sorry I haven't looked at this yet; super-slammed at work this week. Will try to take a look this weekend.

mbland:

install.sh
```diff
@@ -34,4 +34,14 @@ cp -R "$BATS_ROOT"/libexec/* "$PREFIX"/libexec
 cp "$BATS_ROOT"/man/bats.1 "$PREFIX"/share/man/man1
 cp "$BATS_ROOT"/man/bats.7 "$PREFIX"/share/man/man7
 
+# fix broken symbolic link file
+if [ ! -L "$PREFIX"/bin/bats ]; then
+    rm -f "$PREFIX"/bin/bats
+    ln -s "$PREFIX"/libexec/bats "$PREFIX"/bin/bats
```
mbland: I just tried this on Git for Windows, and for some reason `ln -s` doesn't want to cooperate:

```bash
$ ./install.sh ../usr
ln: failed to create symbolic link '../usr/bin/bats': No such file or directory
```

Since the "link" is just a copy on this platform anyway, perhaps we should just use `cp`?

xnum: It seems some issue on NTFS
ref: [1](https://cygwin.com/ml/cygwin/2015-05/msg00122.html) 
I got the solutionfrom [2](https://stackoverflow.com/questions/4175264/bash-retrieve-absolute-path-given-relative)
assign absolute path instead of relative path.
It works with `git-bash.exe`

install.sh
```diff
@@ -36,8 +36,9 @@ cp "$BATS_ROOT"/man/bats.7 "$PREFIX"/share/man/man7
 
 # fix broken symbolic link file
 if [ ! -L "$PREFIX"/bin/bats ]; then
-    rm -f "$PREFIX"/bin/bats
-    ln -s "$PREFIX"/libexec/bats "$PREFIX"/bin/bats
+    dir=$(readlink -e "$PREFIX")
```
mbland: Sorry for the delay, but confirmed locally this approach works. Just one thing: please enclose the `$(readlink -e "$PREFIX")` in double-quotes, in case the result includes spaces.

ntnn:
@mbland As I've noted there are nomerous problems with this PR, why was it merged nonetheless?

See here:
https://github.com/bats-core/bats-core/pull/32/files/062160105638ebb101d867d1e9de160b00977e35#diff-3fbb47e318cd8802bd325e7da9aaabe8
 https://github.com/bats-core/bats-core/pull/32/files/062160105638ebb101d867d1e9de160b00977e35#diff-3fbb47e318cd8802bd325e7da9aaabe8
https://github.com/bats-core/bats-core/pull/32/files/062160105638ebb101d867d1e9de160b00977e35#diff-3fbb47e318cd8802bd325e7da9aaabe8
https://github.com/bats-core/bats-core/pull/32/files/062160105638ebb101d867d1e9de160b00977e35#diff-3fbb47e318cd8802bd325e7da9aaabe8

This is quite off-putting for the project, if even a simple install.sh as that can't be maintained properly.

mbland:
@ntnn I never saw that comment...that commit link doesn't show anything, and I can't find it anywhere else in the PR.

That is a fair point about the absolute path and using `ln -sf` instead. Sorry I missed it. Feel free to submit a follow-up PR.

ntnn:
@mbland Huh, github is weird - I'll post a PR later this week. Thanks for answering so quickly.

xnum:
sorry, but `ln -sf` couldn't solve the problem

mbland:
Hmm, that is strange. It seems dependent on the path of the source, not on the path of the target. It's probably best just to copy the file rather than symlink it. Any reason not to do just that?

xnum:
As I mentioned before, the problem is caused with NTFS probably
[reference](https://cygwin.com/ml/cygwin/2015-05/msg00122.html)

copy the file is not working because `bin/bats` links to `libexec/bats`
and find `libexec/bats-exec-test` which not exists in `bin` dir.

bvberkum:
Can we not adjust the installation, or package for windows?

PR #34, opened 2017-10-26
Add test with exported function execution

Pawamoy:
It covers the issue in https://github.com/sstephenson/bats/issues/225.

The test is skipped because it only passes on Bash 4.4. I don't know if this is something that can be solved in Bats itself, but at least we have a test for it.


test/fixtures/bats/exported_function.bats
```diff
@@ -0,0 +1,10 @@
+# see issue https://github.com/sstephenson/bats/issues/225
+
+if exported_function; then
```
mbland: What happens if you try `a=$(declare -F exported_function >/dev/null)` and run it with different Bash versions?

mbland:
Also wondering what chapter and verse from https://tiswww.case.edu/php/chet/bash/CHANGES may apply. Maybe:

```
This document details the changes between this version, bash-4.4-alpha, and
the previous version, bash-4.3-release.

1.  Changes to Bash
ss. Changed the way bash encodes exported functions for inclusion in the
    environment to avoid name collisions with valid variable names and to
    indicate that they are exported functions.
```

test/fixtures/bats/exported_function.bats
Pawamoy: It works with the following:

```bash
# in exported_function.bats
a=$(declare -F exported_function)
```
```bash
# in bats.bats
exported_function() { :; }
```

I downgraded my Bash to 4.3.48(2), and it gives me this:
```
$ exported_function() { :; }
$ export -f exported_function
$ bin/bats --tap test/fixtures/bats/exported_function.bats 
1..1
not ok 1 failing test
# (in file , line 9,
#  from function `bats_perform_test' in file libexec/bats-exec-test, line 313,
#  from function `main' in test file , line 363)
```

### Trying different approaches (original one with `if exported_function; then` only works in 4.4).

This one works in all versions:
```bash
# in exported_function.bats
a=$(exported_function && echo 'exported_function')
```
```bash
# in bats.bats
exported_function() { true; }
```

---

This one works only in 4.4:
```bash
# in exported_function.bats
exported_function && a='exported_function'
```
```bash
# in bats.bats
exported_function() { true; }
```

---

This one works in all versions:
```bash
# in exported_function.bats
a=$(exported_function)
```
```bash
# in bats.bats
exported_function() { echo exported_function; }
```

---

The last one works only in 4.4:
```bash
# in exported_function.bats
exported_function && a='exported_function'
```
```bash
# in bats.bats
exported_function() { return 0; }
```

---

To me it seems that executing `exported_function` in the current process breaks things up, whereas in a subprocess (like `$(exported_function)`) works properly.

mbland:

test/fixtures/bats/exported_function.bats
mbland: Interesting, especially when you consider that `run bats` should be launching a subshell anyways.

Out of curiosity, what happens if you put this `if` condition in `setup()` or inside a `@test` case?

Pawamoy: If I put it in `setup() { ... }` at the beginning of `exported_function.bats`, it works only in Bash 4.4. If I put it inside the `failing test` test case, same thing, it works only in Bash 4.4.

Pawamoy:
I updated the bats test to see what the actual output is when running the failing `exported_function` test:

```
not ok 44 execute exported function without breaking failing test output
# (in test file test/bats.bats, line 378)
#   `false' failed
# 
# Output was:
# 1..1
# not ok 1 failing test
# # (in file , line 8,
# #  from function `bats_perform_test' in file libexec/bats-exec-test, line 313,
# #  from function `main' in test file libexec/bats-exec-test, line 363)
```

Pawamoy:
I added the files needed to run the tests against multiple bash versions, but of course you would not merge this. I don't know what the process is, if I delete this last commit when ready to merge or if I send a new, clean PR.

PR #38, opened 2017-11-17
Added todo directive

andrewfowlie:
I have added the todo directive as described in the TAP specification - https://testanything.org/tap-version-13-specification.html

The syntax is similar to the skip directive:

```
@test "todo test that passes" {
  todo "this todo is done"
  return 0
}

@test "todo test that fails" {
  todo "this todo is outstanding"
  return 1
}
```

This results in 

```
 ✓ todo test that passes (done: this todo is done)
 - todo test that fails (todo: this todo is outstanding)
   (in test file todo.tap, line 7)
     `todo "this todo is outstanding"' failed

2 tests, 0 failures, 1 todo, 1 done
```
and in TAP format:
```
1..2
ok 1 todo test that passes # done this todo is done
not ok 2 todo test that fails # todo this todo is outstanding
# (in test file todo.tap, line 7)
#   `todo "this todo is outstanding"' failed
```


andrewfowlie:
If you like this I can also update docs + add some tests

andrewfowlie:
is it useful?

sublimino:
Hi @andrewfowlie, it looks useful!

bats is currently trying to get to a place where 1.0 can be shipped, I guess this would make sense after that release? Paging @btamayo @mbland @jasonkarns

andrewfowlie:
Cool. The changes are pretty small and it's useful for me because e.g., Jenkins can read TAP and knows about todos, and it'll let me write tests, then write my codes to pass the tests...

I don't know much about the development plan so have no strong opinions about whether it should be in 1.0 or later.

libexec/bats-exec-test
```diff
@@ -80,6 +80,11 @@ skip() {
   exit 0
 }
 
+BATS_TEST_TODOED=''
+todo() {
+  BATS_TEST_TODOED=${1:-1}
```
mbland: The `${1:-1}` needs to be double-quoted.

libexec/bats-exec-test
```diff
@@ -270,14 +277,23 @@ bats_exit_trap() {
     fi
   fi
 
+  if [ -n "$BATS_TEST_TODOED" ]; then
```
mbland: Use `[[` and `]]` for all new code, please.

libexec/bats-exec-test
```diff
@@ -270,14 +277,23 @@ bats_exit_trap() {
...
+    todoed=" # todo"
+    todone=" # done"
+    if [ "1" != "$BATS_TEST_TODOED" ]; then
```
mbland: `[[` and `]]`.

libexec/bats-format-tap-stream
```diff
@@ -51,6 +53,22 @@ skip() {
   advance
 }
 
+todo() {
+  local reason="$1"
+  [ -z "$reason" ] || reason=": $reason"
```
mbland: Use `[[` and `]]`, and expand this into an `if` statement to make it easier to scan.

andrewfowlie:
Thanks for the comments - will make the changes soonish

andrewfowlie:
Is there anything more I can do here?

mbland:
Sorry for the delay. I think we were all running a little lean during August, due to vacations and such (I traveled at the beginning of the month, and organized a large-ish event at work at the end). Will do my best to loop back and take a solid look again soon.

tterranigma:
@andrewfowlie Since you are editing the tap formatter, could you output todo tests with color? Perhaps some blue or cyan? See also #162 .

andrewfowlie:
@tterranigma I’m afraid I’m super busy right now, unlikely I’ll do that any time soon. Feel free to jump in and add colours

Aazhar:
Hello,
I would like to know if this feature will merged anytime soon ?
Thanks

sublimino:
@andrewfowlie (due to the passing of time, sorry!) this branch has some conflicts, could you rebase master onto it and ensure the tests still pass?

In general LGTM, although I'd like a second opinion and to test this in anger for a while. If you'd like to test the latest (e.g. rebased) version of this @Aazhar please do so and let us know here : ]

Aazhar:
ok so after reviewing this PR, I'm questioning the meaning of 'done' todos in the initial post.
Actually imho, all todos should be counted as todos, regardless of their supposed result.

andrewfowlie:
I will try to rebase next week.

Done means that the todo is complete because the test passes. I think it’s useful to indicate that (although could change the wording)

andrewfowlie:
arggh, something has gone wrong with my PR/rebase against master. There was some small complication because some of my modified files had been removed from the upstream repo, so I did some things manually. But I thought it worked (https://github.com/bats-core/bats-core/compare/master...andrewfowlie:master), but this PR seems to include all kinds of commits and changes.

Maybe it's this problem - https://stackoverflow.com/questions/18323665/pull-request-on-github-showing-commits-rebased-from-master

jasonkarns:
I'm assuming part of the problem here is that the changes were made on the master branch instead of on a feature branch.

jasonkarns:
As to "done" todos...

@Aazhar the TAP spec itself indicates that tests marked as TODO that _pass_ should be counted as "bonus" tests.

Personally I disagree with the TAP spec here. I believe rspec's [behavior](https://relishapp.com/rspec/rspec-core/v/3-8/docs/pending-and-skipped-examples/pending-examples#%60pending%60-any-arbitrary-reason-with-a-passing-example) is better here. A test that's marked as TODO should _not_ pass. If it does, the test is either not a valid test (it's testing the wrong thing) and needs to be updated; or the test is valid and the functionality is complete, in which case, the test needs to be updated (remove the TODO marker). Ergo, in both scenarios, the test needs "fixed". Rspec reports `pending` tests that pass as failures, and I believe that is the proper behavior. However, since the TAP spec is explicit, and at least TAP reports them separate from regular passing tests, I'll concede that spec-compliant behavior is acceptable.

However, as a test suite that contains DONE TODOs requires changes, I think we should follow rspec's lead here and return a non-zero exit for the suite in such a case.

andrewfowlie:
That makes sense. I am very busy travelling right now. I’m happy for anyone to finish this PR/if anyone knows how to fix the aforementioned GitHub problem with the PR please let me know

PR #40, opened 2017-11-24
added assert_true method that prints expected and observed value

psteinb:
not sure if this addition complies with [TAP](https://testanything.org/), but it is super helpful to have something that prints the values which break a test

### DIFF
--- a/libexec/bats-exec-test
+++ b/libexec/bats-exec-test
@@ -3,8 +3,6 @@ set -e
 set -E
 set -T
 
-BATS_EXPECTED=""
-BATS_RECEIVED=""
 BATS_COUNT_ONLY=""
 if [ "$1" = "-c" ]; then
   BATS_COUNT_ONLY=1
@@ -187,8 +185,9 @@ bats_print_failed_command() {
     echo -n "failed with status $status"
   fi
 
-  if [ -n ${BATS_EXPECTED} ];then
+  if [ "${BATS_EXPECTED}" ];then
     echo -e ", received: ${BATS_RECEIVED}, expected: ${BATS_EXPECTED}"
+    unset BATS_EXPECTED
   else
     echo
   fi
--- a/libexec/bats-exec-test
+++ b/libexec/bats-exec-test
@@ -3,6 +3,8 @@ set -e
 set -E
 set -T
 
+BATS_EXPECTED=""
+BATS_RECEIVED=""
 BATS_COUNT_ONLY=""
 if [ "$1" = "-c" ]; then
   BATS_COUNT_ONLY=1
@@ -63,6 +65,14 @@ run() {
   [ -z "$E" ] || set -E
   [ -z "$T" ] || set -T
   IFS=$oldIFS
+}
+
+assert_equal() {
+  if [ ! "$1" = "$2" ];then
+    BATS_EXPECTED="$2"
+    BATS_RECEIVED="$1"
+  fi
+  [ "$1" = "$2" ]
 }
 
 setup() {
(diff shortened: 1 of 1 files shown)


psteinb:
Maybe I should refine, what this PR tries to achieve. Given the following simple `bats` file:
```
#!/usr/bin/env bats

@test "this works" {

    run hostname
    [ $status -eq "0" ]

}

@test "this fails" {

    run hostname
    [ $status -eq "1" ]

}
```
I get the following on the command line:
```
$ bats test.sh
 ✓ this works
 ✗ this fails
   (in test file test-noasserts.sh, line 13)
     `[ $status -eq 1 ]' failed

2 tests, 1 failure
```
While this is all nice, it doesn't really tell me which value of `$status` caused the tests to break. To mitigate this, I added a function called `assert_equal` which can help to capture the expected outcome and the received content of the result to check. Now we can do:
```
@test "this works" {

    run hostname
    assert_equal $status "0"

}

@test "this fails" {

    run hostname
    assert_equal $status "1"

}
```
which gives:
```
$ bats ./test.sh 
 ✓ this works
 ✗ this fails
   (in test file test.sh, line 12)
     `run hostname' failed, received: 0, expected: 1

2 tests, 1 failure
```

vikin91:
You could use [bats-assert](https://github.com/ztombol/bats-assert#assert_failure) for that.

psteinb:
Looks very much like the behavior I was looking for. But I wonder why does the bats-core repo not point to these super helpful utilities? Or perhaps I missed that somehow. I have to dig into how I can obtain the entire bats distro.

BraisGabin:
A page in the wiki poiting to that project would be great. Or even a line in the README.

vikin91:
Please note that BATS was not maintained for some years, and the new maintainer joined this project two or three months ago. I guess, there may be still some rough edges here and there.

BraisGabin:
Sure, no blames here. Just ideas 🎉

vikin91:
No worries, I didn't think of it as a blame :) I would also like to see some more movement here in the repo, but I guess people maintaining it are as busy as most of us are.

mbland:
@vikin91 Yeah, sorry, ended up _really_ slammed this month, and Thanksgiving didn't yield all the time I'd hoped. Things are slowing down now, so will do my level best to help move some issues along.

vikin91:
@mbland let me know (or announce to all) where some help is needed the most - I can't promise anything, but I may review or join discussion from time to time.

psteinb:
I just discovered bats last week and find it super useful (I am working 
in the HPC community). As this PR conveys, I am happy to submit PRs if 
they are sensible and useful to others.

Regarding the documentation, how would you install a full bats distro?

mbland:
@psteinb @vikin91 Sorry to leave this/you hanging for so long... Now that we have v1.0.2 out, I think I'd like to look into either including some basic assertions in Bats, and/or making it easy to use separate assertion libraries.

Regarding installation, check out the updated ["Installation" section of our README](https://github.com/bats-core/bats-core#installation). There's more to come, but in a nutshell:

```bash
# Straight from the repo
$ git clone https://github.com/bats-core/bats-core.git
$ bats-core/install.sh /opt/bats-core . # Or directly into /usr, /usr/local, etc.
$ ln -s /opt/bats-core/bin/bats /usr/local/bin/bats # Optional

# Via brew
$ brew install bats-core

# Via npm
$ npm install -g bats
```

There's more to come: Alpine should land soon per alpinelinux/aports#3182, and hopefully we'll make progress on #103 soon.

Going to close this, but feel free to continue the conversation here or in another thread (#44 being a good candidate).

nylen:
Please have a look at #96 which does something similar for the built-in bash assertions `[` and `[[`.  I am pretty convinced that building better support for the built-in shell assertions is the best path for `bats-core`.

PR #46, opened 2017-12-13
Add bad command line option errors

qneill:
Same patch as I posted over in https://github.com/sstephenson/bats/pull/247

### DIFF
--- a/libexec/bats
+++ b/libexec/bats
@@ -6,7 +6,9 @@ version() {
 }
 
 usage() {
+  local errmsg="$1"
   version
+  [ -n "$errmsg" ] && echo "Error: $errmsg"
   echo "Usage: bats [-c] [-p | -t] <test> [<test> ...]"
 }
 
@@ -126,7 +128,7 @@ if [[ "${#options[@]}" -ne '0' ]]; then
       pretty="1"
       ;;
     * )
-      usage >&2
+      usage "Bad command line option '-$option'" >&2
       exit 1
       ;;
     esac
@@ -134,7 +136,7 @@ if [[ "${#options[@]}" -ne '0' ]]; then
 fi
 
 if [ "${#arguments[@]}" -eq 0 ]; then
-  usage >&2
+  usage "Must specify at least one <test>" >&2
   exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


tterranigma:
Maybe out of scope for this PR, but it would be nice to have this also for the `install.sh` script.

libexec/bats
```diff
@@ -6,7 +6,9 @@ version() {
 }
 
 usage() {
+  local errmsg="$1"
   version
+  [ -n "$errmsg" ] && echo "Error: $errmsg"
```
mbland: Use `[[` and `]]` and expand the `&&` expression into an `if` statement.

test/bats.bats
```diff
@@ -3,10 +3,18 @@
 load test_helper
 fixtures bats
 
-@test "no arguments prints usage instructions" {
+@test "no arguments prints message and usage instructions" {
   run bats
   [ $status -eq 1 ]
-  [ $(expr "${lines[1]}" : "Usage:") -ne 0 ]
+  [ $(expr "${lines[1]}" : "Error: Must specify at least one") -ne 0 ]
```
mbland: Extra credit: Replace the `"$(expr ...) -ne 0` expressions with `=` expressions like `${lines[1]} = "Error: Must specify at lest one "*`. Helps the test run a little faster by avoiding subshells/subprocesses.

qneill: will take a look ...

mbland: Actually, I'm a little off base here; `expr` is how we're able to do regular expressions using `[  ]`, since we can't rely on `[[ ]]` working under 3.2.57(1)-release on macOS. So nevermind this comment.

mbland:
@qneill Want to brush this off and get it ready in the near future? I can take your branch and do it myself if sometime soon if you haven't the time.

jasonkarns:
Thoughts about creating `abort` function that takes and prints the error message to stderr, prints usage to stderr, and exits 1?

mbland:
Sounds like a good idea; removes a little duplication and make the intent more clear.

qneill:
Sorry I dropped off the earth - job change and all.  Thanks for resolving

PR #48, opened 2017-12-20
Don't truncate skip message when ')' is included.

cdevinesr:
Addresses sstephenson/bats#240.  When outputting a 'pretty' formatted
output stream, the 'skip' message was being processed with a character
class of [^)].  I couldn't find a reason for this to be the case, so
I've swapped it out to check instead for printable characters, spaces
included ([:print:]).  Please advise if there is a problem with this
approach.

Tested using Bash 4.3.11(1)-release.

### DIFF
--- a/libexec/bats-format-tap-stream
+++ b/libexec/bats-format-tap-stream
@@ -152,7 +152,7 @@ while IFS= read -r line; do
     flush
     ;;
   "ok "* )
-    skip_expr="ok $index (.*) # skip ?(([^)]*))?"
+    skip_expr="ok $index (.*) # skip ?(([[:print:]]*))?"
     if [[ "$line" =~ $skip_expr ]]; then
       let skipped+=1
       buffer skip "${BASH_REMATCH[2]}"
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -242,7 +242,7 @@ fixtures bats
 @test "skipped test with parens (pretty formatter)" {
   run bats --pretty "$FIXTURE_ROOT/skipped_with_parens.bats"
   [ $status -eq 0 ]
-  [[ "${lines[0]}" =~ "- a skipped test with parentheses in the reason (skipped: a reason (with parentheses))" ]]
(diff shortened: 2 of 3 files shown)


mbland:
This is a delayed response, but if you're still game, could you add a test case?

mbland:
Odd that the only failure was in the Alpine container... Try this to debug it:

- Temporarily add `emit_debug_output` after the `run` command in your test.
- Look at the `.travis.yml` for the commands to build and run the tests in Docker locally. (Presuming you have Docker installed, or don't mind installing it.)

cdevinesr:
I agree that it's strange that it's failing on Alpine and nowhere else.  I was planning to test it on Docker locally to figure out why it failed.

cdevinesr:
Ah.  Alpine apparently handles output from the pretty formatter very differently.  This brings to light the question of whether the TAP vs pretty formatter output comparison is currently a valid test on the Alpine container, since it's just an equality check.

Alpine, for a single line test redrawn with carriage returns, outputs this for ${lines[@]}:

   a skipped test with parentheses in the reason
 - a skipped test with parentheses in the reason (skipped:  a reason (with parentheses))

On other systems, only the last line (with a single space after 'skipped:') is returned.  I have an augmented test to fix this, but it's kind of hack-ish, and seems like it'd be awfully prone to breakage with future changes:

```diff
@test "skipped test with parens (pretty formatter)" {
  run bats --pretty "$FIXTURE_ROOT/skipped_with_parens.bats"
  [ $status -eq 0 ]
-  [[ "${lines[@]}" =~ "- a skipped test with parentheses in the reason (skipped: a reason (with parentheses))" ]]
+  [[ "${lines[@]}" =~ "- a skipped test with parentheses in the reason (skipped: "+"a reason (with parentheses))" ]]
}
```

cdevinesr:
No specific reason not to use `$output` instead of '${lines[0]}`.  To be honest, I'm coming back to this change after months of working on other stuff, so I actually forgot about it when writing the test initially, since the stuff I've been working on recently is mostly Python, and so has given me a bit of rust on BATS at the moment.

PR #50, opened 2017-12-27
added uninstall feature

capnajax:


MichaelPereira:
Some feedback: I think it's a good think to have, but maintenance-wise, the proposed changes could be improved to make them more readable and maintainable.

Here are the two main issues that I see:
- It's really not clear which part of the script is doing the installation and the uninstallation. Adding comments would help, but it probably better to start using function at this point.
- The current code or future function will look quite look-alike, and future changes to one may not be replicated to the other. Instead of copy/pasting the install locations ([uninstall](https://github.com/capnajax/bats-core/blob/b3522eced3a9d88ef685499be906e066d29ed9da/install.sh#L89-L92), [install](https://github.com/capnajax/bats-core/blob/b3522eced3a9d88ef685499be906e066d29ed9da/install.sh#L131-L134)), having a common list from which both function do things would make it harder for a future new folder to be installed and not uninstalled and vice-versa.

gtsop:
Hello, are there any updates on this? It is a useful feature

masatake:
Duplicated with https://github.com/bats-core/bats-core/pull/400 ?

martin-schulze-vireso:
Since this is outdated I am in favor of #400. We should compare and see whether there are nuggets in here that should be ported into that.

PR #51, opened 2018-01-01
Update .gitignore

dschaper:
A `.gitignore` file of `*` in the same directory that is the target to being ignored will recursively ignore the `.gitignore`. 

Changing the glob to `[^.gitignore]*` will ignore all files that are not in the regex `[.gitignore]`. Keeps IDE's like IDEA from complaining that you're tracking an ignored file.


tterranigma:
Seems good to me. Why not add a test that checks it too?

mbland:
I think a better expression may be:

```
*
!.gitignore
```

Can you give that a spin and see how it works? I'm concerned that what `[^.gitignore]*` actually does is ignore any file that doesn't start with `.`, `g`, `i`, `t`, etc.

jasonkarns:
Is there a reason we need the test/tmp dir to exist at all? I'm personally strongly in favor of centralizing the ignore rules in `/.gitignore` rather than having them dispersed throughout the codebase.

Bonus: `/test/tmp` in the root gitignore, will solve this issue of a versioned file being matched by the ignore rules.

mbland:
It's used in the test for `setup` and `teardown`. We could update the tests to do something differently. (I've a few minor cleanup ideas; this could be one of them.)

jasonkarns:
Perhaps we make the test robust to create it as necessary instead of versioning empty placeholder? That's typically my preference anyway, instead of committing cruft that are tertiary to the repo's purpose. (And eventually succumb to tragedy of the commons.)

Issue #44, opened 2017-12-06
Incorporate bats extension projects


tterranigma:
I recently came across `bats-support` and `bats-assert`. Maybe there are others too. It seems to me that there are useful features in these projects and there is the danger that they become obsolete due to lack of maintenance. 

I would like to suggest merging such extensions directly into bats due to the following advantages:
- ensured maintenance side-by-side core bats
- same quality of code as core bats
- easy discovery by users
- better overview of project and its dynamics, instead of having to keep track of several git repos.

Has there been past discussion on this?

psteinb:
see also the discussion in #40 ... this merge is needed from my point of view

mbland:
I'm open to the possibility—I've done a bunch of my own work in https://github.com/mbland/go-script-bats (see `lib/bats`) that could be merged with some of the existing libraries into a standard assertion library.

That said, the trend across languages lately appears to be to provide assertion libraries separately from frameworks. Would like to hear more perspectives, particularly from the other @bats-core/bats-core members.

jasonkarns:
Not incidentally, the two people involved in those projects are @ztombol and myself :D 
And @ztombol has been extended an invitation to join bats-core as a maintainer (#4) so I would fully expect these projects to be folded in at some point.

BraisGabin:
I think that the framework and the assertion libraries are two different things. So there's no need to merge them.

We just need to point of those libraries from this one in the documentation. And **maybe** later create a meta-package that includes all of them.

sublimino:
As we already have assertions in core I'd be tempted to support this proposal and merge the better assertion libraries into core.

Assertion libraries tend to provide different styles of assertion (e.g. http://chaijs.com/guide/styles/ for `assert`, `expect`, and `should`). 

How many styles do we expect bats to have? AFAICS it's a single style, `assert`: e.g. `<function> <...args>`. `should`'s fluent interface is not possible in BASH, and `expect` is possible to imitate, but it would be messy without OO.

(It might be worth comparing some other languages here)

The assertion libraries listed here significantly improve the usability of bats. I'd prefer to carefully merge and fold the existing libraries into core, so core's good enough that it doesn't need extending any more.

szaydel:
[...]

PR #54, opened 2018-01-09
WIP: Add CppUnit XML as an output option

Ferruck:
CppUnit is a widely used framework which outputs its own XML format. In case you're using it for nearly all of your test, you don't like to add another plugin to your CI server just to analyze one project outputting TAP. Thus, give Bats the possibility to output CppUnit XML.

I've tested this as good as I could, with some real-world input and some faked Bats files and compared the output with real-world CppUnit XML. All seems good, there may be edge cases, though.

### DIFF
--- a/libexec/bats
+++ b/libexec/bats
@@ -20,6 +20,7 @@ help() {
   echo "  -h, --help     Display this help message"
   echo "  -p, --pretty   Show results in pretty format (default for terminals)"
   echo "  -t, --tap      Show results in TAP format"
+  echo "  -u, --cppunit  Show results in CppUnit XML format"
   echo "  -v, --version  Display the version number"
   echo
   echo "  For more information, see https://github.com/bats-core/bats-core"
@@ -102,6 +103,7 @@ done
 unset count_flag pretty
 count_flag=''
 pretty=''
+cppunit=''
 [ -t 0 ] && [ -t 1 ] && pretty="1"
 [ -n "${CI:-}" ] && pretty=""
 
@@ -124,6 +126,9 @@ if [[ "${#options[@]}" -ne '0' ]]; then
       ;;
     "p" | "pretty" )
       pretty="1"
+      ;;
+    "u" | "cppunit" )
+      cppunit="1"
       ;;
     * )
       usage >&2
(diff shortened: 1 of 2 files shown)


jasonkarns:
Thanks for the PR!!

Unfortunately, I'm mostly -1 to adding a host of different output formats, considering the TAP output can just be piped to a tap-to-xunit formatter.

Is there a use case that isn't covered this way?

Ferruck:
Hi,

> Is there a use case that isn't covered this way?

No, not really. It's just that any converter I've found (none of them has been for CppUnit XML, though, and as far I can tell there are some differences) has some dependencies or needs some kind of runtime that isn't available in the build environment in use. Thus, I had to implement my own nonetheless and as the way the `pretty` formatter is used is highly modular, I've implemented it as a kind of plugin.

I still like the idea of Bats being able to output multiple (the most important) formats itself. But I understand the Unix philosophy, as well.

Therefore, it's OK if you don't like the idea. I just wanted to publish my outcome to the public.

tterranigma:
I think it would be great to allow for bats 'addons'. This formatter could be an example of formatter addons. Instead of having `-u` in the same manner as `-t`, `-p`, there could be a `--formatter=` option to choose arbitrary formatters. 

There is also discussion [here](#44) on whether to have assertion libraries incorporated to bats, it might become relevant.

Ferruck:
> there could be a `--formatter=` option to choose arbitrary formatters.

That sounds like a reasonable idea. Sadly your link points to this discussion here.

Please let me know when the interest gets real, I think this change needs some love (tests with a lot of error output have shown that the runtime in these cases are, well, not optimal).

tterranigma:
@Ferruck Strange, the issue is here: https://github.com/bats-core/bats-core/pull/44

martin-schulze-vireso:
I am closing this to keep the PR list clean. The code has moved on quite a bit so a rewrite will be necessary. Instead of clogging up the PR list, I am going to close this and linked it in a newly created issue about how to deal with new output formats. In the end I hope your code is not in vain but just delayed a bit longer. Thanks for you contribution and patience. :pray:

PR #55, opened 2018-01-09
Fix #53: install shouldn't change foreign files

### LINKED ISSUE #53: install.sh needs to not change permissions on files it didn't install
Currently `install.sh` changes permissions on everything in `$PREFIX/{bin,libexec}`, which is bad if installing to /usr/local.

tpraxl:
Fixes issue #53 by only changing the execution bit of target files that exist in the src folder (`$BATS_ROOT`), thus only touching files that have been copied during the installation.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -19,6 +19,42 @@ abs_dirname() {
   cd "$cwd"
 }
 
+# performs chmod a+x for all files in $target/$folder 
+# that exist in $src/$folder
+# 
+# Example:
+#     fix_folder_file_permissions $src $target $folder
+#
+fix_folder_file_permissions() {
+  src=$1
+  target=$2
+  folder=$3
+  # scan $src directory for the filenames to fix in $target
+  directory="${src}/${folder}"
+  for f in "${directory}"/*; do
+    # get basename of file (strip the path)
+    filename="${f##*/}"
+    # change the execution bit for the file in $target
+    chmod a+x "${target}/${folder}/$filename"
+  done
+}
+
+# executes fix_folder_file_permissions for the folders
+# bin and libexec.
+#
+# Example: 
+#     fix_installed_file_permissions "${BATS_ROOT}" "${PREFIX}"
+# 
+fix_installed_file_permissions() {
+  src=$1
+  target=$2
+  # we only need to adapt execution bits for bin and libexec
+  folders=(bin libexec)
+  for folder in ${folders[@]}; do
+    fix_folder_file_permissions "${src}" "${target}" "${folder}"
+  done
+}
+
 PREFIX="$1"
 if [ -z "$1" ]; then
   { echo "usage: $0 <prefix>"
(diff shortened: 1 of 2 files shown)


tpraxl:
Manual test of this PR is simple: 

In the project root:

```bash
$ mkdir -p tmp/bin && touch tmp/bin/file-without-x-flag && ./install.sh ./tmp && la tmp/bin
```

produces the following output:
```bash
Installed Bats to ./tmp/bin/bats
total 12K
lrwxrwxrwx 1 thomas thomas 15 Jan 10 17:18 bats -> ../libexec/bats
-rw-rw-r-- 1 thomas thomas  0 Jan 10 17:18 file-without-x-flag
```
Before this PR, it would have created the following output

```bash
Installed Bats to ./tmp/bin/bats
total 12K
lrwxrwxrwx 1 thomas thomas 15 Jan 10 17:18 bats -> ../libexec/bats
-rwxrwxr-x 1 thomas thomas  0 Jan 10 17:18 file-without-x-flag
```
So this PR prevents touching pre-existing files in the installation folder.

install.sh
```diff
@@ -19,6 +19,42 @@ abs_dirname() {
...
+#
+# Example: 
+#     fix_installed_file_permissions "${BATS_ROOT}" "${PREFIX}"
+# 
+fix_installed_file_permissions() {
+  src=$1
+  target=$2
+  # we only need to adapt execution bits for bin and libexec
+  folders=(bin libexec)
+  for folder in ${folders[@]}; do
```
tterranigma: Wouldn't be nice to have all variables wrapped in double quotes?

install.sh
```diff
@@ -19,6 +19,42 @@ abs_dirname() {
...
+  target=$2
+  folder=$3
+  # scan $src directory for the filenames to fix in $target
+  directory="${src}/${folder}"
+  for f in "${directory}"/*; do
+    # get basename of file (strip the path)
+    filename="${f##*/}"
+    # change the execution bit for the file in $target
+    chmod a+x "${target}/${folder}/$filename"
+  done
```
tterranigma: Stylistically, I think it is better to skip `{}` around variable names when not really needed.

tterranigma:
Just saw your #56. Why don't you close this PR?

tpraxl:
@tterranigma thanks for your suggestions. I'll consider the changes as soon as there's time for it. 

> Just saw your #56. Why don't you close this PR?

#56 requires an adaption of the dockerfile. I thought it would be better to split #55 and #56 up, just to get #55 merged in quickly as it fixes something you really don't want. I didn't want to risk #55 not being merged.

install.sh
mbland: Not just nice, but imperative.

install.sh
```diff
@@ -19,6 +19,42 @@ abs_dirname() {
...
 }
 
+# performs chmod a+x for all files in $target/$folder 
+# that exist in $src/$folder
+# 
+# Example:
+#     fix_folder_file_permissions $src $target $folder
+#
+fix_folder_file_permissions() {
+  src=$1
```
mbland: Label all local variables with `local`, e.g. `local src="$1"`.

Issue #28, opened 2017-10-09
Deprecating the focus on posix sh compatibility


ntnn:
The maintainers of sstephenson/bats made it clear that they wanted to use as much posix as possible.

While this makes sense in some cases (and would entirely be possible), I think that it doesn't fit the projects' scope.

1. BATS stands for *bash* automated testing system
2. The codebase uses features such as `pipefail`, which are not part of the POSIX sh spec (and likely won't be for the near future)

A negligible point is the cost of the additional calls to external programs, as modern machines don't have a problem with those (`[` e.g. was an alias to `test` until recently in bash, instead of `=~` `grep` or `expr` have to be used).

In general bash is more friendly to develop and read as a posix compliant `test`, e.g., does not allow for compounds like `[ 1 -eq 0 || 0 -eq 0 ]`, which is a common construct to use.

I'd like a clear statement if this project intents to go full posix or full bash. I'd be fine with either way, but currently it is a mix.

mbland:
I'm in the full-Bash camp when it comes to the implementation details and what we support in the test syntax. There's already tons of Bash-isms throughout the code that existed from the beginning (i.e. before my updates in #8), and #24 unlocks double-bracket capability.

That said, for whatever external programs we might employ, ensuring POSIX compatibility with _those_ would be good practice—eliminating the need for them altogether via an equivalent Bash function or script implementation would be even better, from a portability and (usually) performance point of view. See #9 and the [dup-warning-#9 branch](https://github.com/bats-core/bats-core/tree/dup-warning-%239) linked from it for an example.

So, FWIW, there's the beginning of a statement, but I'd like to hear other @bats-core/bats-core members (currently @btamayo and @jasonkarns) chime in as well.

ntnn:
On that notion I'd like to link to the [tools reference](https://devmanual.gentoo.org/tools-reference/index.html) of the gentoo dev manual, it largely focuses on POSIX (if it doesn't it is usually marked, e.g. as it is for `find`) and a bit friendlier to navigate then the [OpenGroup Pubs](http://pubs.opengroup.org/onlinepubs/9699919799/).

sublimino:
Great point @ntnn! I can venture an opinion: 

[...]

PR #57, opened 2018-01-14
Fix posix incorrectness

wilhelmtell:
Two things break Bats on my machine: Bash identifiers that include a dash, and process substitution. These two features are Bash POSIX extensions.

The two commits in this pull-request remove these dependencies. The first commit removes the dashes from generated test names, and the second commit uses command-substitution and input-redirection for the process-substitution effect.

### DIFF
--- a/libexec/bats-preprocess
+++ b/libexec/bats-preprocess
@@ -7,8 +7,8 @@ encode_name() {
   local hex_code
 
   if [[ ! "$name" =~ [^[:alnum:]\ _-] ]]; then
-    name="${name//_/-5f}"
-    name="${name//-/-2d}"
+    name="${name//_/5f}"
+    name="${name//-/2d}"
     name="${name// /_}"
     result+="$name"
   else
@@ -22,7 +22,7 @@ encode_name() {
       elif [[ "$char" =~ [[:alnum:]] ]]; then
         result+="$char"
       else
-        printf -v 'hex_code' -- "-%02x" \'"$char"
+        printf -v 'hex_code' -- "%02x" \'"$char"
         result+="$hex_code"
       fi
     done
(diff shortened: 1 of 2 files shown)


wilhelmtell:
I notice the first commit, that removes the dashes, breaks the AppVeyor build, in test 38, which the output says is test/bats.bats but that seems to me coming from test/fixtures/bats/single_line.bats.

These dashes originates from the preprocessing stage that translates test descriptions to test names. Why are the dashes there in test names?

wilhelmtell:
Is it possible that AppVeyor's test 38 fails because it runs on msys and has no /dev/stdin? I used the `$(</dev/stdin)` trick because the Bash manual says it's a faster `cat`. Would replacing that with `cat` fix the issue?

I'm not sure it's right for me to "pollute" this pull request with tryouts until I get this resolved. What's the proper way to go about this, when the automated build fails on a pull-request?

wilhelmtell:
I forgot to mention: I'm running on a Mac with OSX 10.11.6. Needless to say, all the bats tests pass with flying colours on my machine.

tterranigma:
What is the original problem you are trying to fix? Are you using a shell other than bash?

tterranigma:
Shouldn't this be closed as invalid?

mbland:
Delayed reaction here... What are the specs of the machine on which Bats breaks? Bats contains a lot of Bash-isms; I'm surprised that these two items would be the only source of breakage.

I am tempted to close this as invalid, as POSIX compatibility isn't a goal for this incarnation of Bats; see #28.

martin-schulze-vireso:
As already said: bats-core is targeted at bash. I think there is no going back now.

Issue #39, opened 2017-11-19
Run some code only once before any test (and probably other after)


BraisGabin:
I'm writing some tests with bats to test my docker image.

All my tests depends on the same image so I have two options:

* Build the image before every test (using `setup()`). Docker use the cache and it is not that slow... but it's slower that do nothing. And the output is poluted.
* Remeber to run the `docker build` before I run bats... this is much faster but it force me to remeber to launch the `docker build` command EVERY time. At the end of the day this is even slower because I forgot to do it and I need to launch bats twice.

It'd be great to define two new functions and call them before and after all the tests run.

BraisGabin:
I'm looking for something similar to [`@BeforeAll`](http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/BeforeAll.html) in JUnit5 (`@BeforeClass` in JUnit4)

andrewfowlie:
I agree that a global teardown and a global setup would be helpful.

mbland:
I'm interested in this as well, possibly for 1.0.

BraisGabin:
How could we call this two new methods? I did a little research:

|options|Once per file (before)|Once per test (before)|Once per test (after)|Once per file (after)|
|-----|-----|-----|-----|-----|
|bats|-|setup|teardown|-|
|Python|setUpClass|setUp|tearDown|tearDownClass|
|JUnit5|beforeAll|before|after|afterAll|
|node js|before|beforeEach|afterEach|after|

The most conservative way would be to use `setupFile` and `teardownFile` (similar to Python). But I think that the JUnit5 or node js names are easier to understand. Any thoughts about this?

andrewfowlie:
Personally, I'd favour lowercase separated with underscores and don't find adding `file` to function name that informative. I'd favour e.g, `initial_setup` and `final_teardown`

jsok:
I've implemented this behaviour in many projects and named the functions `{setup,teardown}_suite`.

That's consistent with the bats nomenclature right?

extsoft:
I suppose the better names will be `before_all` and `after_all`. IMXO, these names will explain the functionality in the best way.

szaydel:
I actually added this to address this need, but it is surely not the way to go. I think `before_all` and `after_all` would be good names and ideally be supported whether only a single file is being processed or a directory with one or more .bats files. What I have done here does not even work with single files.
```
diff --git c/libexec/bats w/libexec/bats
[...]

PR #58, opened 2018-01-31
Adds a BATS_BUCKET exported variable for test inter-dependency

nkakouros:
Regardless of whether test inter-dependency should be discouraged or not, there are times that you need to share info between tests. It is also useful to avoid rerunning expensive operations more than once and for keeping test-code repetition low. Right now there is no way in BATS for tests to communicate without messing around with writes to files. Every test is evaluated in its own subprocess and scope.

As an example from other frameworks, PHPUnit supports test dependency by passing the return value of a dependee test as argument to the dependent test.

This PR is a much simpler approach and adds a variable that is shared among tests. It is meant to act as a 'bucket' where test (and the setup/teardown) functions can put and access info. The variable is initialized as a string but can be assigned an array, etc.

My implementation seems a little hacky to me. But I could not come up with any other implementation that does not involve writing to files.

Specifics for my use case
---
In the README it is mentioned that BATS 

>  ... provides a simple way to verify that the UNIX programs you write behave as expected.

My application is one that involves setting up remote machines and configuring them to act as pentest targets. This requires that, apart from local functions, I also test services, brute-forceability, etc. Some of these stuff take a lot of time to perform. An example is a series of tests that:

1. check if a service is running
2. check if the login function is accessible
3. check if the login is bruteforceable
4. check if a specific value is returned after login
and so on...

In this case, test 4 requires a username/password (which cannot be hardcoded as it is auto generated at server creation time). Rerunning the bruteforce operation a second time in test 4 is pretty time consuming and redundant as the username/password were already retrieved.


tterranigma:
My first attempt with this was simply naive and not thoroughly tested. I have updated the PR and have edited the original PR comment to reflect the current situation. I would appreciate some thoughts on this.

mbland:
I'm really against the idea of test case interdependency, and believe the perceived need is better served by a test redesign. I could be wrong, but that's my perspective based on my own experiences.

Any resources/initial state that should be shared between test cases (particularly state that's expensive or complicated to set up/tear down) would be handled by the resolution to #39; that I see as legitimate. But an individual test case should be able to run without depending on the result of previous test cases, and should theoretically produce the same results whether run alone or in random order relative to other test cases. The initial state of one shouldn't depend on the final state of another.

Regarding the "service running/login available/etc." test cases mentioned above, I'd consider having the service running would be a precondition for all other cases (i.e. `setup_all()`), which could then execute independently from one another. In other words, brute force-ability sounds like an orthogonal concern to getting an expected value after login, and both cases should run regardless of the result of the other.

I'm still open to contrary arguments, but it'll take a lot to convince me that an interdependency mechanism is warranted.

tterranigma:
Then, I 'll try my best to convince you that adding support for test dependency is sth worth having! 

## Opinions are a private matter, tools are a common resource
I have my own views and you have yours regarding whether test interdependency is good to implement or not, but the scope of these views I believe should be kept at individual project level. Bats is a tool that a lot of people use with conflicting views and different needs. To make a parallel that quickly comes to mind while keeping it in the IT realm, enforcing one view is similar to saying that cursor navigation in vim should only happen with `hjkl` and not with arrows keys, thus arrows keys are hard disabled from performing this action ever. It is limiting to Bats' utility to the developer world to purposefully perceive and design Bats to support only the "true unit-testers", the crusaders of clear-cut testing. What about the rest of us that get depressed that our life cannot become better because we do not fit the ideal?

## More people asking about this
There seem to be more people and use cases that need for test interdependency in Bats, specifically #86 . While not the optimal solution for every use case, the solution in my PR is flexible enough to accommodate for different needs. For instance, `BATS_BUCKET` could hold the success status of a test so that it can be read by subsequent ones, thus resolving #86. Maybe #86 gets a different, specific solution, but still the argument holds that the implementation here is flexible enough to cater for many use cases, mainly because, conceptually, it is just really a _bucket_ for arbitrary info to be passed **among** (not just between) tests.

## It is already possible
It is now already possible to have test interdependency in Bats by ...writing/reading to/from files. Still, this is not a proper or viable solution. First, it can generate a lot of garbage files in one's project. It also adds a burden for the test developer to keep track of files, delete them before each test run, etc. But more importantly it is slow as hell. With all the efforts (that you spearheaded) to make Bats faster and faster, it seems inappropriate to have to rely on sth like this. In reality, what this PR does is make life a little easier for a few of us.

## It is a matter of speed and effectiveness
On a related note, my use cases for needing this functionality is to actually save huge amounts of time. I have tests that for instance brute force a service or that need to run mimikatz for username and password hashes extraction on windows instances. These take minutes to complete. What are my options? One option is to run everything in a single test function that, for instance checks, that the service is bruteforceable, checks that the retrieved username is correct, checks that the same username/password is valid for a different service, checks that this password produces the correct hash that is located in /etc/shadow in another instance, etc. But this is contrary to the "ideal" of unit testing that a test should only test specific, limited functionality. The other alternative is to rerun every step incrementally. For instance, the 1st test checks for bruteforceability, the 2nd checks for bruteforceability plus validity of user/pass, etc. But thiss is a waste of time and it also, semantically, adds test dependency. If the first fails, the rest are bound to fail. Having the ability to pass the retrieved username and password around is a blessing. I can get as fast as possible and have results that pinpoint the exact point of failure.

## Bats does not equal unit testing
One argument against the above could be that what I am doing is basically using the wrong tool for my use case, ie unit testing when what I should be aiming for is integration testing (or not... I never fully understood what each type of testing stands for). But this is not really true. First, Bats' README does not even mention the phrase `unit testing`. Second, there is nothing inherent to what Bats does that says it should be limited to idealized, pure unit testing. 

## Theory does always work in practice
Another point to mention has to do with the philosophy of writing tests per se. As I understand it, the reasoning behind independent tests is that a test should relate to as little and as isolated functionality as possible in order to know exactly what the point of failure is (in case of actual failure). But, if I am correct, this is almost never the case. Take for instance your project, go-script-bash. If you go and arbitrarily introduce an error in some random piece of code, chances are that more than one tests will fail. This reveals that while a test might try to address a specific piece of code, the test suite as a whole fails to treat code functionality as an amalgam of isolated units of code. And that is super fine. It works and it does so in a useful and sufficient way. In other words, the ideal case is nice to have as a direction but treating it as a the goal is more of a tyranny, it does not add actual value.

## Practice always works in theory
One point you make is that it should be possible to run tests in any order and still have them report reliable results. First, there is no support for that in Bats. Tests in a test file are always run from top to bottom and you cannot specify a specific one. Second, this assumption should be left up to the test designer. Developers have to make decisions for their projects for eg where to put bootstrap code, in `setup()`, in the beginning of the test function or outside of functions altogether, OR how to split tests among files OR how to name them. They are also free (and able) to make (truly, really) bad decisions like testing in one test function how a library function behaves in all possible cases (bad input, no input, return codes, whatever). This is not a good test. Should Bats have (or not have) functionality that preaches the true way to the developer?

## Other ecosystems have already support for test interdependency
Looking at the "literature" is also handy in seeing how other ecosystems have built their solutions. A quick search reveals that PHPUnit (for PHP) and unittest (for Python) support test dependencies. JUnit (for Java) does not but there is a fork TestNG that does.(I would really not like to see a fork of Bats just to support test dependencies). And so on. This means that there is a need for that functionality.

To conclude, if I had only one sentence to write, it would be that flexibility (without complexity) makes tools powerful and enables people to work with their issues at hand without the need to first solve the problems of the tool or without the need to re-invent the wheel. In any case, it would be interesting to see what others have to say (preferably on specific points, not simply expressing preference).

tterranigma:
To add one more dimension to the discussion, I will try to present some questions about when a proposed change/addition should be considered for merging.

- Is the change adding heavy maintenance cost?
  - I don't think this is true here. The PR adds ~25 new lines of code that are quite easy to understand (I think) and Bats itself is a quite small project still.

- Is the change adding value to the project?
  - I think yes as it adds new functionality (I do not consider using files as I mentioned in the previous comment to be a feature), it makes the tool more flexible and thus more use cases can be supported.

- Is it sth the project can live without?
  - Sure.

- Is the PR introducing bad/under-thought/conflicting changes that in the future will create problems for the users or developers (stemming from the tool itself, not how it is used by users)?
  - As I explained in my previous comment I think not. There is no strange behavior introduced to the code that would baffle developers and users have to explicitly opt in to use the feature (by using the `BATS_BUCKET` variable), thus unexpected behavior should not be an issue.

- Is there a better solution available (if we accept that the issue exists)?
  - I couldn't think of any.

- Does the change conflict the paradigm the tool is striving to support?
  - I don't think there is such an explicit paradigm. Moreover, if we accept unit testing as the paradigm Bats supports, test interdependency has to do with an idealistic vs pragmatic approach in the implementation. I think.

tterranigma:
With the recent activity here, could I hope for someone looking into this?

martin-schulze-vireso:
Hey @nkakouros, I am currently working through the backlog of issues and PRs and saw this lying around. While I am skeptical of your approach I still am curious about the use case and potential interaction with other features. Could you add an example of code that uses your implementation at #86 and describe why it needs this feature?

> I have tests that for instance brute force a service or that need to run mimikatz for username and password hash extraction on windows instances. These take minutes to complete.

Would it be possible with the current `setup_file` interface to pull that step up there and rely on it in each individual test?

PR #59, opened 2018-03-13
Moves argument check to entry script

tterranigma:
Just tidying up.

### DIFF
--- a/libexec/bats-exec-test
+++ b/libexec/bats-exec-test
@@ -16,15 +16,7 @@ if [ "$1" = "-x" ]; then
 fi
 
 BATS_TEST_FILENAME="$1"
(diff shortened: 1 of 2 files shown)


libexec/bats-exec-test
```diff
@@ -16,15 +16,7 @@ if [ "$1" = "-x" ]; then
 fi
 
 BATS_TEST_FILENAME="$1"
-if [ -z "$BATS_TEST_FILENAME" ]; then
```
mbland: I'm in favor of keeping this as a defense mechanism against bats-exec-test getting called incorrectly.

tterranigma: Do you mean `bats-exec-test` could be called directly?

PR #60, opened 2018-03-15
Want basic logging support for logging output and other side-effects

szaydel:
I am not sure if there's any interest in this change, or something remotely close to it, but I found that I really lacked having even basic logging. This qualifies as perhaps really basic logging, and if there's interest and feedback, I could spend some time on improving it. This is less a PR and more a request for comment I suppose.

Thanks in advance!

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -14,6 +14,7 @@ abort() {
   printf 'Error: %s\n' "$1" >&2
   usage >&2
   exit 1
+  echo "Usage: bats [-c] [-l <logfile>] [-p | -t] <test> [<test> ...]"
 }
 
 help() {
@@ -28,6 +29,7 @@ help() {
 
   -c, --count      Count the number of test cases without running any tests
   -h, --help       Display this help message
+  -l  --log      Set path to log file used by all tests with logit calls
   -p, --pretty     Show results in pretty format (default for terminals)
   -r, --recursive  Include tests in subdirectories
   -t, --tap        Show results in TAP format
@@ -77,8 +79,12 @@ for arg in "$@"; do
     arguments[${#arguments[*]}]="$arg"
   fi
 done
+<<<<<<< HEAD
 
 unset count_flag pretty recursive
+=======
+unset count_flag pretty logfile
+>>>>>>> Want basic logging support for logging output and other side-effects
 count_flag=''
 pretty=''
 recursive=''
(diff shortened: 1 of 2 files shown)


szaydel:
Glad test caught a problem, will fix.

mbland:
Not sure I understand the use case. Care to expand further? Is this to emit output while debugging test failures?

szaydel:
Unless I am missing something very obvious, I don't see a way to generate logging from the tests. In other words sometimes I want to log certain outputs, sometimes I want to log tests, etc., with a timestamp and some context, like part of this test, part of this specific run, etc. Relative time may be more useful than absolute time, mainly so that one could tell how long things take, in particular if some parts of the test suite are potentially long-running. Does this make sense @mbland?

tterranigma:
Personally, I would be much more in favor of an option to enable general debug output. If I understand correctly, this PR provides a `logit` function that can be used selectively by the user. Instead, it would be more transparent and general to log all "background" output to a file (or stdin even).

szaydel:
I am not sure I follow why anything would go to `stdin`, but I don't disagree, until one notices that some programs generate profuse amounts of output and it is not necessarily something that one always wants to collect. I think cases vary enough between what I need and what you may be observing that we simply talking about solving slightly different problems. A more general option to "debug" log everything is not a bad idea, but I don't think that would address what I am looking to address. Having said, I am in favour of a more general debug logging nonetheless.

kakawait:
I've to maintain my own fork to get that essential feature. As @szaydel, I may miss something but when running tests (most of time if it fails but not always) I would see program (that I'm testing) output in order to:

1. Debug when it fails
2. Help writing code when TDD to check that output is evolved in good way you want

Whatever the option `--log <PATH>` or as @tterranigma said a `debug flag` to display output (but I would like to redirect file, if possible, to debug much more easily).

vanosg:
A little late to the game, but I'd still love to see some form of debug logging implemented (for me, a log file of everything intended for stdout, to assist in determining at what point something went wrong during review). Is there still any maintainer interest in something like this PR, or has it already been merged elsewhere? Thanks

martin-schulze-vireso:
There are several issues with the code in this PR, so I will focus more on the general idea: The stated motivation is to log the output of commands for debugging purposes.

> In other words sometimes I want to log certain outputs, sometimes I want to log tests, etc., with a timestamp and some context, like part of this test, part of this specific run, etc. 

I am not sure what `log tests` means but I think this sounds like the trace feature (#137) plus printing output.

> Relative time may be more useful than absolute time, mainly so that one could tell how long things take, in particular if some parts of the test suite are potentially long-running.

On a per test level we already have this with the timing feature now.

The code issues, plus the distance to current master don't make this PR viable right now. I would propose you add an issue about this where we can flesh out the use case and what is really needed. I have a feeling that there won't be much left with the timing being already landed and a trace feature pending.

PR #62, opened 2018-03-20
More support for strict mode

nkakouros:

I do not know why I got some more unbound variables compared to #26 . My approach was to 'set -euo pipefail' in all `libexec/` files and see what needs fixing. Commit 157ec232d95 is there just to show where the `set -euo pipefail` line was added and is not meant to be left as is. 

I am not sure how to properly solve the failing test, especially as #59 is pending.

### DIFF
--- a/bin/bats
+++ b/bin/bats
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -euo pipefail
 
 BATS_READLINK='true'
 if command -v 'greadlink' >/dev/null; then
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -euo pipefail
 
 export BATS_VERSION='1.1.0'
 
--- a/libexec/bats-core/bats-exec-suite
+++ b/libexec/bats-core/bats-exec-suite
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -euo pipefail
 
 count_only_flag=''
 extended_syntax_flag=''
--- a/libexec/bats-core/bats-exec-test
+++ b/libexec/bats-core/bats-exec-test
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -eET
+set -eETuo pipefail
 
 BATS_COUNT_ONLY=''
 BATS_TEST_FILTER=''
@@ -330,10 +330,6 @@ bats_perform_test() {
   BATS_TEST_NAME="$1"
   if declare -F "$BATS_TEST_NAME" >/dev/null; then
     BATS_TEST_NUMBER="$2"
-    if [[ -z "$BATS_TEST_NUMBER" ]]; then
-      printf '1..1\n'
-      BATS_TEST_NUMBER=1
-    fi
 
     # Some versions of Bash will reset BASH_LINENO to the first line of the
     # function when the ERR trap fires. All versions of Bash appear to reset it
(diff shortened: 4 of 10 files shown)


bvberkum:
I thought I'd drop in and try to do some analysis. I'm no Bash expert and this is one of the reasons. 

Let me set aside `set -o execfail` first, it gives an error with the Alpine/Busybox/Bash 4.3. And I cant find much about it [1], but I am also not sure why the `exec` later on is needed either.

The source of the issue ``filenames[@]: unbound variable`` [2]. Ofcourse the array is not filled for test "running a suite with no test files", `set -u` takes this to take that "$filenames[@]" expands to undefined. 

I've seen code branch here at ``$# -eq 0`` iow. duplicate the lines and add an if-statement. And call the function twice. But that may leave the routine to handle an undefined '$1', etc.

For what it is worth, this is why I personally shy away from Bash or arrays in shell scripting.
Ie. '$1: unbound variable' seems really silly to me. Ofcourse the proper way to expand to empty is adding the magic hyphen '-'. But that gives problems e.g. in a for. 

I tried fixing this strict-mode on my own branch [3], and ended up no longer supporting the empty-suite test case and I don't know if that is desirable. My vote would be to let that go, but I'm not sure wether at the plumbing level something should be spec'ed or tested so that empty-suite streams can still be generated outside of the CLI.

I also tried to take the other advice `IFS=$'\n\t'` from [4] but I need some more time with that too. 

[1] http://wiki.bash-hackers.org/internals/shell_options#execfail
[2] https://travis-ci.org/bvberkum/bats-core/jobs/356652724
[3] https://travis-ci.org/bvberkum/bats-core/jobs/356664357
[4] http://redsymbol.net/articles/unofficial-bash-strict-mode/

tterranigma:
Let me, first, explain the purpose of this PR. This PR does not __solve__ anything. But I think it is important to test bash programs with strict mode enabled and the reason is that strict mode highlights ambiguities in the code. Even if Bats does not operate in strict mode by default (my personal view is that it should), it is important that the code is structured as if it were enabled. Without strict mode what essentially happens is that you write a program, you run it (and test it even) and because it works you think that your assumptions in the code were all correct. 

For Bats in particular, the code needs some mental gymnastics for contributor that are not so well versed with bash. Having the code handle edge cases explicitly (eg what happens if no filename is provided or the test suite is empty), will make it both easier for new contributors and prevent errors. I should also mention as an argument for having strict mode enabled by default that, as Bats runs tests in non-posix subshells, the errexit option is not inherited and, thus, user tests run without it. 

The failing test can be solved one way or another. I did not want to spend more time as there is also #59 is pending.

ddnomad:
Just tried to use BATS in Travis to test my (early stage) bash project. First build has failed and I presume it was because docker runs everything by default using `/bin/sh`. Here is an original make lines causing trouble: 

```
.PHONY: docker-test
docker-test: docker-build
	docker run --rm -it  $(DOCKER_TAG) make install && make test
```

I'm testing a little bash utility script that do have `set -euo pipefail` in it. Originally I was getting an unbound variable in BATS sources and everything went bananas. After changing the above make file lines to the below I've managed to get rid of a problem (build passes):

```
.PHONY: docker-test
docker-test: docker-build
	docker run --rm -it  $(DOCKER_TAG) /bin/bash -c "make install && make test"
```

I'll see how it will go in a future though. Source (if somebody interested): https://github.com/ddnomad/wtd

ddnomad:
@bvberkum well, might be. The thing is that tests are passing with this fix. Originally they were failing with an unbound variable in BATS source. 

So it did the trick after all.

tterranigma:
I replaced the previous commits with a fresh take.

Issue #43, opened 2017-12-05
Document significance of file descriptor 3


mbland:
This is a carry-over from sstephenson/bats#80, and also discussed in mbland/go-script-bash#226 and mbland/go-script-bash#227.

Bats duplicates standard output as file descriptor 3 so that its framework functions will produce TAP output that isn't captured into `BATS_OUT`, i.e. the file used to capture output from the code under test. When the code under test spawns a child process that blocks with this file descriptor open, Bats will also block (i.e. hang) until that process unblocks and exits.

I don't think there's a general fix within Bats itself, other than to very clearly document this behavior. I tried setting `BATS_OUTPUT_FD=3` and using that to close the file descriptor in mbland/go-script-bash#227, but while it's possible to use `{BATS_OUTPUT_FD}>&-` in Bash 4.x, it doesn't appear possible in Bats 3.2.57(1)-release to use anything but a bare integer. Chapter and verse from https://tiswww.case.edu/php/chet/bash/CHANGES:

```
This document details the changes between this version, bash-4.1-alpha,
and the previous version, bash-4.0-release.

3.  New Features in Bash

p.  If the optional left-hand-side of a redirection is of the form {var}, the
    shell assigns the file descriptor used to $var or uses $var as the file
    descriptor to move or close, depending on the redirection operator.
```

We could still set and document `BATS_OUTPUT_FD=3`, and advise users that they must use `3>&-` in code that must run on 3.2.57(1)-release, but can use `{BATS_OUTPUT_FD}>&-` if running only on Bash 4.1-alpha and higher.

*Or...*, we might consider porting the `background-process` utilities into bats-core, or some such extra module. Thoughts?

tterranigma:
If you give `eval "exec $BATS_OUTPUT_FD>&-"` on bash 3, doesn't it close the descriptor?

mbland:
`eval` would do the trick, yes. We could offer that as an option.

Issue #64, opened 2018-03-31
Add function to print custom diagnostics


nkakouros:
According to the TAP spec, diagnostic lines should start with `#`. Otherwise, while not treated as an error, it is up to the _harness_ what it will do with it. Right now in Bats, if the user needs to output some text from within a test function and still get proper TAP output, they need to:

- redirect to FD 3
- prepend text with `#`

Eg, `echo "# $variable" >&3`

This seems to me like an unnecessary burden. It would be nice to have a function, perhaps called `bats_print_diagnostic()` that will print whatever arguments are passed to it in the proper manner.

PR #65, opened 2018-03-31
Document significance of file descriptor 3

### LINKED ISSUE #41: Code outside of setup/teardown
Right now I have a test file with some code in the setup function. If I remove the code outside of the setup function it still gets run and, output aside, the result is the same. Is there any essential difference between the two ways? I am asking because I want to print some text in the setup function but I cannot. Outside of the setup I can.

tterranigma:
Also, moves #43 a little forward.


README.md
```diff
@@ -214,6 +214,28 @@ you print in code outside of `@test`, `setup` or `teardown` functions
...
+
+A side-effect of using file descriptor 3 is that, under some circumstances, it can cause Bats to block and execution to seem dead without reason. This can happen if a child process is spawned in the background from a test. In this case, the child process will inherit file descriptor 3. Bats, as the parent process, will wait for the file descriptor to be closed by the child process before continuing execution. If the child process takes a lot of time to complete (eg if the child process is a `sleep 100` command or a background service that will run indefinitely), Bats will be similarly blocked for the same amount of time. To prevent this from happening, you need to close the file descriptor when calling the child process, eg `command_name 3>- &`.
+
+### Printing to the terminal
+
+Bats produces output compliant with [version 12 of the TAP protocol](https://testanything.org). The produced TAP stream is by default piped to a pretty formatter for human consumption, but if BATS is called with the `-t` flag, then the TAP stream is directly printed to the console.
```
sublimino: s/BATS/Bats/

README.md
```diff
@@ -214,6 +214,28 @@ you print in code outside of `@test`, `setup` or `teardown` functions
...
+
+### Printing to the terminal
+
+Bats produces output compliant with [version 12 of the TAP protocol](https://testanything.org). The produced TAP stream is by default piped to a pretty formatter for human consumption, but if BATS is called with the `-t` flag, then the TAP stream is directly printed to the console. 
+
+The above have implications if you try to print custom text to the terminal. As mentioned in [File descriptor 3](#file-descriptor-3), bats provides a special file descriptor, `&3`, that you should use to print your custom text. Here are some detailed guidelines to refer to:
```
sublimino: > The above have
Hard to read. Perhaps "This has" or "The above has" ?

And s/bats/Bats/

README.md
```diff
@@ -214,6 +214,28 @@ you print in code outside of `@test`, `setup` or `teardown` functions
...
+
+- Printing **from within a test function**:
+  - To have text printed from within a test function, you need to redirect the output to file descriptor 3, eg `echo 'text' >&3`. This output will become part of the TAP stream. You are encouraged to prepend text printed this way with a hash (eg `echo '# text' >&3`) in order to produce 100% TAP compliant output. Otherwise, depending on the 3rd-party tools you use to analyze the TAP stream, you can encounter unexpected behavior or errors. 
+  - The pretty formatter that Bats uses by default to process the TAP stream, will filter out and not print text output to file descriptor 3.
```
sublimino: Remove comma

README.md
```diff
@@ -214,6 +214,28 @@ you print in code outside of `@test`, `setup` or `teardown` functions
...
 cause Bats to fail by polluting the TAP stream on `stdout`.
 
+### File descriptor 3
+
+Bats makes a separation between output from the code under test and output that forms the TAP stream (which is produced by Bats internals). This is done in order to produce TAP-compliant output. In the [Printing to the terminal](#printing-to-the-terminal) section, there are details on how to use file descriptor 3 to print custom text properly. 
+
+A side-effect of using file descriptor 3 is that, under some circumstances, it can cause Bats to block and execution to seem dead without reason. This can happen if a child process is spawned in the background from a test. In this case, the child process will inherit file descriptor 3. Bats, as the parent process, will wait for the file descriptor to be closed by the child process before continuing execution. If the child process takes a lot of time to complete (eg if the child process is a `sleep 100` command or a background service that will run indefinitely), Bats will be similarly blocked for the same amount of time. To prevent this from happening, you need to close the file descriptor when calling the child process, eg `command_name 3>- &`.
```
mbland: Nit: s/side-effect/side effect/

mbland:
Closes #41. This declaration will cause the issue to close upon merge; not sure having it in the title of the issue works. Also, updated the PR to have the same name as #41 for ease of reading.

Still reviewing.

README.md
```diff
@@ -214,6 +214,28 @@ you print in code outside of `@test`, `setup` or `teardown` functions
 must be redirected to `stderr` (`>&2`). Otherwise, the output may
 cause Bats to fail by polluting the TAP stream on `stdout`.
 
+### File descriptor 3
```
mbland: Suggest this rewrite to keep from burying the lede:

```markdown
### File descriptor 3 (read this if Bats hangs)

Bats duplicates standard output to file descriptor 3 to separate its TAP-compliant output
(written to FD 3) from the standard output of the `run` command, which is redirected into a
temp file. A potential side effect of this mechanism is that long-running background
processes launched by the test case will cause Bats to block until they exit, since
they will inherit FD 3 without closing it. Depending on the background process, the test
may hang indefinitely until the process is killed.

**To prevent this from happening, close FD 3 explicitly when running any command that may
launch long-running child processes**, e.g. `command_name 3>- &`.
```

README.md
```diff
@@ -214,6 +214,28 @@ you print in code outside of `@test`, `setup` or `teardown` functions
...
+
+Bats makes a separation between output from the code under test and output that forms the TAP stream (which is produced by Bats internals). This is done in order to produce TAP-compliant output. In the [Printing to the terminal](#printing-to-the-terminal) section, there are details on how to use file descriptor 3 to print custom text properly. 
+
+A side-effect of using file descriptor 3 is that, under some circumstances, it can cause Bats to block and execution to seem dead without reason. This can happen if a child process is spawned in the background from a test. In this case, the child process will inherit file descriptor 3. Bats, as the parent process, will wait for the file descriptor to be closed by the child process before continuing execution. If the child process takes a lot of time to complete (eg if the child process is a `sleep 100` command or a background service that will run indefinitely), Bats will be similarly blocked for the same amount of time. To prevent this from happening, you need to close the file descriptor when calling the child process, eg `command_name 3>- &`.
+
+### Printing to the terminal
```
mbland: I don't think we actually want to encourage this. If there is a legitimate reason for users to inject output into the TAP stream, we should provide concrete examples. Otherwise, we should remove this section.

README.md
tterranigma: I changed the title and the final bold line.

README.md
tterranigma: @mbland TBH, I included this information for selfish reasons. I had often spent time going through the same code over and over again to understand exactly what goes on inside Bats and why, that I decided to put it down to paper. I often find myself in need to debug my test scripts and printing from different places is crucial. But your comment made be curious and I looked into the TAP specification, version 12. It mentions for lines starting with `#`:

> Additional information may be put into the testing output on separate lines. Diagnostic lines should begin with a #, which the harness must ignore, at least as far as analyzing the test results. The harness is free, however, to display the diagnostics.

And for arbitrary lines not starting with `#`:

> Any output line that is not a version, a plan, a test line, a diagnostic or a bail out is considered an “unknown” line. A TAP parser is required to not consider an unknown line as an error but may optionally choose to capture said line and hand it to the test harness, which may have custom behavior attached. This is to allow for forward compatability. 

Version 13 agrees on custom diagnostics (although it seems to deprecated them in favor of yaml blocks) but considers random lines without `#` an error:

> Any output that is not a version, a plan, a test line, a YAML block, a diagnostic or a bail out is incorrect.

So, for custom diagnostics lines I think it is useful to have these instruction at least until sth like #64 is implemented. It is valid behavior mentioned in the spec that Bats is capable of exhibiting, but it is jot documented anywhere explicitly.

For non diagnostic, arbitrary lines, I agree that it is a bad idea to use them. They can be useful in debugging, but the same debugging can be with the "more" proper diagnostic lines. I think I read somewhere that Bats is version 12 compliant, in which case it is not wrong behavior, though.

But, apart from the above, I think this passage is also useful for understanding how Bats works internally. Maybe it is too much for a README and it makes more sense in a wiki page (but this also brings the issue of maintaining two complementary but separate documentations).

mbland:
Just moving the ball forward here. If we want more tweaks/updates, let's open a new PR. This write-up does a pretty good job of explaining the current state.

PR #66, opened 2018-04-01
Adds next version changelog

tterranigma:
Two things here:

- judging by the "Added ..." lines of the changelog, this justifies pumping to 0.5.0 instead of a patch release, ie 0.4.1
- I changed the order of mentioning changes, I mention first the new stuff ("Added..."), then the improved stuff and then the bug fixes.


sublimino:
Awesome @tterranigma, thanks! I think `Added support for tabs in test names.`

Added this to the 0.5.0 milestone as the final step before release.

README.md
```diff
@@ -290,7 +290,17 @@ To learn how to set up your editor for Bats syntax highlighting, see
...
+## Version 
+
+*0.5.0* (April xx, 2018)
+
+* Added support for Docker.
+* Added support for test scripts that have the [unofficial strict mode](redsymbol.net/articles/unofficial-bash-strict-mode/) enabled.
+* Added support for tabs in test names.
+* Improved stability on Windows and macOS platforms.
+* Fixed an installation bug where if Bats was uploaded from Windows to a server installation would fail ([commit](7afdf57bd3cbd216f6dd98b2112f2cf4fa521870))
+* Fixed TAP compliance for skipped tests
```
mbland: Some more items to add:

* Massive performance improvements, especially on Windows (#8)
* Workarounds for inconsistent behavior between Bash versions (#82)
* Workaround for preserving stack info after calling an exported function under Bash < 4.4 (#87)

It might also be worth noting at the top that 0.5.0 generally preserves compatibility with 0.4.0, but with some compatibility improvements and a massive performance boost. In other words:

* all existing tests should remain compatible
* tests that might've failed or exhibited unexpected behavior on earlier versions of Bash should now also pass or behave as expected

README.md
```diff
@@ -290,7 +290,26 @@ To learn how to set up your editor for Bats syntax highlighting, see
...
+
+* Added support for Docker.
+* Added support for test scripts that have the [unofficial strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) enabled.
+* Improved stability on Windows and macOS platforms.
+* Massive performance improvements, especially on Windows (#8)
+* Workarounds for inconsistent behavior between Bash versions (#82)
+* Workaround for preserving stack info after calling an exported function under Bash < 4.4 (#87)
+* Fixed TAP compliance for skipped tests
+* Added support for tabs in test names.
+* Fixed an installation bug where if Bats was uploaded from Windows to a server installation would fail ([commit](https://github.com/bats-core/bats-core/commit/7afdf57bd3cbd216f6dd98b2112f2cf4fa521870))
```
mbland: Ah...actually, I think that bug's finally fixed by #88, wherein I changed `bin/bats` from a symlink to a wrapper script, given the continued problems from #32.

tterranigma:
Last changes are in place. This looks like a solid changelog. How about tagging 0.5.0? That would be really awesome.. Moving almost!

q.bats
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bats
```
mbland: This file looks extraneous. Remove it?

README.md
```diff
@@ -384,6 +384,24 @@ on the wiki.
 
 Bats is [SemVer compliant](https://semver.org/).
 
+*0.5.0* (June xx, 2018)
```
mbland: This should be `1.0.0` now. Let's go ahead and put `June 8, 2018` in there, too. (If we slip, I can tweak it when it's time to tag and release per #16.)

README.md
```diff
@@ -384,6 +384,24 @@ on the wiki.
...
+
+`0.5.0` generally preserves compatibility with `0.4.0`, but with some Bash compatibility improvements and a massive performance boost. In other words:
```
mbland: `1.0.0`

README.md
```diff
@@ -384,6 +384,24 @@ on the wiki.
...
+
+* Added support for Docker.
+* Added support for test scripts that have the [unofficial strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) enabled.
+* Improved stability on Windows and macOS platforms.
+* Massive performance improvements, especially on Windows (#8)
+* Workarounds for inconsistent behavior between Bash versions (#82)
+* Workaround for preserving stack info after calling an exported function under Bash < 4.4 (#87)
+* Fixed TAP compliance for skipped tests
+* Added support for tabs in test names.
+* Fixed an installation bug where if Bats was uploaded from Windows to a server installation would fail (#88)
```
mbland: Actually, that would be #91 now. Also it seems it could be worded better, something like: `bin/bats` and `install.sh` now work reliably on Windows (#91).

mbland: I meant that the new line should replace this one, but no matter; I'll delete the old line during the cleanup pass.

PR #74, opened 2018-05-01
Want method to compare output against expected checksum of given output

szaydel:
I am not sure if this is something others will find useful. I wanted to have a quick method to pass a checksum and results captured in `$output` and get a pass|fail. That's what this changeset accomplishes. If there's interest, but tests must be provided before it could be integrated, I am happy to create them.

Thanks!

### DIFF
--- a/libexec/bats-core/bats-exec-test
+++ b/libexec/bats-core/bats-exec-test
@@ -71,6 +71,21 @@ skip() {
   exit 0
 }
 
+md5match() {
+  string="$1"
+  expsum="$2"
+  if which md5sum >/dev/null 2>&1; then
+    cmd=`which md5sum`
+    sum=`"$cmd" <<< "$string" | cut -d' ' -f1`
+  else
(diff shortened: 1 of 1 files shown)


tterranigma:
This would be suitable for an assertion lib. In #44 there is discussion on whether to incorporate extension projects into bats-core. But I think this kind of assertion you have here is too specific to be included in any case. If basic assertions are included in bats-core then the same could be achieved with sth like:

`assert_equal "$output" "$(md5sum <<< "$string")"`

szaydel:
This makes sense. In my case I did not want to generalize, and explicitly wanted to have a checksum comparison to allow for checksums to be stored as opposed to arbitrary strings, because I put together a sort of a library of tests which also have a config-like file that relevant tests source and it was cleaner in my case to have fixed length items in the config, like checksums as opposed to in some instances multiline strings with in some cases tabs and several newlines, etc.

martin-schulze-vireso:
I am closing this, since it has already been found unsuitable for bats-core.

Issue #81, opened 2018-05-11
silent failure when accessing unbound variable with set -u


eedwards-sk:
bash version: GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin17)

similar to #72 where toggling `set +/-e` during the command seems to be a workaround

test file

```
#!/usr/bin/env bats

setup() {
  set -u
}

@test "example" {
  echo $unbound_variable
}
```

output

```
→ ./test.bats
 ✗ example

1 test, 1 failure
```

with workaround

```
#!/usr/bin/env bats

setup() {
  set -u
}

@test "example" {
  set +e
  echo $unbound_variable
  set -e
}
```

workaround output

```
→ ./test.bats
 ✗ example
   (in test file test.bats, line 9)
     `echo $unbound_variable' failed
   /var/folders/dm/2_68m_h13h5dxq2f05wsc47c0000gq/T/bats.76966.src: line 9: unbound_variable: unbound variable

1 test, 1 failure
```

tterranigma:
I cannot reproduce this with `GNU bash, version 3.2.51(1)-release (x86_64-unknown-linux-gnu)`. I get the stack trace as expected.

mbland:
Rehashing my methodology from #72, here's the content of `unset.bash`:

```bash
#! /usr/bin/env bash

trap 'echo ERR:  status == "$?", funcname == "${FUNCNAME:-main}"' ERR
trap 'echo EXIT: status == "$?", funcname == "${FUNCNAME:-main}"' EXIT

do_it() {
  echo "$unset_var"
}
do_it
```

And as before, `/bin/bash` is 3.2.57(1)-release (stock Bash on macOS), and `/usr/local/bin/bash` is 4.4.19(1)-release:

```
$ /bin/bash -u unset.bash 
unset.bash: line 7: unset_var: unbound variable
EXIT: status == 1, funcname == do_it

$ /bin/bash -eu unset.bash 
unset.bash: line 7: unset_var: unbound variable
EXIT: status == 0, funcname == do_it

$ /usr/local/bin/bash -u unset.bash 
unset.bash: line 7: unset_var: unbound variable
EXIT: status == 1, funcname == main

$ /usr/local/bin/bash -eu unset.bash 
unset.bash: line 7: unset_var: unbound variable
EXIT: status == 1, funcname == do_it
```

Interesting things to note:

[...]

Issue #80, opened 2018-05-10
Work around `source` bug when run under `set -e` in Bash 3.2.57(1)-release


mbland:
Per my analysis in #72, we should update `load()` to toggle `set +e/-e` around the source command in its implementation, along with an explicit error check. Then we should strongly recommend its use over the builtin `source` command.

eedwards-sk:
Thanks for opening this! Please also consider #79 if we're going to recommend usage of `load()`

martin-schulze-vireso:
The current code in load seems to work around this. However, the documentation should clearly recommend using `load` over `source`.

PR #82, opened 2018-05-14
Show errors when Bash EXIT trap fires without ERR

### LINKED ISSUE #72: Failure in setup function causes test to fail without displaying errors
This may be my lack of understanding of bash, so forgive me if it is but I've come across the following situation which doesn't 'feel' right.

I've created a setup function which currently fails. All the setup function is configured to do is source a file which will contain constants but, at the moment, doesn't yet exist. Obviously this is a problem I will fix, but what concerned me was it resulted in the subsequent test failing but without displaying an error.

This can be replicated with the following test script:

```
function setup {
    source "/non-existant-file"
}

@test "some random test" {
    # This should always pass
    return 0
}
```

This will result in:

```
 ✗ some random test
```

which doesn't feel right. If I force the setup function to explicitly return 1:

```
function setup {
    source "/non-existant-file" || return 1
}
```

The result makes more sense:
```
 ✗ some random test
   (from function `setup' in test file test.bats, line 2)
     `source "/non-existant-file" || return 1' failed
   /tmp/bats.14184.src: line 2: /non-existant-file: No such file or directory
```
Is this my lack of understanding of bash or is something else at work here?

mbland:
Closes #72 and #81. Both of those issues deal with the `bats_error_trap` executing with an incorrect value for `$?`. Bats exited with an error, but provided no stack trace information to pinpoint its source.

* #72 deals with the inconsistent failure-mode behavior of the `source` builtin between Bash versions. I believe it's feasible to update `load` with robust error checking around `source` and strongly recommend its use in Bats test files instead of using `source` directly. I've already opened #80 to track this.

* #81 deals with inconsistent failure-mode behavior of `set -u`, particularly when code within a Bats test file itself references an unbound variable.

With credit to sstephenson, Bats was smart enough to know and to report that an error occurred, even when it couldn't tell exactly where the error came from.

Building on that existing mechanism, this change produces output for these failure cases even when `bats_error_trap` isn't called with the correct value for `$?`. It passes all the existing tests under Bash 3.2.57(1)-release and 4.4.19(1)-release. I also timed the original suites with and without the change to ensure the runtime cost was negligible.

A couple more notes:

* This change obsoletes the suggestion in #81 that a test case to validate testing code running under `set -u` is necessary. `run()` disables `set -e` for the code under test to begin with, sidestepping the problem with `set -eu` interactions in Bash versions prior to 4.1.0.

* I've noticed that this mechanism does _not_ work when the problematic behavior is in `teardown()`. The change preserves existing behavior, the remedy for that issue requires further thinking (and a new issue).

cc: @cpmills1975 @tterranigma @eedwards-sk

### DIFF
--- a/libexec/bats-exec-test
+++ b/libexec/bats-exec-test
@@ -244,10 +244,10 @@ bats_debug_trap() {
 bats_error_trap() {
   local status="$?"
   if [[ -z "$BATS_TEST_COMPLETED" ]]; then
-    if [[ "$status" -eq 0 ]]; then
-      status=1
+    BATS_ERROR_STATUS="$status"
+    if [[ "$BATS_ERROR_STATUS" -eq 0 ]]; then
+      BATS_ERROR_STATUS=1
     fi
-    BATS_ERROR_STATUS="${status:-1}"
     BATS_ERROR_STACK_TRACE=( "${BATS_PREVIOUS_STACK_TRACE[@]}" )
     trap - debug
   fi
--- a/libexec/bats-exec-test
+++ b/libexec/bats-exec-test
@@ -95,6 +95,7 @@ bats_test_function() {
 
 BATS_CURRENT_STACK_TRACE=()
 BATS_PREVIOUS_STACK_TRACE=()
+BATS_ERROR_STACK_TRACE=()
 
 bats_capture_stack_trace() {
   if [[ "${#BATS_CURRENT_STACK_TRACE[@]}" -ne '0' ]]; then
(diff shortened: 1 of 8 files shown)


mbland:
@jasonkarns @sublimino and @btamayo Life pulled me away from this and other projects for a while, but it seems I've regained the bandwidth to start incrementally moving things forward again.

If at least one of y'all could review this, I'd appreciate it. I think it can safely sneak in before our "release"—which I'd also like to finally get across the finish line soon.

mbland:
Gang, I'm gonna go ahead and merge this. If there's any issues after-the-fact, I'm happy to file a new issue and address them later.

PR #88, opened 2018-06-01
Cleanup

mbland:
Various cleanup commits, best reviewed one-at-a-time. Can split some of these out into separate PRs if desired. All together shaves about 0.5s off the suite's running time on my MacBook Pro with a 2.9 GHz Intel Core i5.

- [X] I have reviewed the [Contributor Guidelines][contributor].
- [X] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: CONTRIBUTING.md
[coc]:         CODE_OF_CONDUCT.md

PR #91, opened 2018-06-07
bin/bats, install.sh: Fix symlink bugs

### LINKED ISSUE #53: install.sh needs to not change permissions on files it didn't install

### LINKED ISSUE #83: Busybox issue
I am running Alpine Linux and when I clone the master branch of bats-core and run `install.sh` I get an error. Here is an example of the output I am seeing.

```
$ cat /etc/alpine-release 
3.7.0

$ ls
CODE_OF_CONDUCT.md  Dockerfile          README.md           install.sh          man                 test
CONTRIBUTING.md     LICENSE.md          bin                 libexec             package.json

$ sudo ./install.sh /usr/local/
readlink: unrecognized option: e
BusyBox v1.27.2 (2018-01-29 15:47:03 UTC) multi-call binary.

Usage: readlink [-fnv] FILE

Display the value of a symlink

        -f      Canonicalize by following all symlinks
        -n      Don't add newline
        -v      Verbose

$ ls -l $(which readlink)
lrwxrwxrwx    1 root     root            12 Feb 17 18:00 /usr/bin/readlink -> /bin/busybox

```

### LINKED ISSUE #90: New file bin/bats breaks existing basher installations
The new ``bin/bats`` is a wrapper to call the bats frontend:
```
  exec "${0%/*}/../libexec/bats" "$@"
```
The ``$0`` in a basher installation and in other installations may be a symlink.
Taking the directory part of the symlink gives the wrong path, so an additional step is needed:
```
  _0=$(realpath "$0")
```
which is why perhaps a symlink is better. I see the objective (#32) is to avoid issues with symlinks on windows.

I'd really like to use a packer manager to get basher. basherpm works well for small projects, as long as some common sense rules are followed. Maybe this isn't supported, or maybe there is something better. What is the preferred installation method?

mbland:
Closes #90. Closes #83. Closes #56. Closes #55. Closes #53. Resolves outstanding issues from #32.

Per all these earlier issues and PRs:

- The original bin/bats symlink was causing problems on Windows.
- The attempted fix from #32 created problems for install.sh.
- Per #90, changing bin/bats to a one-line script in #88 broke some Bats installations that rely on symlink schemes (such as when installed via https://github.com/basherpm/basher).

For an idea of why I wanted to keep bin/bats as a script due to how symlinks complicate things on Windows, see the following results I discovered during a search for "git symlink windows":

- https://github.com/git-for-windows/git/wiki/Symbolic-Links
- https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/
- https://github.com/libgit2/libgit2/issues/4107
- https://stackoverflow.com/q/5917249

In the course of applying these changes, I realized libexec/bats computed some extraneous variables, so I removed them, eliminating a few external processes and subshells. I also cleaned up other small bits of logic.

On top of making install.sh, bin/bats, and libexec/bats more resilient and compact, the existing test suite (before adding the new test/installer.bats file) sped up significantly, running 0.6s-0.7s faster on my MacBook Pro (2.9 GHz Intel Core i5, 8GB RAM):

```
  Bash 3.2.57(1)-release before:
  58 tests, 0 failures

  real    0m4.924s
  user    0m3.045s
  sys     0m1.798s

  Bash 3.2.57(1)-release after:
  58 tests, 0 failures

  real    0m4.341s
  user    0m2.808s
  sys     0m1.540s

  Bash 4.4.23(1)-release before:
  58 tests, 0 failures

  real    0m5.228s
  user    0m3.046s
  sys     0m1.952s

  Bash 4.4.23(1)-release after:
  58 tests, 0 failures

  real    0m4.582s
  user    0m2.791s
  sys     0m1.643s
```

Also tweaks the Dockerfile to update the symlink to point to bin/bats, not libexec/bats.

cc: @bvberkum @timdaman @tpraxl @tterranigma @NHellFire @jfelchner @xnum @ntnn 

### DIFF
--- a/bin/bats
+++ b/bin/bats
@@ -1,3 +1,35 @@
 #! /usr/bin/env bash
 
-exec "${0%/*}/../libexec/bats" "$@"
+set -e
+
+export BATS_READLINK='true'
+if command -v 'greadlink' >/dev/null; then
+  BATS_READLINK='greadlink'
+elif command -v 'readlink' >/dev/null; then
+  BATS_READLINK='readlink'
+fi
+
+bats_resolve_link() {
+  if ! "$BATS_READLINK" "$1"; then
+    return 0
+  fi
+}
+
+bats_absolute_path() {
+  local cwd="$PWD"
+  local path="$1"
+
+  while [[ -n "$path" ]]; do
+    cd "${path%/*}"
+    local name="${path##*/}"
+    path="$(bats_resolve_link "$name")"
+  done
+
+  printf -v "$2" -- '%s' "$PWD"
+  cd "$cwd"
+}
+
+export BATS_ROOT
+bats_absolute_path "$0" 'BATS_ROOT'
+BATS_ROOT="${BATS_ROOT%/*}"
+exec "$BATS_ROOT/libexec/bats" "$@"
--- a/install.sh
+++ b/install.sh
@@ -1,23 +1,6 @@
 #!/usr/bin/env bash
+
 set -e
-
-resolve_link() {
-  $(type -p greadlink readlink | head -n1) "$1"
-}
-
-abs_dirname() {
-  local cwd="$(pwd)"
-  local path="$1"
-
-  while [ -n "$path" ]; do
-    cd "${path%/*}"
-    local name="${path##*/}"
-    path="$(resolve_link "$name" || true)"
-  done
-
-  pwd
-  cd "$cwd"
-}
 
 PREFIX="$1"
 if [ -z "$1" ]; then
@@ -27,15 +10,48 @@ if [ -z "$1" ]; then
   exit 1
 fi
 
-BATS_ROOT="$(abs_dirname "$0")"
-mkdir -p "$PREFIX"/{bin,libexec,share/man/man{1,7}}
-cp -R "$BATS_ROOT"/bin/* "$PREFIX"/bin
-cp -R "$BATS_ROOT"/libexec/* "$PREFIX"/libexec
-cp "$BATS_ROOT"/man/bats.1 "$PREFIX"/share/man/man1
-cp "$BATS_ROOT"/man/bats.7 "$PREFIX"/share/man/man7
+export BATS_READLINK='true'
+if command -v 'greadlink' >/dev/null; then
+  BATS_READLINK='greadlink'
+elif command -v 'readlink' >/dev/null; then
+  BATS_READLINK='readlink'
+fi
 
-# fix file permission
-chmod a+x "$PREFIX"/bin/*
-chmod a+x "$PREFIX"/libexec/*
+bats_resolve_link() {
+  if ! "$BATS_READLINK" "$1"; then
+    return 0
+  fi
+}
+
+bats_absolute_path() {
+  local cwd="$PWD"
+  local path="$1"
+
+  while [[ -n "$path" ]]; do
+    cd "${path%/*}"
+    local name="${path##*/}"
+    path="$(bats_resolve_link "$name")"
+  done
+
+  printf -v "$2" -- '%s' "$PWD"
+  cd "$cwd"
+}
+
+bats_install_scripts() {
+  local scripts_dir
+  local scripts=()
+
+  for scripts_dir in "$@"; do
+    scripts=("$BATS_ROOT/$scripts_dir"/*)
+    cp "${scripts[@]}" "$PREFIX/$scripts_dir"
+    chmod a+x "${scripts[@]/#$BATS_ROOT[/]/$PREFIX/}"
+  done
+}
+
+mkdir -p "$PREFIX"/{bin,libexec,share/man/man{1,7}}
+bats_absolute_path "$0" 'BATS_ROOT'
+bats_install_scripts 'bin' 'libexec'
+cp "$BATS_ROOT/man/bats.1" "$PREFIX/share/man/man1"
+cp "$BATS_ROOT/man/bats.7" "$PREFIX/share/man/man7"
 
 echo "Installed Bats to $PREFIX/bin/bats"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM alpine:3.6
 
 RUN apk --no-cache add bash \
-    && ln -s /opt/bats/libexec/bats /usr/sbin/bats
+    && ln -s /opt/bats/bin/bats /usr/sbin/bats
 
 COPY . /opt/bats/
 
--- a/libexec/bats
+++ b/libexec/bats
@@ -26,57 +26,23 @@ help() {
   echo
 }
 
-export BATS_READLINK='true'
-if command -v 'greadlink' >/dev/null; then
-  BATS_READLINK='greadlink'
-elif command -v 'readlink' >/dev/null; then
-  BATS_READLINK='readlink'
-fi
-
-resolve_link() {
-  if ! "$BATS_READLINK" "$1"; then
-    return 0
-  fi
-}
-
-abs_dirname() {
-  local cwd="$PWD"
-  local path="$1"
-
-  while [[ -n "$path" ]]; do
-    cd "${path%/*}"
-    local name="${path##*/}"
-    path="$(resolve_link "$name")"
-  done
-
-  printf -v "$2" -- '%s' "$PWD"
-  cd "$cwd"
-}
-
 expand_path() {
   local path="${1%/}"
   local dirname="${path%/*}"
 
   if [[ "$dirname" == "$path" ]]; then
     dirname="$PWD"
-  elif cd "$dirname" 2>/dev/null; then
+  else
+    cd "$dirname"
     dirname="$PWD"
     cd "$OLDPWD"
-  else
-    printf '%s' "$path"
-    return
   fi
   printf -v "$2" '%s/%s' "$dirname" "${path##*/}"
 }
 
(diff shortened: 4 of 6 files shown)


bin/bats
```diff
@@ -1,3 +1,35 @@
 #! /usr/bin/env bash
```
jasonkarns: Do we have a preference for `#!/usr` vs `#! /usr`? I don't really care, but just noticed that not all the scripts are consistent.

install.sh
```diff
@@ -1,41 +1,57 @@
...
 
-PREFIX="$1"
-if [ -z "$1" ]; then
-  { echo "usage: $0 <prefix>"
-    echo "  e.g. $0 /usr/local"
-  } >&2
-  exit 1
-fi
+bats_install_scripts() {
+  local scripts_dir
```
jasonkarns: Style question, any reason not to one-line the locals?

install.sh
```diff
@@ -1,41 +1,57 @@
...
 mkdir -p "$PREFIX"/{bin,libexec,share/man/man{1,7}}
-cp -R "$BATS_ROOT"/bin/* "$PREFIX"/bin
-cp -R "$BATS_ROOT"/libexec/* "$PREFIX"/libexec
-cp "$BATS_ROOT"/man/bats.1 "$PREFIX"/share/man/man1
-cp "$BATS_ROOT"/man/bats.7 "$PREFIX"/share/man/man7
-
-# fix file permission
-chmod a+x "$PREFIX"/bin/*
-chmod a+x "$PREFIX"/libexec/*
+bats_absolute_path "$0" 'BATS_ROOT'
```
jasonkarns: I'm not following how `bats_absolute_path` works to set BATS_ROOT. (And my macOS/bsd `printf` doesn't have `-v` documented... I think these things are related 😄 )

bin/bats
mbland: I've tended to prefer `#! /usr`, but don't feel strongly. I think we might need to do one more tiny cleanup pass before releasing to tighten up little details like this.

install.sh
mbland: In my old age, I've begun to favor "letting the code breathe" by splitting things across multiple lines. Which is to say, I used to optimize for vertical space in several languages, only to curse myself later for being so clever that I had to spend a disproportionate amount of time and mental energy unpacking my compact expressions when I had to revisit them later.

Granted, putting a couple of locals on the same line isn't a big deal, but I just split them to be consistent.

install.sh
mbland: Ah, that's because `printf` is a Bash builtin, and the `-v` option takes the name of a variable into which to store the result. This is much quicker than capturing stdout into a variable via `$()`.

Was that the only question you had about `bats_absolute_path`? Happy to unpack it some more.

jasonkarns: Thanks! That was all I didn't quite grok. And got lazy trying to find docs on `-v` (even a cursory google search didn't include -v in their docs, so thanks for the lazyweb explain 😄 )

install.sh
jasonkarns: Yep. My admittedly narrow opinion is to distinguish "noise" from signal in the code. Variable declarations qualify as noise to me since they aren't really interesting, so I try to compress the amount of space that noise takes up, which in turn let's me more easily uncompress the meaningful parts of the code and let the _signal_ breathe 😄

jasonkarns:
Awesome to see all those issues closed!

install.sh
```diff
@@ -1,41 +1,57 @@
...
+
+bats_resolve_link() {
+  if ! "$BATS_READLINK" "$1"; then
+    return 0
+  fi
 }
 
-abs_dirname() {
-  local cwd="$(pwd)"
+bats_absolute_path() {
```
sublimino: As above, also wonder if this should be refactored into a shared file as it's reused

sublimino:
Sorry, late to the party!

bin/bats
```diff
@@ -1,3 +1,35 @@
...
+  local cwd="$PWD"
+  local path="$1"
+
+  while [[ -n "$path" ]]; do
+    cd "${path%/*}"
+    local name="${path##*/}"
+    path="$(bats_resolve_link "$name")"
+  done
+
+  printf -v "$2" -- '%s' "$PWD"
```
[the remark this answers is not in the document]
mbland: There is a reason: `printf -v` will write to a variable, and in this case, to a variable in the caller's scope, the name of which the caller specifies as `$2`. To avoid name collisions with variables outside of the scope of functions like this, I try to use positional parameters directly (whereas I normally do assign them to local variables for clarity's sake).

But I'm just realizing I've got two other local variables here, and I'm not following my usual practice of mangling their names to reduce the likelihood of collision (where "mangling" here is prefixing variable names with something like `__bats_abs_path_` to reduce the likelihood of collision). I could also add such a local variable for `$2` as well.

Any preference?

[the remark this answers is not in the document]
sublimino: Thanks for the detail! Personal preference is to be explicit (I too forgot about `printf`'s variable assignment), so a "mangled reference" (sounds like that should be a programming term!) would be my suggestion.

install.sh
mbland: I was thinking this through, and still am, but there's a chicken-and-egg problem—or there might not be, and I'd be interested in your thoughts.

Apologies in advance for the length.

The way the system works depends upon `bin/bats` discovering the real Bats root directory. For example, if Bats is installed at `/opt/bats`, then `/opt/bats/bin/bats` needs to find `/opt/bats/libexec/bats`. `/opt/bats/libexec/bats` then adds `/opt/bats/libexec` to the front of `PATH`, so then Bats can find the rest of the `/opts/bats/libexec` scripts. Because of this, `/usr/local/bin/bats` can symlink to `/opt/bats/bin/bats`, and the rest of Bats's internals can remain encapsulated within `/opt/bats` such that they don't pollute the `/usr/local` namespace (the man pages notwithstanding, unless `/opt/bats/man` is added to `MANPATH` somehow).

Previously, having `bin/bats` be a symlink to `../libexec/bats` worked fine in general—but broke down in the `/usr/local/bin/bats` with `/opt/bats` scenario I described above, and was creating complications for some Windows-based platforms. Hence why I made `bin/bats` a normal script, then pushed the real path logic in there.

Now `install.sh` from sstephenson/bats also originally contained this real path-finding logic. Having duplicate code rubs me terribly wrongly, but if `bin/bats` needs to find its real path, and `install.sh` needs to find its real path...then how do they each find their real path so that they can import a shared script containing the real path logic?

One possibility I thought of, but I'm not sure it's worth the complication: If there were more than one Bats executable, they could all source `${0%/*}/bats`, which could use `${BATS_SOURCE[1]}` to discover which script sourced it, and could then dispatch logic between them. We could set up `install.sh` to source `${0%/*}/bin/bats` to achieve this—but then we presume `install.sh` isn't symlinked directly, which would land us back in the same predicament.

But what if we assumed (or specified that) `install.sh` itself won't (shouldn't) ever be symlinked? In that case we probably _could_ just compute the Bats root directory as the parent of `install.sh` (`${0%/*}`) without finding the real path. In other words, we can't assume `/usr/local` is the correct Bats root for `/usr/local/bin/bats`, but we may be safe in assuming (or specifying) that `/opt/bats` is the correct Bats root for `/opt/bats/install.sh`. This would eliminate the need for the duplicated code in the latter script.

Thoughts?

bin/bats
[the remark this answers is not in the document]
mbland: OK; I'll add those in #92. (Except for the one-liners, though.)

BTW, have you never done C++? "Name mangling" is a C++ implementation detail, but one that C++ coders need to be familiar with.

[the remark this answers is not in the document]
sublimino: Only casual commits, never anything serious. Glad to have learned it though 🤣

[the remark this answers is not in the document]
mbland: So I pushed a new commit to #91, but only went so far, because the functions in `libexec/bats-exec-test` were mostly short, but all called frequently enough that there was a slight-yet-noticeable performance impact. I also decided not to mangle literally everything since these functions are all internal, so we control how they're called, and they're easier to read without all the mangling.

install.sh
sublimino: Hehe, I've been mulling it over and hacking around. Will pull the branch and check it out, cheers.

sublimino: Sorry to cause more problems (!) - in writing the usage guide in #94 I found an edge case, which I've put into the travis build as per #93. Essentially the changes in #92 break running the tests when the tests are mounted into the container at a path that isn't a direct decedent of the bats base directory.

Current master doesn't fail under these conditions. I don't suspect anybody except us is using the pattern described above, but I can't be sure of the repercussions as I'm not as close to this code as you are,

mbland: Since it's the unofficial Bats strict mode test that's failing, I think I need to add a couple `declare` statements back into `install.sh` and that should do the trick. Will try in a couple hours.

mbland: Wait, that might not be it...either way, sure it's something small, and I'm doing a terrible job multitasking right now. :-P

mbland: To close the loop on this: I found and explained the breakage in #93, which actually wasn't due to #92. The "unofficial Bash strict mode" test failed because it was looking for `bats` relative to `BATS_TEST_DIRNAME`, which it couldn't because the `test` directory was mounted into the container in isolation. Mounting the root of the Bats project resolved the issue.

PR #92, opened 2018-06-08
Pre v1.0.0 cleanup

mbland:
Contains several commits with very small touch-ups prior to launching v1.0.0 soon per #16. Also includes a pretty major refactor of the README.md file that preserves most of the existing content, but makes the formatting more consistent and the content flow a bit better (I think).

@jasonkarns Let me know how you'd like to proceed with tagging/releasing after this gets merged. Happy to do it myself. I'm also thinking of including the v1.0.0 version history from README.md in a signed v1.0.0 tag (which I think will automatically show up in https://github.com/bats-core/bats-core/releases).

- [X] I have reviewed the [Contributor Guidelines][contributor].
- [X] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: CONTRIBUTING.md
[coc]:         CODE_OF_CONDUCT.md

PR #93, opened 2018-06-08
build: add travis test for mounted test files

sublimino:
This is to cover the use case of developing against bats in a container
with a watch command.

It currently fails due to issues resolving bats' own path as per the issues under discussion in https://github.com/bats-core/bats-core/pull/91

---


mbland:
Odd...the failure is due to a violation of unofficial Bash strict mode. Will look a little later.

mbland:
Ah, now I see it, plain as day (from https://travis-ci.org/bats-core/bats-core/jobs/389756078):

```
not ok 42 ensure compatibility with unofficial Bash strict mode
[...snip...]
# output:
# bash: //libexec/bats: No such file or directory
```

This because the test case runs Bats thus:

```bash
  run bash -u "${BATS_TEST_DIRNAME%/*}/libexec/bats" \
    "$FIXTURE_ROOT/unofficial_bash_strict_mode.bats"
```

So `${BATS_TEST_DIRNAME%/*}/libexec/bats` resolves to `//libexec/bats`, when it should resolve to `/opt/bats/libexec/bats`. (Also, it should be `${BATS_TEST_DIRNAME}/bin/bats`. Pushed a commit to #92 for that.)

So I'd suggest the command should be some variation of the following:

```bash
docker run -it -v $(pwd):/opt/bats-dev bats:latest /opt/bats-dev/test
```

Yep, that runs very nicely!

mbland:
I should add, to be clear: the failure was due to the `/test` directory being isolated from the rest of Bats. It's reasonable for those file to expect to access files in the same project, so I'd recommend mounting the root of the project in the container, not just the test directory alone.

mbland:
And actually, I'm beginning to wonder if the extra build step is of any value to our build, given the documented technique in #94. What do you think?

mbland:
Just added a comment to your latest update in #93. I think I understand what you're saying now, and suggested more explicit wording.

sublimino:
I re-tested it earlier today and your assertion is correct @mbland. Will get it fixed

martin-schulze-vireso:
I am closing this since we don't use travis anymore. The usecase might still be valid. If so, we should put up an issue for that.

PR #94, opened 2018-06-08
docs: add usage guide

### LINKED ISSUE #30: Write up Docker usage guide
Placeholder - I intend to work on this in the coming weeks. 

- Basic usage
- Docker-y gotchas
- Extending from the base image
- `slim` vs `full` images
- Examples of running in various CIs

Any further topics?

sublimino:
It should be noted that the tests fail when run in the manner suggested in this help guide.

```bash
docker run -it \
  -v $(pwd)/test:/test \
  bats:latest \
  /test
```

I believe that this should be a valid use case for testing, and that
this will be solved by resolving the bats directory and support files as
per the issues under discussion in https://github.com/bats-core/bats-core/pull/91

Once this is merged I'll paste it into the GitHub wiki, or remove it from this tree - would just like it to be reviewed first.

closes #30 
---


wiki/usage.md
```diff
@@ -0,0 +1,51 @@
...
+Receiving objects: 100% (1222/1222), 327.28 KiB | 1.70 MiB/s, done.
+Resolving deltas: 100% (661/661), done.
+
+$ cd bats-core/
+$ docker build --tag bats:latest .
+...
+$ docker run -it bats:latest --tap /opt/bats/test
+```
+
+To mount your tests into the container, first built the image as above. Then:
```
mbland: "built" => "build"

wiki/usage.md
```diff
@@ -0,0 +1,51 @@
...
+```bash
+$ docker run -v $(pwd)/test:/test bats /test
+```
+This has run the `test/` directory from the bats-core repository inside the bats Docker container.
```
mbland: "has run" => "runs"

wiki/usage.md
```diff
@@ -0,0 +1,51 @@
...
+
+$ cd bats-core/
+$ docker build --tag bats:latest .
+...
+$ docker run -it bats:latest --tap /opt/bats/test
+```
+
+To mount your tests into the container, first built the image as above. Then:
+```bash
+$ docker run -v $(pwd)/test:/test bats /test
```
mbland: As noted in #93, this should be something like:

```bash
docker run -it -v $(pwd):/opt/bats-dev bats:latest /opt/bats-dev/test
```

mbland: I take this back slightly; the `/opt/bats-dev` dir should be `/opt/bats`, i.e.:

```
docker run -it -v $(pwd):/opt/bats bats:latest /opt/bats/test
```

And I'm suddenly finding this command extremely useful. 🙂

wiki/usage.md
```diff
@@ -23,12 +23,18 @@ $ docker build --tag bats:latest .
 $ docker run -it bats:latest --tap /opt/bats/test
 ```
 
-To mount your tests into the container, first built the image as above. Then:
+To mount your tests into the container, first built the image as above. Then, for example with `bats`:
 ```bash
-$ docker run -v $(pwd)/test:/test bats /test
+$ docker run -it -v $(pwd):/opt/bats bats:latest /opt/bats/test
 ```
 This has run the `test/` directory from the bats-core repository inside the bats Docker container.
```
mbland: This "has run" still reads funny.

wiki/usage.md
```diff
@@ -23,12 +23,18 @@ $ docker build --tag bats:latest .
...
 
+
+Or for test files without any dependencies:
```
mbland: I think I understand now what you're saying... It would help to explain some more, something like: "For test suites that are intended to run in isolation from the project (i.e. no tests depend on project files outside of the test directory), you can mount the test directory in isolation and execute them like so:"

sublimino:
Thanks for the reviews on this @mbland

wiki/usage.md
```diff
@@ -23,14 +23,14 @@ $ docker build --tag bats:latest .
...
 
-To mount your tests into the container, first built the image as above. Then, for example with `bats`:
+To mount your tests into the container, first build the image as above. Then, for example with `bats`:
 ```bash
 $ docker run -it -v $(pwd):/opt/bats bats:latest /opt/bats/test
 ```
-This has run the `test/` directory from the bats-core repository inside the bats Docker container.
+This runs the `test/` directory from the bats-core repository inside the bats Docker container.
 
+For test suites that are intended to run in isolation from the project (i.e. tests do not depend on project files outside of the test directory), you can mount the test directory by iteself and execute the tests like so:
```
mbland: Nit: "iteself" => "itself"

wiki/usage.md
```diff
@@ -23,14 +23,14 @@ $ docker build --tag bats:latest .
...
 
-Or for test files without any dependencies:
 ```bash
 $ docker run -v $(pwd)/test:/test bats /test
```
mbland: Two nits, one substantial comment.

I'd prefer `-v "$PWD:/opt/bats"` (including double quotes). Though it doesn't matter in terms of performance, it seems folks aren't aware that in Bash, many times you can use `$PWD` where you would commonly use `$(pwd)`, so I like to promote its use.

Should `bats` be `bats:latest`?

Finally, I don't think `-it` is necessary for any of these commands. If you agree, mind not only updating this document, but the `.travis.yml` command?

sublimino: `$PWD` - good point!

`bats` and `bats:latest` are equivalent, but probably more obvious the include the tag. Will update.

`-i` is useful to kill hanging processes (otherwise has to be done via `docker stop` command), and `-t` to simulate a tty (often not used, but most similar to most test runs). Former important to a user, not a build, and latter probably the inverse. Everything's least-surprising to a new Docker use if both are used I think.

wiki/usage.md
sublimino: D'oh

mbland:
So I've approved, but will leave merging up to you, since it seemed you weren't sure you wanted to add this to the repo or just port it to the wiki.

One more thing up to you: Want to add `-it` to that last command, and maybe even the explanation you gave me?

mbland:
Nice! Approval stands. Again I'll leave it up to you whether to merge or not.

PR #96, opened 2018-06-09
Attempt to evaluate and show the contents of a failing assertion

nylen:

This is along the lines of https://github.com/bats-core/bats-core/pull/40, but tries to make the same thing work with built-in bash test commands like `[` and `[[`.

The idea is that we can use `eval` to replace any variable in a test statement with its actual value.  This is far from perfect, and I've accounted for a couple of situations where this shouldn't be done or won't work, but it works and gives really helpful debugging information for most of the tests I am working on so far.

Here's an example.

```sh
@test "example failure" {
  a=x
  [ $a = y ]
}
```

```
$ bats example.bats 
 ✗ example failure
   (in test file svn-wrapper/example.bats, line 5)
     `[ $a = y ]' failed:
     `[ x = y ]'

1 test, 1 failure
```

If there is a possibility of side effects from an `eval` breaking things, it should also be possible to do most of the same work in a subprocess.  Here is an attempt at that:  https://github.com/nylen/bats-core/commits/add/assertion-eval-subprocess


nylen:
It looks like this broke some tests that are looking for specific output, but they are failing now because the output format has changed.

Let me know if this is a change that you're interested in making, and I can clean those failures up.

nylen:
Pinging some folks who have been active here recently for review:  @sublimino @mbland @bvberkum @tterranigma

mbland:
I'm not able to dig deeply into this yet, and it's an interesting idea, but at the moment I'm thinking assertion libraries are a better route. Now that v1.0.2 is out, I'm hoping to give some attention to the problem of either incorporating some basic assertions into Bats, or finding it a way to make it easy to apply a standalone library.

nylen:
I really like the way bats encourages built-in shell tests as assertions. I am going to continue using this technique in my own fork of bats.

nylen:
See https://github.com/bats-core/bats-core/issues/199#issuecomment-1146404757.

Issue #103, opened 2018-06-14
Create bats-core packages to install easilly on linux


kenden:
As the moment it's only possible to install bats from bats-core from source on linux.

We should create packages to install easily on Linux.
It could be:
- `apt install bats-core` for Debian/Ubuntu/Linux Mint
- `yum install bats-core` for Fedora/RedHat/CentOS
- a package on (Flathub](www.flathub.org)
- a snap on https://snapcraft.io/
...

Flathub is probably the easiest to support most distributions in one package.
Starting point: https://github.com/flathub/flathub/wiki/App-Submission

kenden:
Related: the package for mac:
https://github.com/bats-core/bats-core/issues/11

mbland:
Sounds like a good idea to me. I hadn't heard of Flathub before; it seems it's primarily geared towards desktop apps, though. Have you any examples of dev tools handy? (Plus, we don't have an app icon yet.)

kenden:
I don't sorry, it was just an idea, I don't know if is a good fit or not for utilities.

bvberkum:
fwiw. Basher is supported but appearantly not listed on wiki/docs (ie. #90 )

jasonkarns:
Since node is included OOTB on many linux distros, I would also mention that bats is available on npm: https://www.npmjs.com/package/bats

and I suspect the bats-core formula probably works via linuxbrew

virgilwashere:
> btw, may be worth removing https://github.com/bats-core/bats-core/wiki/Install-Bats-Using-a-Package#ubuntu for now from the wiki as one could assume they get `bats-core` if they followed the guide, but don't at the time of writing this.

I had actually editted this a yesterday, before seeing this.  Added the ⚠️ icon and suggestion to use `npm i -g bats`.

As for #193 and the original issue here, I've been researching what's involved in Debian packaging and git recently.  Will try to create a packaging fork and see hard it is to maintain... then continue down the road of getting it accepted upstream upstream.

abegosum:
Noticing that the Fedora package is listed in the wiki- but, none of the RHEL based distributions are in the supported operating systems list.  Is this an oversight, or are RHEL / CentOS / Fedora officially not supported?

martin-schulze-vireso:
Keeping up with men different distributions and package managers seems to be wasteful. We should only provide a standardized means (e.g. install.sh) for distro package managers to package Bats easily.

brokenpip3:
[...]

PR #111, opened 2018-06-28
Included RPM spec contrbution to build OS packages for EL7

pixdrift:
Signed-off-by: pixdrift <support@pixeldrift.net>

### DIFF
--- a/contrib/rpm/bats.spec
+++ b/contrib/rpm/bats.spec
@@ -1,0 +1,52 @@
+%global provider        github.com
+%global project         bats-core
+%global repo            bats-core
+
+Name:           bats
+Version:        1.0.2
+Release:        1%{?dist}
+Summary:        Bash Automated Testing System
+
+Group:          Development/Libraries
+License:        MIT
+URL:            https://%{provider}/%{project}/%{repo}
+Source0:        https://%{provider}/%{project}/%{repo}/archive/v%{version}.tar.gz
+
+BuildArch:      noarch
+
+Requires:       bash
+
+%description
+Bats is a TAP-compliant testing framework for Bash.
+It provides a simple way to verify that the UNIX programs you write behave as expected.
+Bats is most useful when testing software written in Bash, but you can use it to test any UNIX program.
+
+%prep
+%setup -q -n %{repo}-%{version}
+
+%install
+mkdir -p ${RPM_BUILD_ROOT}%{_prefix} ${RPM_BUILD_ROOT}%{_libexecdir} ${RPM_BUILD_ROOT}%{_mandir} 
+./install.sh ${RPM_BUILD_ROOT}%{_prefix}
+
+%clean
+rm -rf $RPM_BUILD_ROOT
+
+%check
+
+%files
+%doc README.md LICENSE.md
+%{_bindir}/%{name}
+%{_libexecdir}/%{repo}
+%{_mandir}/man1/%{name}.1.gz
+%{_mandir}/man7/%{name}.7.gz
+
+%changelog
+* Mon Jun 18 2018 pixdrift <support@pixeldrift.net> - 1.0.2-1
+- Increase version to match upstream release
+- Relocate libraries to bats-core subdirectory
+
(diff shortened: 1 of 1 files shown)


pixdrift:
ack @mbland and thanks for reviving this project, very much appreciated.

A package built from the above .spec is available here (built on CentOS 7) if anyone needs/wants to test it.
http://rpmbuild.pixeldrift.net/bats-core/el7/noarch/Packages/

This package upgrades cleanly from the older `bats` version that currently ships in EPEL.

jasonkarns:
I don't know the rpm spec format, though everything seems in order. 🤷‍♂️  

I'm not terribly excited about having _yet another_ copy of the changelog, though. If/when we extract that to its own file, I'm assuming we can just reference the file instead of duplicating its contents into the rpm spec?

pixdrift:
Understand the concern. As an end user it's nice to have the changelog represented in the spec because it makes it visible through the package manager to the end user.

I guess the changelogs generation could be scripted/extracted from the main changelog if the main changelog maintained a consistent parseable format.

Alternatively, drop the changelog and keep the spec without it, for use as a basic template.

mbland:
Before I forget: this PR is germane to #103.

I imagine changelog dissemination must be a solved problem for other projects that are packaged via multiple mechanisms... Created #112 to track this.

Right after creating this, I discovered (quoted from https://github.com/bats-core/bats-core/issues/112#issuecomment-401047135):

>  Also, note that https://rpm-packaging-guide.github.io/ mentions listing a `ChangeLog` in `%files`, and that the `%changelog` directive is for packaging changes, not software changes.

So @pixdrift, perhaps you can update the `%changelog` to reflect the packaging changes instead? @jasonkarns any concerns/objections once that's done?

jasonkarns:
@mbland nope, with that I'd be 👍

pixdrift:
PR updated. Linked compiled packages above have also been updated (if anyone wants to test).

Can add the tests to the PR, let me know how you would like to handle it. 👍

mbland:
Maybe I'm just tired, but that Fedora `%changelog` advice does seem to contradict the RHEL advice from the guide I dug up. Weird.

I'm OK getting this in for now, and considering the automation to keep version numbers in sync later, as that's probably we can include within the scope of #112 as well.

BTW, what would we need to do to make sure that new RPMs are built from the spec as new versions become available?

pixdrift:
@mbland, not too much. Now that a spec is in the repo, you can trigger a build process for each commit potentially.

I set up similar for my own use for buildah, resulting packages are here:
http://rpmbuild.pixeldrift.net/projectatomic/el7/x86_64/Packages/

This builds (in a container) on every commit to the repo to speed up my own testing/validation. Source is here:
https://gitlab.com/pixdrift/buildah-scripts/blob/master/el7.buildah.rpmbuild.sh

mbland:
Oh, I was thinking this was for including them in the RHEL and/or Fedora distributions. Guess it's a necessary first step, at least.

okorz001:
Dumb question from an RPM newbie desperately trying to get this fork of bats into our internal yum repos: How am I supposed to build an RPM with this? 😅 

```
$ rpmdev-setuptree
$ rpmbuild -v -bb contrib/rpm/bats.spec 
error: File /home/user/rpmbuild/SOURCES/v1.1.0.tar.gz: No such file or directory
```

Do I need to stage something locally before running rpmbuild? I don't understand why rpmbuild is even looking for a file there since there is a github URL in the Source0 field. 😕 

I'm running RHEL 7.7.

pixdrift:
@okorz001 try with the following:
`rpmbuild --undefine=_disable_source_fetch -ba bats.spec`

`rpmbuild` doesn't have a native way to validate the source download, so fetching sources is disabled by default. The above enables source fetching with an override which should only be overridden in situations where you trust the network/source.

There is also a tool called `spectool` which can download missing sources defined in the `.spec`

Sorry I didn't see your message earlier, don't look in old PR's too often 😄

PR #115, opened 2018-07-01
Fix BATS_ROOT resolution when /bin => /usr/bin

### LINKED ISSUE #113: bats can't find BATS_ROOT on CentOS 7 when /bin -> /usr/bin
I suspect that bats_absolute_path() could be reimplemented as simply:
```bash
local path="$1"
local result="$2"
local canonical
canonical=$(readlink -e "$1")
printf -v "${result}" -- '%s' "${canonical%/*}"
```
Currently, when /bin is symlinked to /usr/bin as on CentOS 7 systems then bats can't find bats-core.

mbland:
Closes #113. This should be robust enough to handle the /bin => /usr/bin symlink on CentOS7 as originally reported, as well as any manner of symlink chains leading to the `bin/bats` executable in the installation directory. Maintains portability between `readlink` implementations by not relying on the GNU-specific `-e` option.

There's no perceptible performance impact.

This renames the original `bats_absolute_path` function to the more obvious `bats_resolve_absolute_root_dir`, and introduces the `BATS_MAX_SYMLINKS` variable to ensure symlink resolution short-circuits in case cycles appear. While this can be user-controlled, I've not documented it in the README, hoping that no one will need to change the setting in a real-world situation.

Also tweaks the value of `BATS_TEST_SUITE_TMPDIR` from `tmp` to `bats-test-tmp` to avoid potential collisions with other `BATS_TMPDIR` directories.

cc: @rgm3

### DIFF
--- a/bin/bats
+++ b/bin/bats
@@ -38,8 +38,7 @@ bats_resolve_absolute_root_dir() {
       path="$(bats_resolve_link "$target_name")"
     else
       printf -v "$result" -- '%s' "${PWD%/*}"
-      set +P
-      set "-$original_shell_options"
+      set +P "-$original_shell_options"
       cd "$cwd"
       return
     fi
--- a/bin/bats
+++ b/bin/bats
@@ -9,8 +9,6 @@ elif command -v 'readlink' >/dev/null; then
   BATS_READLINK='readlink'
 fi
 
-BATS_MAX_SYMLINKS="${BATS_MAX_SYMLINKS:-16}"
-
 bats_resolve_link() {
   if ! "$BATS_READLINK" "$1"; then
     return 0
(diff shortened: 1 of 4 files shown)


mbland:
Whoops... I realized that if the `bin/bats` script is executing, the operating system has already traversed all the symbolic links to it, so enforcing our own threshold is pointless. Hence I've removed `BATS_MAX_SYMLINKS` and the tests that exercised it.

mbland:
And today I learned about `set -P`. See my comments for commit 8d9afca for details.

mbland:
Whoops, I posted this to #113 yesterday, meant to post it here:

@bats-core/bats-core barring any suggestions/objections, I'll look to merge this by tomorrow.

PR #10, opened 2017-09-30
Add Appveyor configuration, update Travis CI configuration, fix macOS bug

mbland:
Spawned from #8. Both CI systems will also report test suite run times, and Travis will run macOS builds.

In order to get the macOS build to pass, this also includes a commit to work around the Bash 3.2.57 ERR trap bug. See the commit message for 0f6dde5 for details.

Also worth noting are the build times for this PR:

```
Linux:
real	0m4.248s
user	0m3.488s
sys	0m1.018s

macOS:
real	0m8.898s
user	0m3.798s
sys	0m4.161s

Windows:
real	0m28.079s
user	0m6.764s
sys	0m14.828s
```

vs. those from #8:
```
Linux:
real	0m1.947s
user	0m1.564s
sys	0m0.419s

macOS:
real	0m4.910s
user	0m2.725s
sys	0m1.507s

Windows:
real	0m11.291s
user	0m3.027s
sys	0m5.625s
```

PR #117, opened 2018-07-03
travis: Enable build with default Linux image Bash

mbland:
Since I'm paranoid that we might need to resolve an issue specific to 4.3.11(1)-release (or whatever the Bash version is in the stock Travis Linux image). Also, I didn't want to hold up #116 any longer with more tiny requests.

cc: @cdevinesr

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
 - linux
 
 env:
+  - BASHVER=
   - BASHVER=3.2
   - BASHVER=4.0
   - BASHVER=4.1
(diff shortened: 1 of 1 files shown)


jasonkarns:
👍  I like that this also fixes the "lie" that the macOS build was using a particular bash version.

Side question: why is the host test command `bash -c 'time bin/bats --tap test'` instead of just `time bin/bats --tap test`?

cdevinesr:
Looks like it's been that way for a while:

```
commit bbac787615907ad06214ca2143464396d599ee5a
Author: Mike Bland <mbland@acm.org>
Date:   Sat Sep 30 11:14:48 2017 -0400

    .travis.yml: Prefix test command with `time`

    As with the update to .appveyor.yml in the previous commit, this enables
    us to get an idea of the performance impact of a change.
```

I don't personally see any reason why it couldn't be done the way @jasonkarns suggests.  It's a bit cleaner, too.

mbland:
Heh, the `bash -c` was bugging me, too... Will sneak that in quickly here and merge once the builds pass.

cdevinesr:
Any reason why we couldn't make the same change to the AppVeyor configuration while we're at it?

mbland:
With AppVeyor, we have to have `bash -c` because the default shell is `CMD`, not `bash`; hence, we have to wrap the Bash commands.

jasonkarns:
Ah, I bet that's how it snuck into travis. (copy over from appveyor?)

mbland:
Probably. It was PR #10. Seems I was on autopilot after adding the `time` command to `.appveyor.yml`.

PR #118, opened 2018-07-03
Add error message for bad command line options

### LINKED ISSUE #45: no real error message for bad command line option (from bats/issues/#245 )
Originally reported by https://github.com/ORESoftware
(This issue brought over from https://github.com/sstephenson/bats/issues/245)

this is what I get:

```
$ bats test/src/dev/bats/a.test.sh --tap-json
Bats 0.4.0
Usage: bats [-c] [-p | -t] <test> [<test> ...]
```

this is what I should get:
```

$ bats test/src/dev/bats/a.test.sh --tap-json
Bats 0.4.0
Bad command line option "--tap-json"
Usage: bats [-c] [-p | -t] <test> [<test> ...]
```

let's make it easy for people to use software eh :) not everything needs to be a struggle

mbland:
Closes #45. Closes #46.

Picks up #46 from where @qneill left off, refactoring it to use `abort()` as suggested by @jasonkarns and refactoring the test cases slightly.


libexec/bats
```diff
@@ -128,15 +132,14 @@ if [[ "${#options[@]}" -ne '0' ]]; then
...
+      abort "Bad command line option '-$option'"
       ;;
     esac
   done
 fi
 
 if [ "${#arguments[@]}" -eq 0 ]; then
-  usage "Must specify at least one <test>" >&2
+  abort 'Must specify at least one <test>'
   exit 1
```
cdevinesr: This 'exit 1' on line 143 is superfluous with the introduction of the `abort()` function.

mbland: Whoops...I seemed to miss that in the original commit, but fixed it in the merge.

cdevinesr:
Looks good to me.

libexec/bats-core/bats
```diff
@@ -7,7 +7,13 @@ version() {
 
 usage() {
   version
-  echo "Usage: bats [-c] [-r] [-p | -t] <test> [<test> ...]"
+  printf "Usage: bats [-c] [-r] [-p | -t] <test> [<test> ...]\n"
```
jasonkarns: why the printf here instead of echo?

mbland: Preference based on the general notion that `echo` isn't standardized and `printf` is more consistent (even though both are Bash builtins). You can do a lot more cool stuff with `printf`, too, so I like to promote its awareness.

Some fun reading on the subject: https://unix.stackexchange.com/a/65819

jasonkarns: I'm cool with it, but we should probably move all echos to printfs, no? Instead of here and there?

mbland: Done in #120.

qneill:
Thanks guys for merging, been AFGH (away from github) for a bit.

PR #124, opened 2018-07-09
Automate Release

jasonkarns:
WIP
===

- [x] script/release kicks off the release (accepts major/minor/patch/<exact version>; defaults to patch)
    - [x] invokes script/release-notes
        - [x] fetches latest from git; switches to master
        - [x] ensures there are features to release (non-empty diff of bin, libexec, contrib since last tag)
        - [x] generates .git/RELNOTES from git history
        - [x] release notes are editable in $EDITOR/vim
    - [x] invokes npm version
        - [x] ensures working directory is clean (OOTB)
        - [x] bumps version in package.json (OOTB)
        - [x] `version` lifecycle hook invokes script/version-sync
            - [x] updates appveyor, rpm spec, libexec/bats versions to match package.json
        - [x] `version` lifecycle hook invokes script/version-changelog
            - [x] insert release notes into CHANGELOG.md
        - [x] commits modified files with release notes as message
            - [x] handle markdown headings/git comments
        - [x] signs/tags commit with release notes as message
            - [x] handle markdown headings/git comments
        - [ ] `postversion` lifecycle hook triggers script/publish ?
- [x] script/publish
    - [ ] ensure proper git remote (upstream, origin)
    - [x] pushes to github with tags
    - [x] craft a github release with message
    - [x] publish to npm
    - [x] trigger homebrew PR
    - [ ]  open alpine linux PR
    - [ ] gitter.im announcement
    - [x] remove RELNOTES file
        


jasonkarns:
script/release added to trigger the `npm-version` command. This ensures that the tag is signed, and will provide the point to control the commit/tag message. For now, the tag message is just the version number. Future work is to generate the changelog (editable), and then use that for the commit, the tag, the changelog itself, and the github release.

jasonkarns:
Evidently, git is only stripping markdown headers for the tag, not the commit itself that npm makes. That is... surprising. (I figured they would both behave the same way, regardless.)

And what's more, git-tag doesn't seem to be affected at all by `commit.cleanup`. (And `tag.cleanup` is not a thing.)

cdevinesr:
Regarding switching to master, I think I like the abort idea better, because it's going to fail the switch anyway if you has unstashed changes for some reason.

Cool with the `appveyor` note in the build number.

Also fine with the dependency on node for releases since we're already requiring npm anyway.

jasonkarns:
> Regarding switching to master, I think I like the abort idea better, because it's going to fail the switch anyway if you have unstashed changes for some reason.

So if the switch fails, then it'd abort naturally. OTOH, if the switch succeeds (clean working directory), then we can proceed. (And I'd expect the end of the script to do `git checkout -` to return to whatever branch it started on.)

cdevinesr:
In that case, the auto-switch would be fine.

jasonkarns:
Added my own brew-publish npm package as a devDep for more easily triggering the homebrew PR (I use this for releasing all the nodenv-* packages).

jasonkarns:
FYI, the way I've been testing this is:

1) separate clone of bats-core
2) that has _only_ `origin` remote pointing to my personal fork (not bats-core/bats-core)
3) does _not_ have an `upstream` remote. (`hub` will prefer upstream over origin if it exists)
4) with `master` reset --hard to this PR branch (the release script requires being on master branch; and it also requires clean working copy so can't simply comment out the master-branch-check)

With all these in place, `./script/release && ./script/publish -n` does a real bump/commit/tag, but does a dry-run "publish".

The dry-run publish does a git-push --dry-run; opens a _draft_ release on github (to my fork), and does npm-pack instead of npm-publish; and triggers the homebrew PR process in dry-run mode.

@mbland in particular, we'd probably want you to at least have dry-run this on your machine as well before merging?

About the only thing remaining on this PR, to my mind, is accounting for re-run ability. If any particular publish step fails, we'd probably like to continue the others; And allow the script to be re-run smoothly. Either this means removing set -e, or we do some more intelligent error handling, (or perhaps just `|| true` each of the distinct steps/functions)`

jasonkarns:
Looking into options for automating the AlpineLinux PR as well, similar to homebrew. Of course, since aports doesn't have the brew-bump-pr CLI feature, we'd need to do the branch/change/git-commit/push/pr ourselves.

I have some old scripts that handle all this; the way nodenv and rbenv used to do it for homebrew before brew-bump-pr was a thing. However, the biggest concern is the assumption about where the aports clone would live on one of our machines.

I'm not sure about your structure, but my setup isn't very conventional. I group my clones under their github orgs, but typically only when I have multiple repos for a given org.

So we could find some way to configure the location of the aports clone. (git config, env var, npm config var)

I even temporarily entertained the idea of making aports a submodule of bats-core, which could be configured to do just a shallow-clone, sparse-checkout. That would at least give us guarantees about where the aports clone lives but... blech, submodules.

Maybe we could each do a symlink to the clone? That would eliminate any configuration necessary, but also to have the scripts rely on a fixed location.

Other suggestions?

Bigger suggestion might be having this (and homebrew) PR bumps be done via AWS Lambda or something. Then our scripts just initiate a ping to lambda and wouldn't require aports or homebrew to be cloned locally? Github webhook even?

mbland:
Been slammed, and have a busy-looking weekend and week ahead, but will try to carve out some time to start looking at/playing with this.

script/publish
```diff
@@ -0,0 +1,63 @@
...
+}
+
+cleanup_relnotes() {
+  rm "$RELNOTES"
+}
+
+while getopts ':dn' opt; do
+  case $opt in
+  d) set -x ;;
+  # f) force=--force ;;
```
mbland: Remove vestigial `-f` option.

script/release
```diff
@@ -0,0 +1,19 @@
...
+
+set -euo pipefail
+IFS=$'\n\t'
+
+BUMP=${1:-patch}
+RELNOTES="$(git rev-parse --git-dir)/RELNOTES"
+
+./script/release-notes
+
+printf -v MESSAGE 'Bats %%s\n\n%s' "$(sed -E 's/^#+[ \t]*//' "$RELNOTES")"
```
mbland: Maybe replace `[  \t]` with `[[:space:]]` just for readability's sake. That might render the `-E` unnecessary in the process.

script/release
```diff
@@ -0,0 +1,19 @@
...
+
+npm version --sign-git-tag --message "$MESSAGE" "$BUMP"
```
mbland: Does this bump the tag with the `v` in front? I didn't think `npm version` typically did that. I guess it just parses the last semver tag, bumps only the numbers, and leaves any prefix alone?

script/release-notes
```diff
@@ -0,0 +1,49 @@
...
+}
+
+
+git fetch --quiet --tags
+
+git checkout --quiet master
+
+# This only finds tags reachable from HEAD
+PREV_TAG="$(git describe --tags --match 'v*' --abbrev=0)"
+RELNOTES="$(git rev-parse --git-dir)/RELNOTES"
```
mbland: Are you putting the `RELNOTES` in the `.git` directory to hide it from `git diff`? Perhaps it'd be better to `.gitignore` it instead.

script/release-notes
```diff
@@ -0,0 +1,49 @@
...
+
+assert_features_since "$PREV_TAG"
+
+changes_since "$PREV_TAG" >> "$RELNOTES"
+
+${EDITOR:-vim} "$RELNOTES"
+
+sed -i '' 's/[ \t]*$//' "$RELNOTES"
```
mbland: Replace `[ \t]` with `[[:space:]]` as above.

script/publish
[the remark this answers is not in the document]
jasonkarns: 👍

script/release
jasonkarns: 👍 I think the `-E` is required for the `+` still, though we could get around that with `##*`, right?

script/release
jasonkarns: `npm version` bumps based on the version from package.json, not based on any git tags. It does _create_ (and sign) git tags, but doesn't use them as a basis for bumping.

And the prefix for the git tag is a configurable option: https://docs.npmjs.com/misc/config#tag-version-prefix

script/release
mbland: Oh, is that what needs it? Then feel free to leave it, then.

script/release
mbland: Ohhh...didn't know about that option, thanks!

script/release-notes
jasonkarns: Not specifically to hide it, but rather because the RELNOTES file is temporary working file. I was modeling it after how git itself writes `COMMIT_MSG` (can't remember the exact filename) to `.git/`as well as how `hub` writes its `PR_MSG` there as well. Again, unsure on the exact filename for each of those, but both `git` and `hub` write their files to .git until the full operation is completed at which point, they are cleaned up.

script/release-notes
jasonkarns: 👍

script/release
cdevinesr: You can actually get around the need for `-E` by escaping the `+` with a backslash.  It works similarly to `grep` in that regard.

mbland: Won't backslash-escaping just match the literal `+` character, though, instead of interpreting it as the symbol for "one or more of the previous expression"?

cdevinesr: Not with `sed` if you're not using `-E`.  `sed` actually requires backslash escaping for `+` and parentheses (at minimum - might be others, too) if `-E` is not used.

From a quick demo to show it on an old Ubuntu 14.04 box I have up:

```
root@ubuntu:~# cat /etc/issue
Ubuntu 14.04.4+ LTS \n \l
Testing parens ()
root@ubuntu:~# sed 's/[0-9]+/replaced/g' /etc/issue
Ubuntu 14.04.replaced LTS \n \l
Testing parens ()
root@ubuntu:~# sed 's/[0-9]\+/replaced/g' /etc/issue
Ubuntu replaced.replaced.replaced+ LTS \n \l
Testing parens ()
root@ubuntu:~# sed 's/\(1\)/words/g' /etc/issue
Ubuntu words4.04.4+ LTS \n \l
Testing parens ()
root@ubuntu:~# sed 's/()/words/g' /etc/issue
Ubuntu 14.04.4+ LTS \n \l
Testing parens words
root@ubuntu:~# sed -E 's/[0-9]+/replaced/g' /etc/issue
Ubuntu replaced.replaced.replaced+ LTS \n \l
Testing parens ()
root@ubuntu:~# sed -E 's/(1)/words/g' /etc/issue
Ubuntu words4.04.4+ LTS \n \l
Testing parens ()
root@ubuntu:~#
```

mbland: Wow. Very good to know. Just when I thought I knew regexes, and `sed`...

jasonkarns: I'm simultaneously happy to TIL, and sad that such a "basic" thing still has so much TIL-able left...

Speaking more generally though, I see the appeal to avoid `-E` because of broad appeals to minimalism. But at the same time, I feel like using `-E` _always_ reduces the risk for future gotchas. (I'm assuming that most everyone has a more solid understanding of the extended form, and are less likely to introduce bugs if all regexes are already forcing extended mode.) Are there any benefits to avoiding `-E`?

script/release-notes
jasonkarns: If you consider just those 2 use cases a convention. 🤷‍♂️ was enough of a pattern that I thought made a good excuse to keep it out of repo root 😸

script/release
mbland: I don't mind keeping it, now that I'm surprised to learn that certain basic regex constructs actually _require_ it to work as I'd expected.

cdevinesr: I don't think it's a problem.  As far as I am aware `-E` (for both `sed` and `grep`) just changes which regex engine is used.  I don't think that opens us up to anything particularly nasty like `echo`'s `-e` does.

cdevinesr: I'd also like to note that it was implemented in this manner for the same reason many things are done the way they are with core tools on *nix systems: to maintain backwards compatibility with scripts.

script/release-notes
mbland: Well, as use cases go, `git` and `hub` seem fairly mainstream enough to follow. 😉

mbland:
@jasonkarns Need anything else from me at the moment before moving forward?

sublimino:
@jasonkarns I wish I'd seen this before https://github.com/bats-core/bats-core/pull/284/files which isn't as comprehensive :facepalm: I'll look at merging the two of them.

From first glance these scripts need the Homebrew and NPM credentials (any others?), are they your personal tokens or could they be moved into this org somehow? Keybase?

jasonkarns:
@sublimino I'd honestly like to start moving back away from leveraging npm for this. It was originally created before GitHub Actions was a thing, so a lot of the work of publishing to homebrew, github, npm and the like all had to be done "locally". Now that can all be automated on GitHub Actions. Which leaves the only real benefit from npm being the literal version bumping and git-tagging.

As I start to incorporate this into github actions, I'll transition this over to a Makefile likely.

sublimino:
@jasonkarns I've added a Bash `semver` and wrapper script in https://github.com/bats-core/bats-core/pull/284 — there's not so much in there right now though. 

Would support Makefile, and where Make breaks (I'm using the `.ONELINE` extension a lot these days) we can just shell out to support scripts to avoid horrid nested escaping :facepalm:

Issue #49, opened 2017-12-24
Print the time of each test and the complete suite


BraisGabin:
The tests must be fast. This way the developers will run tests more often and the bugs will decrease.

Print the time of each tests and the time of the complete suite helps to find the slow tests and fix them.

tterranigma:
I would like this functionality but behind a flag, eg `-b | --benchmark`.

gnikonorov:
If this is still up for grabs, I would like to implement this feature.

jasonkarns:
I'm probably 👍 on this as well. Though I might suggest `-p | --profile` which could be further enhanced to take numeric argument to print out the N slowest tests. prior art: https://relishapp.com/rspec/rspec-core/v/3-7/docs/configuration/profile-examples

To be clear, I'm suggesting the initial feature just print time per test as the original poster suggests, but the `-p` would more naturally lead to the additional profiling output.

gnikonorov:
Thanks @jasonkarns, I'll work on this over the next few days.

mbland:
The tricky thing with capturing the timing is that I don't know of a way off the top of my head to grab the output of `time` without creating a subshell, in order to report individual test case timing. We try to keep subshells to a minimum for performance reasons, which has an outsized effect on Windows platforms in particualr. So just keep that constraint in mind as you start wrestling with the problem.

And another thought to keep in mind: If there are a couple of slow tests dragging down the entire time of the suite, it's probably going to be obvious which ones they are. On the other hand, there's value in having the timing of the entire suite, to see if specific optimizations produce gains across the board (which is why I run `time bats` all the time).

It'd still be nice to have a switch to have Bats itself run `time` for the entire suite (since we hit complications running `time bats` in #116), which might be an easier first problem to solve.

gnikonorov:
Thank you for the background @mbland, I'll look for an efficient way to solve this.

gnikonorov:
Sorry for the delay on this, I've been caught up with a few other things. I'm polishing up a solution to timing of the entire suite without the `time` command ( to avoid subshells ) now.

gnikonorov:
@mbland - I am not that familiar with `TAP` format. Would we want to suppress this functionality if we are printing out in `TAP` format? Or is it valid to include under `TAP` format as well.

cdevinesr:
TAP has support for comments.  Timing information can be included in TAP output if the line begins with `#`.

gnikonorov:
Thank you @cdevinesr !

[...]

PR #127, opened 2018-07-18
Add duration tracking option for each test

gtkramer:
In some validation cases, it makes sense to track the duration of each
test to understand which ones are executing quickly and which ones are
executing slowly.  Trends may be built up over time to see how duration
changes given the other changes being tested.

Rebasing from harschware/bats/yaml_durations.

Because this adds a YAML block to the TAP output, the TAP producer needs
to write "TAP version 13" as the first line.

Signed-off-by: George T Kramer <george.t.kramer@intel.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ supports:
 
 ```
 Bats x.y.z
-Usage: bats [-c] [-r] [-p | -t] <test> [<test> ...]
+Usage: bats [-c] [-r] [-p | -t] [-d] <test> [<test> ...]
 
   <test> is the path to a Bats test file, or the path to a directory
   containing Bats test files.
@@ -190,6 +190,7 @@ Usage: bats [-c] [-r] [-p | -t] <test> [<test> ...]
   -p, --pretty     Show results in pretty format (default for terminals)
   -r, --recursive  Include tests in subdirectories
   -t, --tap        Show results in TAP format
+  -d, --duration   Track duration of each test in YAML when using TAP format
   -v, --version    Display the version number
 ```
 
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -7,7 +7,7 @@ version() {
 
 usage() {
   version
-  printf 'Usage: bats [-c] [-r] [-p | -t] <test> [<test> ...]\n'
+  printf 'Usage: bats [-c] [-r] [-p | -t] [-d] <test> [<test> ...]\n'
 }
 
 abort() {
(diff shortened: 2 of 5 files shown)


gtkramer:
The Contributor Guidelines mention quoting all variables.  I'm not sure if I should do this, or remain consistent with the existing code's usage of _flags variables...

BraisGabin:
Related with #49 right?

gtkramer:
@BraisGabin, yes, this is related to #49.  This adds duration when outputting test results in the TAP format, but not the regular format.  A YAML document was chosen because some TAP consumers, such as Jenkins with it's TAP plugin, look for duration in this structured manner to display it in a CI system.  Diagnostic information, in my experience, is more useful in the TAP format to show a backtrace when a test failed.

Looks like I need to rebase.  I'll work on doing this.

phillipao:
+1, would like to use this. @gtkramer, what's the status?

gzhifang:
would like to use it . Can anyone merge this PR?

sublimino:
Hi @gtkramer, sorry this has been open so long. 

Could you rebase master onto this branch to resolve the conflicts please?

I wonder about optional `Tap 13` syntax. Perhaps we should migrate entirely to Tap 13 instead of just for one flag's output? WDYT @mbland @jasonkarns?

yahliu:
Looking forward to use this function. Could someone please help to review and merge it?

sublimino:
@yahliu due to the time elapsed since submission (apologies!) these changes conflict with master and need rebasing to be merged. 

If somebody is able to do that work we can get this merged. Feel free to discus here first, at first glance:
- there are now multple output formats that each require this patched in
- tests for each format
- I'm not sure if [this](https://github.com/bats-core/bats-core/pull/127/files#diff-9a8fb8c717871504ab2c0035d4a6fe4fR372) will conflict with other work on traps recently

yahliu:
@sublimino  I see latest code already have the time option `-T, --timing`, but it didn't shown in the Jenkins TAP plugin.
Screen Shot 2020-07-01 at 3 53 58 PM

Based on my checking, the TAP plugin will get keywords `duration_ms`, but `-T, --timing` have no such it. so I think it's better to enhance `-T, --timing` to support it.

martin-schulze-vireso:
I am looking into getting this to fly with the current codebase. My observations:

1. the current option measures in full seconds, this one uses milliseconds
2. I see big problems with maintaining both TAP 1.0 and TAP 1.3 output in parallel, as @sublimino  already said it would probably be benificial to go full TAP 1.3 then. However, this would require large redesigns of the test suite which relies on the old TAP format
3. using TAP 1.3 internally would require a yaml parser or some very strict formatting to apply formatters

My proposal would be: 
1. update the timing code to use millisecond precision
2. Add a Tap1.3 formatter for consumption by Jenkins TAP, this could even be an external project

martin-schulze-vireso:
I am closing this since #337 has been merged.

PR #137, opened 2018-08-09
WIP: Add a --trace feature

### LINKED ISSUE #136: Add a test trace option
Similar to the execution trace feature (enabled via `set -x`) in many shells, it would be nice to have a similar feature in BATS to enable "test tracing". Since setting `set -x` in a BATS test script is too verbose to be useful, I would like to see a test trace print out the series of commands executed in the course of a test execution, ignoring all BATS-internal routines.

mbland:
Closes #136. The final implementation will set the BATS_TRACE variable via a command-line flag, but setting `BATS_TRACE=1 bin/bats ...` will work.

This did require explicitly filtering out the line from `bats_perform_test` that executes the test case; somehow the `"$BATS_SOURCE" != "$1"` condition doesn't filter it. If I (or anyone else) can think of a cleaner way to filter out this line than quoting it directly, I'm all for it.

Two minor caveats:

- If a command fails, it will be printed twice. Then, of course, it'll appear in the failure output, for a total of three times.
- When using the pretty formatter, the first line of the trace from a test case will overlap with the line printing the name of the test case before execution begins.

Also, I still need to add tests, of course.

cc: @gtkramer

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -20,7 +20,7 @@ usage() {
   while IFS= read -r line; do
     printf '%s\n' "$line"
   done <<END_OF_HELP_TEXT
-Usage: $cmd [-cr] [-f <regex>] [-p | -t] <test>...
+Usage: $cmd [-crx] [-f <regex>] [-p | -t] <test>...
        $cmd [-h | -v]
 
   <test> is the path to a Bats test file, or the path to a directory
@@ -33,6 +33,7 @@ Usage: $cmd [-cr] [-f <regex>] [-p | -t] <test>...
   -r, --recursive  Include tests in subdirectories
   -t, --tap        Show results in TAP format
   -v, --version    Display the version number
+  -x, --trace      Display Bash commands with file:line info as they're executed
 
   For more information, see https://github.com/bats-core/bats-core
 
@@ -61,7 +62,7 @@ export PATH="$BATS_ROOT/libexec/bats-core:$PATH"
 
 arguments=()
 
-# Unpack single-character options bundled together, e.g. -cr, -pr.
+# Unpack single-character options bundled together, e.g. -cr, -pr, -tx.
 for arg in "$@"; do
   if [[ "$arg" =~ ^-[^-]. ]]; then
     index=1
@@ -114,6 +115,9 @@ while [[ "$#" -ne 0 ]]; do
     ;;
   -p|--pretty)
     pretty=1
+    ;;
+  -x|--trace)
+    flags+=('--trace')
     ;;
   -*)
     abort "Bad command line option '$1'"
--- a/libexec/bats-core/bats-exec-suite
+++ b/libexec/bats-core/bats-exec-suite
@@ -16,6 +16,9 @@ while [[ "$#" -ne 0 ]]; do
     flags+=('-f' "$filter")
     shift
     ;;
+  --trace)
+    flags+=('--trace')
+    ;;
   -x)
     extended_syntax_flag='-x'
     flags+=('-x')
--- a/libexec/bats-core/bats-exec-test
+++ b/libexec/bats-core/bats-exec-test
@@ -222,34 +222,30 @@ bats_trim_filename() {
 }
 
 bats_debug_trap() {
-  local status="$?"
-
   if [[ "$BASH_SOURCE" != "$1" ]]; then
     bats_capture_stack_trace
     BATS_LINENO="$BATS_CURRENT_LINENO"
     BATS_CURRENT_LINENO="${BASH_LINENO[0]}"
 
     if [[ -n "$BATS_TRACE" ]]; then
-      bats_emit_trace "$status"
+      bats_emit_trace
     fi
   fi
 }
 
 bats_emit_trace() {
-  local status="$?"
-
   case "$BASH_COMMAND" in
   '"$BATS_TEST_NAME" >> "$BATS_OUT" 2>&1'|'bats_test_begin '*)
     return
     ;;
   esac
 
-  # If `status` is nonzero but the command is the same, then this DEBUG trap
-  # is firing before the ERR trap of a failing command. Since we've printed
+  # If the `BASH_COMMAND` is the same, but `BATS_CURRENT_LINENO` is less than or
+  # equal to `BATS_LINENO` (the previous trap's `BASH_LINENO`), then this DEBUG
+  # trap is firing before the ERR trap of a failing command. Since we've printed
   # the command before it failed, we don't want to print it a second time.
-  if [[ "$status" -eq 0 && \
-        ( "$BASH_COMMAND" != "$BATS_PREVIOUS_COMMAND" || \
-          "$BATS_LINENO" -lt "$BATS_CURRENT_LINENO" ) ]]; then
+  if [[ "$BASH_COMMAND" != "$BATS_PREVIOUS_COMMAND" || \
+        "$BATS_LINENO" -lt "$BATS_CURRENT_LINENO" ]]; then
     local filename="${BASH_SOURCE[2]}"
 
     if [[ "$filename" == "$BATS_TEST_SOURCE" ]]; then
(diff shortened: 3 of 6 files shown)


mbland:
OK, tests are added; this PR is ready for review. From the last commit message:

This final version will filter out any stack trace values and trace output where:

- the `DEBUG` trap fires for the function definition itself, resulting in redundant --trace output lines (comparing `BASH_COMMAND` against `${FUNCNAME[1]}`)

- the `DEBUG` trap fires before the `ERR` trap on error; apparently when both functions fire, `$?` refers to the value returned by the last command before either trap is fired (checking `$status` before emitting the trace output)

The use of `case` patterns instead of multiple tests in an `if` condition is a bit more compact, and very slightly faster on average. Hence, I took the opportunity to apply a `case` statement in `bats_capture_stack_trace` as well.

mbland:
I pushed a new commit to cover recursive function calls and `else` clause commands; see the message from commit 73952a6bac55f5f6f64fedef200d65c50d2efd07 for details.

However, test case 57, "trace a failing test case", now fails consistently on Bash 4.0.44(1)-release, 4.1.17(2)-release, and 4.2.53(2)-release. I'll have to investigate and fix a little later.

Regarding the slight performance regression mentioned in the last commit message, I'm thinking of leaving it for now, as performance overall doesn't seem to suffer from the new feature. However, I do have ideas for making a more radical performance improvement—I _think_ I see a way to get away with only capturing the stack trace on error—but I'll try it in another PR.

mbland:
FYI, I just filed #138 with some stack trace optimizations. It doesn't eliminate the need to capture the stack trace on every call, but eliminates some unnecessary calls and makes the code a little cleaner.

I'll likely try to merge that PR first and update this one, and then attempt to fix the breakages. Still, @gtkramer, the semantics won't change, and I'd be interested in hearing some feedback once you try out this branch in the meanwhile.

mbland:
The strangest thing is happening with these failures under Bash 4.0.44, 4.1.17, and 4.2.53. When I run the tests under those versions locally, adding `emit_debug_output` to the failing test case, I get:

```
$ time PATH="/usr/local/bash/4.2.53/bin:$PATH" bin/bats test -f 'trace a failing'                                     
 ✗ trace a failing test case
   (in test file test/bats.bats, line 488)
     `[ "${#lines[@]}" -eq 8 ]' failed
   output:
   1..1
   # test/fixtures/bats/trace.bats:6: run printf 'The TEST_RESULT is: %s\n' "$TEST_RESULT"                                                                      
   # test/fixtures/bats/trace.bats:7: evaluate_result "$output"
   # test/fixtures/bats/trace.bats:2: [ "$1" = 'The TEST_RESULT is: PASS' ]
   # test/fixtures/bats/trace.bats:257: [ "$1" = 'The TEST_RESULT is: PASS' ]
   not ok 1 traced test case
   # (from function `evaluate_result' in file test/fixtures/bats/trace.bats, line 2,                                                                            
   #  in test file test/fixtures/bats/trace.bats, line 7)
   #   `evaluate_result "$output"' failed

1 test, 1 failure


real    0m0.412s
user    0m0.283s
sys     0m0.085s
```

That last line number, `257`, corresponds to the last command executed by `bats_debug_trap`, which is the last line executed by `bats_emit_trace`. I know this because trying different commands in those functions changes that line number.

This appears to be attributable to the fact that starting with Bash 4.3, the `line_number_for_err_trap` variable in `execute_cmd.c` went from being declared a `static int` to a global `int`, so that `parse_prologue` in `builtins/evalstring.c` could wrap it in an `unwind_protect_int` call. (In Bash 3.2.57, the problem was different, in that the `BASH_LINENO` value would get clobbered with the number of the first line of the function.)

This is tricky to work around. I'll let it sit and think about it some more, or whether it's worth adjusting/skipping the test under Bash 4.0, 4.1, and 4.2. Ideas welcome.

mbland:
Some notes to myself about how to handle this:

* Perhaps arrange it so the trace is printed for the _previous_ command, rather than the current one. Would need to take care to see how that would play with teardown, however.
* Let go of trying to filter out the beginning-of-function-definition traces.

andrewmiller1:
<details><summary>It works on 5 beta as well</summary>

```
andrew@pop-os:~/opt/bats-core (trace * u= origin/trace)
$ bash -c 'echo $BASH_VERSION'
5.0.0(1)-beta
andrew@pop-os:~/opt/bats-core (trace * u= origin/trace)
$ time bin/bats --tap test
1..76
ok 4 -h and --help print help
ok 5 invalid filename prints an error
ok 6 empty test file runs zero tests
ok 7 one passing test
ok 8 summary passing tests
ok 9 summary passing and skipping tests
ok 10 tap passing and skipping tests
ok 11 summary passing and failing tests
ok 12 summary passing, failing and skipping tests
ok 13 tap passing, failing and skipping tests
ok 14 BATS_CWD is correctly set to PWD as validated by bats_trim_filename
ok 15 one failing test
ok 16 one failing and one passing test
ok 17 failing test with significant status
ok 18 failing helper function logs the test case's line number
ok 19 test environments are isolated
ok 20 setup is run once before each test
ok 21 teardown is run once after each test, even if it fails
ok 22 setup failure
ok 23 passing test with teardown failure
ok 24 failing test with teardown failure
ok 25 teardown failure with significant status
ok 26 failing test file outside of BATS_CWD
ok 27 load sources scripts relative to the current test file
ok 28 load aborts if the specified script does not exist
ok 29 load sources scripts by absolute path
ok 30 load aborts if the script, specified by an absolute path, does not exist
ok 31 output is discarded for passing tests and printed for failing tests
ok 32 -c prints the number of tests
ok 33 dash-e is not mangled on beginning of line
ok 34 dos line endings are stripped before testing
ok 35 test file without trailing newline
ok 36 skipped tests
ok 37 skipped test with parens (pretty formatter)
ok 38 extended syntax
ok 39 pretty and tap formats
ok 40 pretty formatter bails on invalid tap
ok 41 single-line tests
ok 42 testing IFS not modified by run
ok 43 expand variables in test name
ok 44 handle quoted and unquoted test names
ok 45 ensure compatibility with unofficial Bash strict mode
ok 46 parse @test lines with various whitespace combinations
ok 47 duplicate tests cause a warning on stderr
ok 48 sourcing a nonexistent file in setup produces error output
ok 49 referencing unset parameter in setup produces error output
ok 50 sourcing a nonexistent file in test produces error output
ok 51 referencing unset parameter in test produces error output
ok 52 sourcing a nonexistent file in teardown produces error output
ok 53 referencing unset parameter in teardown produces error output
ok 54 execute exported function without breaking failing test output
ok 55 output printed even when no final newline
ok 56 trace a passing test case
ok 57 trace a failing test case
ok 58 trace a test case calling a recursive function
ok 59 install.sh creates a valid installation
ok 60 install.sh only updates permissions for Bats files
ok 61 bin/bats is resilient to symbolic links
ok 62 #113: set BATS_ROOT when /bin is a symlink to /usr/bin
ok 63 set BATS_ROOT with extreme symlink resolution
ok 64 running a suite with no test files
ok 65 running a suite with one test file
ok 66 counting tests in a suite
ok 67 aggregated output of multiple tests in a suite
ok 68 a failing test in a suite results in an error exit code
ok 69 running an ad-hoc suite by specifying multiple test files
ok 70 extended syntax in suite
ok 71 recursive support (short option)
ok 72 recursive support (long option)
ok 73 run entire suite when --filter isn't set
ok 74 use --filter to run subset of test cases from across the suite
ok 75 --filter can handle regular expressions that contain [_- ]
ok 76 --filter can handle regular expressions that start with ^

real	0m6.437s
user	0m4.325s
sys	0m2.257s
andrew@pop-os:~/opt/bats-core (trace * u= origin/trace)
$ 
```

</details>

grahamwhaley:
Just wondering, if this PR stuck? I had 'fun' following the history of requests for a trace/verbosity/debug feature for `bats`. It is **so** needed. So much that without it ultimately I'd start pushing folks off bats, as trying to debug broken or flaky tests without some built in debug feature is just too much pain :-)
Would love to see this land :-)

NorseGaud:
Bump

martin-schulze-vireso:
I am closing this in favor of #467

PR #138, opened 2018-08-13
Optimize stack trace capture

mbland:
With these changes, the current test suite runs O(0.2s-0.3s) faster on a Mid 2015 Macbook Pro with a 2.5GHz Intel Core i7 with 16GB RAM. Not all of the optimizations have the same degree of impact, but all make the code a little less complex and a bit more readable.

Most notably, there is now only the `BATS_STACK_TRACE` variable, rather than `BATS_CURRENT_STACK_TRACE`, `BATS_PREVIOUS_STACK_TRACE`, and `BATS_ERROR_STACK_TRACE`. The only difference between the two previous stacks was the line number of the frame at the top of each. The stack trace is captured both before a command runs and after it triggers the `ERR` or `EXIT` trap.  Depending on the Bash version or type of error, the `BASH_LINENO` would change when capturing the second stack trace. We now manage that difference via `BATS_LINENO` and `BATS_CURRENT_LINENO`.

Note that these timings are a little different than the ones in the commit messages, since I rebooted the machine before collecting these in order to ensure there was as little interference as possible from other processes and memory pressure.

Bash 3.2.57(1)-release before:

```
  real    0m6.532s
  user    0m3.351s
  sys     0m1.939s
```

Bash 3.2.57(1)-release after:

```
  real    0m6.223s
  user    0m3.046s
  sys     0m1.934s
```

Bash 4.2.23(1)-release before:

```
  real    0m6.893s
  user    0m3.290s
  sys     0m2.069s
```

Bash 4.2.23(1)-release after:

```
  real    0m6.660s
  user    0m3.081s
  sys     0m2.053s
```

### DIFF
--- a/libexec/bats-core/bats-exec-test
+++ b/libexec/bats-core/bats-exec-test
@@ -100,34 +100,32 @@ bats_test_function() {
   BATS_TEST_NAMES+=("$test_name")
 }
 
-BATS_CURRENT_STACK_TRACE=()
+# Avoid a bug in Bash < 4.4-rc2 where "${BATS_CURRENT_STACK_TRACE[@]}" will
+# throw an unbound variable error when empty by initializing it with 'main'.
+BATS_CURRENT_STACK_TRACE=('main')
 BATS_PREVIOUS_STACK_TRACE=()
 BATS_ERROR_STACK_TRACE=()
 
 bats_capture_stack_trace() {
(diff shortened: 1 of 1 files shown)


mbland:
Merging this now to get things moving; per usual, comment or open a new issue if any of these changes provide cause for concern.

PR #140, opened 2018-08-15
Add 'background' convenience function

cdevinesr:
Some tests could require something else to run asynchronously.

Since there is often confusion around FD 3 and how TAP output works
in BATS, this helper function would allow for easier backgrounding
of tasks within tests.

This is less of a need-to-have thing, and more of a nice-to-have thing.  I've encountered some test cases at work where folks were confused by how I was handling backgrounding processes, so I figured a simple utility function to handle this would be acceptable to make this behavior a bit more accessible.

This is mostly because I know there are people using BATS (like my boss) who are not terribly familiar with file descriptor manipulation in Bash, and this will make it clearer what is being done (and less likely that someone backgrounding a process forgets to close FD 3 when doing so).

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ commit [0360811][].  It was created via `git clone --bare` and `git push
 - [Writing tests](#writing-tests)
   - [`run`: Test other commands](#run-test-other-commands)
   - [`load`: Share common code](#load-share-common-code)
+  - [`background`: Run a command asynchronously](#background-run-a-command-asynchronously)
   - [`skip`: Easily skip tests](#skip-easily-skip-tests)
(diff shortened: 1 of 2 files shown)


mbland:
Forgot about this one...this would be a welcome addition, when tested. ;-)

Note I also implemented (and tested) a similar mechanism here, in case it makes sense to merge them: https://github.com/mbland/custom-links/blob/master/tests/scripts/serve.bats#L69-L156 (per https://github.com/mbland/custom-links/issues/141).

cdevinesr:
Yours seems much more thought out.  Mine was more of trying to quickly work around a particular annoyance that I knew would likely trip up others on my team (have had my boss ask me about this particular "problem" at least once already).

PR #146, opened 2018-08-24
add documentation around 'run' executing a subshell

spikegrobstein:

this has bitten me a few times over the years since I always forget that `run` is executed in a subshell. since many of my tests are for environmental side-effects, it would be good to call this out in the docs.


sublimino:
LGTM @spikegrobstein, thanks :+1: 	

Wondering if you use a general fix when testing this issue, like writing to temporary files or FDs?

PR #169, opened 2019-01-15
colorize pass and skip tests

bruce-szalwinski:
- [x ] I have reviewed the [Contributor Guidelines][contributor].
- [x ] I have reviewed the [Code of Conduct][coc] and agree to abide by it

### DIFF
--- a/libexec/bats-core/bats-format-tap-stream
+++ b/libexec/bats-core/bats-format-tap-stream
@@ -36,6 +36,7 @@ begin() {
 
 pass() {
   go_to_column 0
+  set_color 2 bold
   buffer ' ✓ %s' "$name"
   advance
 }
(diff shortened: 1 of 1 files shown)


bruce-szalwinski:
I went with the same colors for pass and skip that Cucumber defaults to, https://github.com/cucumber/cucumber/wiki/Console-Colours.

bruce-szalwinski:
I couldn't figure out a way to test that output was colorized.  I didn't see a way to test that escape sequences were in the `output`.  I see that `output` from tests is run through `filter_control_sequences`.  I tried removing the `filter_control_sequences` from the test, but `output` didn't contain any control sequences.  I'll take another pass at this and see if I can come up with something.

PR #172, opened 2019-02-02
bats: support parallel execution of tests

### LINKED ISSUE #171: running tests in parallel
Currently tests are all run in serial, which results in suboptimal test running performance (in the case of `umoci` we've started to hit against the 50 minute limit that Travis has because we run ~200 tests that all take a fairly long amount of time -- but are easily parallelisable).

There was an issue about this in sstephenson/bats#143, but I think the easiest solution would be to use GNU parallel if it's available (which knows how to collate data between different programs and can also do a bunch of other clever things when running). In theory the fix should be quite trivial (a minor rewrite of `bats-exec-suite` to use GNU parallel for running `bats-exec-test`).

cyphar:

This allows for tests to be run in parallel, by using GNU parallel (if
available). It's a fairly trivial extension (though it required a fair
amount of work in order for it to be usable).

Most well-behaved bats tests (where they aren't writing to shared data
between test executions) are able to be executed in parallel, but to
date bats has always executed them in serial. This is incredibly
wasteful for trivially parallelisable tests, and can result in fairly
inflated test times for no good reason. Parallel exeuction is done
suite-wide, so it makes no difference how the tests are structured.
    
Unfortunately, because some suites might be implicitly serial (like
bats' own test suite) we cannot make the default mode execute in
parallel. In addition, there is currently no way of tagging that a
*specific* test has to be run in serial (so that the rest can be run in
parallel) -- but that can be done in future work.

Fixes #171 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,12 @@ ARG bashver=latest
 
 FROM bash:${bashver}
 
-RUN ln -s /opt/bats/bin/bats /usr/sbin/bats
+# Install parallel and accept the citation notice (we aren't using this in a
+# context where it make sense to cite GNU Parallel).
+RUN apk add --no-cache parallel && \
+    mkdir -p ~/.parallel && touch ~/.parallel/will-cite
 
+RUN ln -s /opt/bats/bin/bats /usr/sbin/bats
 COPY . /opt/bats/
 
 ENTRYPOINT ["bash", "/usr/sbin/bats"]
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ supports:
 
 ```
 Bats x.y.z
-Usage: bats [-cr] [-f <regex>] [-p | -t] <test>...
+Usage: bats [-cr] [-f <regex>] [-j <jobs>] [-p | -t] <test>...
        bats [-h | -v]
 
   <test> is the path to a Bats test file, or the path to a directory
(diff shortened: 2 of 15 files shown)


cyphar:
Any updates? @sublimino?

sublimino:
This is good to go as far am I am concerned, loads of good discussion on https://github.com/bats-core/bats-core/issues/171 and those who have run existing test suites report that they are unaffected. 

Thanks for the contribution @cyphar :pray:

cyphar:
Neat. I will send the randomised ordering PR next.

patthiel:
This commit: https://github.com/bats-core/bats-core/commit/97975a05b22a0026357cbbe320c00c296f1cf056

Introduces `greadlink` as a dependency, which isn't included on osx. It results in errors running bats. 

Stacktrace for those living on the bleeding edge running from source:

```bash
/usr/bin/readlink: illegal option -- f
usage: readlink [-n] [file ...]
```


 `brew install coreutils` resolved issues running from source for me.

cyphar:
Yes. I was considering re-implementing `readlink -f` in bash to not have the dependency but it's really quite annoying to to do correctly. I can give it a shot though.

cyphar:
Scratch that, I just figured out I can implement it with a really hacky use of `paste`.

PR #177, opened 2019-02-06
fix doc error in fd 3 redirect-and-close example (fixes #176)

### LINKED ISSUE #176: documentation error: redirection '3>-' creates file named '-'
In the docs at https://github.com/bats-core/bats-core#file-descriptor-3-read-this-if-bats-hangs it says 

````
...
To prevent this from happening, close FD 3 explicitly when running any 
command that may launch long-running child processes, 

   e.g.  command_name 3>- &.
```` 

But that creates an empty file with the name `-`. I think the fix is to use

`command_name 3>&-`

as documented at https://www.tldp.org/LDP/abs/html/io-redirection.html with the example

`exec 3>&-                 # Close fd 3.`

joshrabinowitz:
Fixes documentation error regarding https://github.com/bats-core/bats-core#file-descriptor-3-read-this-if-bats-hangs at https://github.com/bats-core/bats-core/issues/176 
(closes #176)

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,7 +367,7 @@ service that will run indefinitely), Bats will be similarly blocked for the same
 amount of time.
 
 **To prevent this from happening, close FD 3 explicitly when running any command
-that may launch long-running child processes**, e.g. `command_name 3>- &`.
+that may launch long-running child processes**, e.g. `command_name 3>&-` .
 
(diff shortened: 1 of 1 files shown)


sublimino:
Good spot @joshrabinowitz, thanks for the fix :+1:

PR #179, opened 2019-02-07
Consistent 'note' format in readme

jasonkarns:

- Updates the two existing notes to have a consistent style: (`Note:` in bold preceding the text; as a regular paragraph, no blockquote)
- Adds the 'note' prefix to one that was missing it.
- Moves an existing note to the bottom of its section. (Was in between the main text and the examples; read as if the examples were related to the note itself.)
- installed toc-me github app to keep the table of contents up to date automatically
- removed the existing Table of Contents and placed the toc-marker in its place
- toc-me bot ran and re-generated the table of contents. The auto-generated TOC uses different markers for the different list item levels so there is a diff of markdown source. However, no visual difference exists in the rendered html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,35 @@ commit [0360811][].  It was created via `git clone --bare` and `git push
 
 <!-- toc -->
 
+- [Installation](#installation)
+  * [Supported Bash versions](#supported-bash-versions)
+  * [Homebrew](#homebrew)
+  * [npm](#npm)
+  * [Installing Bats from source](#installing-bats-from-source)
+  * [Running Bats in Docker](#running-bats-in-docker)
+    + [Building a Docker image](#building-a-docker-image)
+- [Usage](#usage)
+- [Writing tests](#writing-tests)
+  * [`run`: Test other commands](#run-test-other-commands)
+  * [`load`: Share common code](#load-share-common-code)
+  * [`skip`: Easily skip tests](#skip-easily-skip-tests)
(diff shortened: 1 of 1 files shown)


sublimino:
LGTM :+1: 

I hadn't seen that bot before. Can it be scoped to just this file or does it have repo-wide access?

jasonkarns:
I don't think they can be scoped since it has push access to the repo. The
author is a githubber, if that makes a difference  😃

sublimino:
> The author is a githubber, if that makes a difference

Hehe it does I suppose. 

Triggering a broader thought - I don't have access to this repo's settings to verify/enable protection for the master branch. As long as the bot can't force push we're fine - do you have access to verify @jasonkarns?

jasonkarns:
@sublimino Unfortunately, I don't believe we can limit force-pushing via the oauth scopes for the app. Which leaves protected branches as the only option. And I have concerns with that approach.

Our current process is to do releases from a "bump" commit on master, which is then tagged and pushed directly to master. (As opposed to having releases tagged from a PR which is merged.)

If we protect master from direct pushing and require PRs, we would have to change the release tagging process.

(Though after pulling up the protected branch options; it appears there's a new option for explicitly listing people who have access to push to a particular branch. And also it seems that admins are not confined by the branch protections by default. So we do have a couple alternatives for increasing the security on master.)

jasonkarns:
Just following up, have enabled branch restrictions so toc-me can't push (nor force push) to master.

PR #181, opened 2019-02-11
test: add Bash 5 as test target

sublimino:
This PR adds the stable Bash 5 release.

It also corrects variable quoting inconsistencies, and adds `set -e` to the test script in case of a failure to build the image. This is documented as a [possible breaking flag to Travis (?!)](https://docs.travis-ci.com/user/common-build-problems/#my-build-fails-unexpectedly) but passes happily in my tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,10 @@ services:
 
 script:
 - |
-  set -e
   if [[ "${TRAVIS_OS_NAME:-}" == 'linux' && -n "${BASHVER}" ]]; then
-    docker build --build-arg bashver="${BASHVER}" --tag "bats/bats:bash-${BASHVER}" .
-    docker run -it "bash:${BASHVER}" --version
-    time docker run -it "bats/bats:bash-${BASHVER}" --tap /opt/bats/test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -21,10 +22,11 @@ services:
 
 script:
 - |
-  if [[ "$TRAVIS_OS_NAME" == 'linux' && -n "$BASHVER" ]]; then
-    docker build --build-arg bashver=${BASHVER} --tag bats/bats:bash-${BASHVER} .
-    docker run -it bash:${BASHVER} --version
-    time docker run -it bats/bats:bash-${BASHVER} --tap /opt/bats/test
+  set -e
```
jasonkarns: Not a blocker for this PR, but we should consider re-working the travis config to use their new [conditions](https://docs.travis-ci.com/user/conditions-v1) branching

Flamefire:
`set -e ` in the travis file should not be used as it will supress any output on failure. Either chain commands with `&&` or put them in a script and add the `set -e` there. You could even get rid of the `TRAVIS_OS_NAME` check (set `env: BASHVER=` directly in the include if you want to be explicit)

This way a developer can just run this bash script (potentially in a loop) to simply test this.

PR #183, opened 2019-02-17
Fix link resolution for bats

waterkip:
This rewrites the old logic a bit by using `type -p`, `readlink -m`  and
on boxes where we don't have coreutils like `readlink` we use `realpath`
to figure out the correct installation directory of bats.

I've also changes the path of the symbolic link so it works on the
default bash image without having sbin (bats isn't a sbin kinda thing).

Also fixes GH#182

Signed-off-by: Wesley Schwengle <wesley@schwengle.net>

### DIFF
--- a/bin/bats
+++ b/bin/bats
@@ -3,42 +3,71 @@
 set -e
 
 BATS_READLINK='true'
+BATS_REALPATH=''
 if command -v 'greadlink' >/dev/null; then
   BATS_READLINK='greadlink'
 elif command -v 'readlink' >/dev/null; then
   BATS_READLINK='readlink'
 fi
 
+set +e
+if ! $BATS_READLINK -m $0 &>/dev/null; then
+  BATS_REALPATH='realpath'
+fi
+set -e
+
+resolve_path() {
+  local path_to_bats=$1
+  local target_path;
+
+  if [[ -L "$path_to_bats" ]]; then
+    # Busybox has no readlink from coreutils
+    if [[ ! -z "$BATS_REALPATH" ]]; then
+      target_path="$("$BATS_REALPATH" "$path_to_bats")"
+      # we need to traverse the paths
+      target_path="$(dirname "$target_path")" # strip bats
+      target_path="$(dirname "$target_path")" # strip bin
+    else
+      target_path="$("$BATS_READLINK" -m "$path_to_bats/../../")"
+    fi
+  elif [[ -f "$path_to_bats" ]]; then
+    target_path=$(dirname "$path_to_bats");
+    if [[ ! -z "$BATS_REALPATH" ]]; then
+      target_path="$("$BATS_REALPATH" "$target_path")"
+    else
+      target_path="$("$BATS_READLINK" -m "$target_path")"
+    fi
+    target_path="$(dirname "$target_path")" # strip bin
+  elif [[ -d $path_to_bats ]]; then
+    if [[ ! -z "$BATS_REALPATH" ]]; then
+      target_path="$("$BATS_REALPATH" "$path_to_bats")"
+      target_path="$(dirname "$target_path")"
+    else
+      target_path="$("$BATS_READLINK" -m "$path_to_bats/../")"
+    fi
+  fi
+
+  echo "$target_path";
+}
+
 bats_resolve_absolute_root_dir() {
   local cwd="$PWD"
-  local path="$1"
+  local path_to_bats="$1"
   local result="$2"
-  local target_dir
-  local target_name
-  local original_shell_options="$-"
+  local target_path
 
-  # Resolve the parent directory, e.g. /bin => /usr/bin on CentOS (#113).
-  set -P
+  if [[ "$path_to_bats" = $(basename $1) ]] ; then
+    path_to_bats=$(type -p "$path_to_bats")
+  fi
 
-  while true; do
-    target_dir="${path%/*}"
-    target_name="${path##*/}"
-
-    if [[ "$target_dir" != "$path" ]]; then
-      cd "$target_dir"
-    fi
-
-    if [[ -L "$target_name" ]]; then
-      path="$("$BATS_READLINK" "$target_name")"
-    else
-      printf -v "$result" -- '%s' "${PWD%/*}"
-      set +P "-$original_shell_options"
-      cd "$cwd"
-      return
-    fi
-  done
+  target_path="$(resolve_path $path_to_bats)"
+  if [[ "$target_path" == '.' ]];
+  then
+    target_path="$(resolve_path "$target_path")"
+  fi
+  printf -v "$result" -- '%s' "$target_path"
 }
 
 export BATS_ROOT
 bats_resolve_absolute_root_dir "$0" 'BATS_ROOT'
-exec "$BATS_ROOT/libexec/bats-core/bats" "$@"
+exec env BATS_ROOT="$BATS_ROOT" "$BATS_ROOT/libexec/bats-core/bats" "$@"
--- a/bin/bats
+++ b/bin/bats
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -euo pipefail
 
 BATS_READLINK='true'
 BATS_REALPATH=''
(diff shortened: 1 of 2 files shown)


waterkip:
You can easily test the bug:
```
docker run -ti --rm --entrypoint /usr/local/bin/bash -v $(pwd):/opt/bats 835f25ff8cf8 bats /opt/bats/test
```

Yields:
```
/usr/sbin/bats: line 44: /libexec/bats-core/bats: No such file or directory
```

This fix:

```
$ docker run -ti --rm --entrypoint /usr/local/bin/bash -v $(pwd):/opt/bats 835f25ff8cf8 bats /opt/bats/test
1..80
ok 4 -h and --help print help
[snip]
```

pawamoy:
Still not merged :confused:

waterkip:
@pawamoy it will feel extra good once it is merged. Good things will happen with those who wait ;)

.gitignore
```diff
@@ -0,0 +1 @@
+docker-compose.override.yml
```
sublimino, line 1: Is this intentional? `docker-compose` isn't in this repo  : ]

Dockerfile
```diff
@@ -7,7 +7,9 @@ FROM bash:${bashver}
 RUN apk add --no-cache parallel && \
     mkdir -p ~/.parallel && touch ~/.parallel/will-cite
 
-RUN ln -s /opt/bats/bin/bats /usr/sbin/bats
-COPY . /opt/bats/
+WORKDIR /opt/bats
```
sublimino, line 10: Is this a breaking change? Users expecting an existing workdir will have to update their use of the Docker container?

Dockerfile
```diff
@@ -7,7 +7,9 @@ FROM bash:${bashver}
...
+RUN ln -s /opt/bats/bin/bats /usr/local/bin/bats
```
sublimino, line 11: Another potentially breaking change I think?

bin/bats
```diff
@@ -3,42 +3,70 @@
...
 
 BATS_READLINK='true'
+BATS_REALPATH=''
 if command -v 'greadlink' >/dev/null; then
   BATS_READLINK='greadlink'
 elif command -v 'readlink' >/dev/null; then
   BATS_READLINK='readlink'
 fi
 
+set +e
```
sublimino, line 13: Could this be an `if` instead? I *think* this is equivalent

```bash
if ! $BATS_READLINK -m $0 &>/dev/null; then
  BATS_REALPATH='realpath'
fi
```

bin/bats
```diff
@@ -3,42 +3,70 @@
...
+set -e
+
+resolve_path() {
+  local path=$1
+  local target_path;
+
+  if [[ -L $path ]]; then
+    # Busybox has no readlink from coreutisl
+    if [[ ! -z "$BATS_REALPATH" ]]; then
+      target_path="$(realpath "$path")"
```
sublimino, line 25: Should this be 

```bash
target_path="$(${BATS_REALPATH} "$path")"
```

?

bin/bats
```diff
@@ -3,42 +3,70 @@
...
+      # we need to traverse the paths
+      target_path="$(dirname "$target_path")" # strip bats
+      target_path="$(dirname "$target_path")" # strip bin
+    else
+      target_path="$("$BATS_READLINK" -m "$path/../../")"
+    fi
+  elif [[ -f $path ]]; then
+    target_path=$(dirname $path);
+    if [[ ! -z "$BATS_REALPATH" ]]; then
+      target_path="$(realpath "$target_path")"
```
sublimino, line 35: As above, should this use the variable rather than invoking `realpath` directly?

bin/bats
```diff
@@ -3,42 +3,70 @@
...
   BATS_READLINK='readlink'
 fi
 
+set +e
+$BATS_READLINK -m $0 &>/dev/null;
+[ $? -ne 0 ] && BATS_REALPATH='realpath'
+set -e
+
+resolve_path() {
+  local path=$1
```
sublimino, line 19: Although variables are case sensitive (*and* this is `local`), it might make sense to rename this for clarity as PATH is obviously used for other things. WDYT?

bin/bats
```diff
@@ -3,42 +3,70 @@
...
+      target_path="$("$BATS_READLINK" -m "$path/../")"
+    fi
+  fi
+
+  echo "$target_path";
+}
+
 bats_resolve_absolute_root_dir() {
   local cwd="$PWD"
   local path="$1"
```
sublimino, line 54: As above, if there aren't further-reaching implications for this it might be clearer with a different name?

bin/bats
```diff
@@ -3,42 +3,70 @@
...
-  local original_shell_options="$-"
-
-  # Resolve the parent directory, e.g. /bin => /usr/bin on CentOS (#113).
-  set -P
+  local target_path
 
-  while true; do
-    target_dir="${path%/*}"
-    target_name="${path##*/}"
+  if [[ $path = $(basename $1) ]] ; then
```
sublimino, line 58: Please quote all variables and use `==` for equality in `[[`

bin/bats
```diff
@@ -3,42 +3,70 @@
...
+    path=$(type -p $path)
```
sublimino, line 59: Quotes around `$path` please

bin/bats
```diff
@@ -3,42 +3,70 @@
...
+set +e
+$BATS_READLINK -m $0 &>/dev/null;
+[ $? -ne 0 ] && BATS_REALPATH='realpath'
+set -e
+
+resolve_path() {
+  local path=$1
+  local target_path;
+
+  if [[ -L $path ]]; then
```
sublimino, line 22: Please put quotes around `$path`

bin/bats
```diff
@@ -3,42 +3,70 @@
...
+    if [[ ! -z "$BATS_REALPATH" ]]; then
+      target_path="$(realpath "$path")"
+      # we need to traverse the paths
+      target_path="$(dirname "$target_path")" # strip bats
+      target_path="$(dirname "$target_path")" # strip bin
+    else
+      target_path="$("$BATS_READLINK" -m "$path/../../")"
+    fi
+  elif [[ -f $path ]]; then
+    target_path=$(dirname $path);
```
sublimino, line 33: Please put quotes around `$path`

bin/bats
```diff
@@ -3,42 +3,70 @@
...
-    if [[ -L "$target_name" ]]; then
-      path="$("$BATS_READLINK" "$target_name")"
-    else
-      printf -v "$result" -- '%s' "${PWD%/*}"
-      set +P "-$original_shell_options"
-      cd "$cwd"
-      return
-    fi
-  done
+  target_path="$(resolve_path $path)"
```
sublimino, line 62: Please put quotes around `$path`

bin/bats
```diff
@@ -3,42 +3,70 @@
...
+  if [[ $target_path = '.' ]];
```
sublimino, line 63: Quotes around `$target_path` please

bin/bats
```diff
@@ -3,42 +3,70 @@
...
+  then
+    target_path="$(resolve_path $target_path)"
```
sublimino, line 65: Quotes around `$target_path` please

docker-compose.override.dist
```diff
@@ -0,0 +1,8 @@
+# Copy this file to docker-compose.override.yml
```
sublimino, line 1: Could we move the `docker-compose` work out into another PR please? It needs some tests around it too please : ]

test/root.bats
```diff
@@ -40,28 +40,45 @@ setup() {
...
+  cd "$BATS_TEST_SUITE_TMPDIR"
+  run ./bin/bats -v
+  [ "$status" -eq 0 ]
+  [ "${output%% *}" == 'Bats' ]
+}
+
+@test "set BATS_ROOT from PATH" {
+  cd "$BATS_TEST_SUITE_TMPDIR"
+  cd /tmp
+  PATH=$PATH:$BATS_TEST_SUITE_TMPDIR/bin
```
sublimino, line 80: Quotes around the variable assignment please

bin/bats
waterkip: I don't think so, but I'll have a look at the logic when I rework the MR.

Dockerfile
waterkip: No, because it is in PATH it isn't a breaking change.

Dockerfile
waterkip: The default workdir for docker is `/`. It could be breaking if people copy do `COPY . to foo` and expect it to be in `/`. Although I think most people either set the workdir before a copy and than use relative paths or they copy it with the absolute path.

bin/bats
waterkip: What would be a better naming, it is the path we are talking about. I'm open for suggestions but I personally think, `local` and the complete lack of case-y-ness should make it clear that is isn't the PATH we are talking about especially in the function called `resolve_path`.

.gitignore
waterkip: I removed the docker bits from this PR and will move it to a different one.

waterkip:
> Sorry, I'm so busy it's difficult to find time, but better late than never I hope 
> 
> Some nits here, mostly quoting, and some possible breaking changes in the Dockerfile I'd prefer to avoid if we can, or add something to the release notes otherwise.
> 
> I'm still intending to release the next version before merging changes, but this looks good to go when that's done, thanks for the hard work on this!

I moved the docker bits to a different PR and updated the PR to reflect your comments. Some bits could be left out I found out during an autosquash. If you could have another look that would be awesome :)

PR #191, opened 2019-02-26
added support for bash compliant syntax allowing for shellcheck parsing

lucastheisen:


sublimino:
This is an interesting PR, thanks!

If we are going to deviate from bats' original syntax, I think a wider discussion is needed. For example many languages parse `test_X()` as a test case, should we adopt that? Is there an alternative syntax altogether?

It seems whatever we do may be a breaking change and will require some thought. Could you open an issue and add some detail around your thinking for this change please? We may want to provide compliance with other tools, such as shfmt, too - this will give people a chance to discuss the feature and any other requirements. Thanks 👍

lucastheisen:
@sublimino , I created [an issue to discuss the problem](https://github.com/bats-core/bats-core/issues/192) per your request.  That said, this _specific_ pull request was designed to avoid

> a breaking change

It is a simple, additional, syntax that required very few modifications to the existing source.  Other options could be:

*prefix test_ or suffix _test:*
```
function test_should_do_what_i_think { }
function should_do_what_i_think_test { }
```

*directive*
```
# bats @test
function should_do_what_i_think { }

# bats @test="optional description instead of method name"
function should_do_what_i_think { }
```

Possibly even more could be allowed without breaking backwards compatibility.

sublimino:
I agree it's quite possibly not breaking at all, just wanted to open the conversation in case smarter people than me can see other cases (compared to the existing bizarre but well-used syntax). Thanks for this PR, and for opening the issue 🙏

libexec/bats-core/bats-preprocess
```diff
@@ -35,8 +35,8 @@ test_file="$1"
 tests=()
 {
   while IFS= read -r line; do
-    line="${line//$'\r'/}"
-    if [[ "$line" =~ $BATS_TEST_PATTERN ]]; then
+    line="${line//$'\r'}"
+    if [[ "$line" =~ $BATS_TEST_PATTERN ]] || [[ "$line" =~ $BATS_TEST_PATTERN_COMMENT ]]; then
```
sublimino, line 39: I wonder what the performance penalty is here (scratching up some perf regression tests I'll try this on).

If it's tangible, and there's a way of merging these regexs, that may help.

test/fixtures/bats/comment_style.bats
```diff
@@ -0,0 +1,20 @@
+function should_be_found { # @test
+  true
+}
+
+function should_be_found_with_trailing_whitespace { # @test   
+  true
+}
+
+function should_be_found_with_no_whitespace{#@test
```
sublimino, line 9: `shfmt` rejects this syntax FWIW, and it doesn't appear to be valid without spaces before and after the `{`. Should we reject this I wonder?

```bash
$ function should_be_found_with_no_whitespace{ echo 123; }
bash: syntax error near unexpected token `echo'
```

test/fixtures/bats/comment_style.bats
```diff
@@ -0,0 +1,20 @@
...
+}
+
+should_be_found_with_parens() { #@test
+  true
+}
+
+should_be_found_with_parens_and_whitespace () { #@test
+  true
+}
+
```
sublimino, line 20: Some additional negative tests could help, perhaps something broadly like:

```bash
should_not_be_found() { 
  true                                          
  #@test                                        
}                                               

should_not_be_found() {  
  true                   
} #@test                 

...
```

libexec/bats-core/bats
```diff
@@ -93,6 +93,7 @@ expand_path() {
 BATS_LIBEXEC="$(dirname "$(expand_link "${BASH_SOURCE[0]}")")"
 export BATS_CWD="$PWD"
 export BATS_TEST_PATTERN="^[[:blank:]]*@test[[:blank:]]+(.*[^[:blank:]])[[:blank:]]+\{(.*)\$"
+export BATS_TEST_PATTERN_COMMENT="[[:blank:]]*([^[:blank:]()]+)[[:blank:]]*\(?\)?[[:blank:]]*\{[[:blank:]]*#[[:blank:]]*@test[[:blank:]]*\$"
```
martin-schulze-vireso, line 96: Just a little nitpick, following invalid patterns are possible with this pattern:

- `should_be_found) {# @test`
- `should_be_found( {# @test`

martin-schulze-vireso: Additionally, could users expect to be able to write `function function_name() { #test`?

test/fixtures/bats/comment_style.bats
lucastheisen: Given that bash fails on this, i see 2 paths.  We fail during pre-processing, or simply let it fail when run.  I would suggest the later and simply remove this test.

test/fixtures/bats/comment_style.bats
lucastheisen: Good suggestion, [added](https://github.com/bats-core/bats-core/pull/191/commits/92f3c92f8f067ba4a99b45b8060481035fe40997)

libexec/bats-core/bats
lucastheisen: > following invalid patterns are possible with this pattern

I agree they would be allowed with this pattern, but the point of this pattern is that _valid bash works_, and neither of the patterns listed are valid bash.  We could make the regex require `[:blank:]` before and after the `{` which would cause these 2 examples to not match, which in turn would cause an error when the test file was run:

```text
 ✗ test comment style
   (in test file test/bats.bats, line 619)
     `[ $status -eq 0 ]' failed
   [1] <<<1..6 /tmp/bats-run-9663/bats.9682.src: line 25: syntax error near unexpected token `)' /mnt/c/Users/ltheisen/git/pastdev-bats-core/lib/bats-core/tracing.bash: line 61: BATS_STACK_TRACE: bad array subscript not ok 1 setup_file failed # bats warning: Executed 1 instead of expected 6 tests>>>
```

But I don't think that's necessary.  I don't know of a way to prevent `(` or `)` without a non-capturing group because `BASH_REMATCH[2]` is used to match the body so if i try `(\(\))?` to enforce `()` or nothing, then it is used as the body of the function and fails.  Thus, it would be a much more extensive modification.  In my opinion, we should probably leave the regex as is and leave it up to the users to ensure their code is valid bash.

> Additionally, could users expect to be able to write function function_name() { #@test? It might be very confusing if these are not translated into tests.

I just [added some tests](https://github.com/bats-core/bats-core/pull/191/commits/ddd169d296fc0efa080968de9565f3ad6c225fab) that demonstrates that `function name()` style functions will indeed be identified as tests.  While not explicitly supported (according to bash `help function`) it does seem to be allowed bash syntax, so seems like they should be found as tests.

libexec/bats-core/bats-preprocess
lucastheisen: I was unable to come up with a way to merge them (mighta been able to with PCRE, but posix regex seems to have too many limitations).  That is not to say someone smarter couldn't :).  That said, this check seems like it would be far less costly than running the tests themselves, so likely not a huge problem in any case.

libexec/bats-core/bats
martin-schulze-vireso: > the point of this pattern is that valid bash works

You're right, I was scanning the regex for completeness and made a list that contained more items but after realising that the others were wrong I posted just those two, hence the "nitpick" part.

> I just added some tests that demonstrates that function name() style functions will indeed be identified as tests.

Oh, I just noticed you already had tests for `function name {` which I actually was getting at, it was not so much about the parens but them working too won't hurt.

libexec/bats-core/bats-preprocess
martin-schulze-vireso: The two regexes could be merged via `MERGED_REGEX="($REGEX1)|($REGEX2)"`. However, that would lead to different capture groups beeing used for the name. The two regex parts would become capture groups so it would be easy to identify which one matched and go on from there. However, I also doubt that this would pose a significant performance problem right now, since the overhead per test is already at 100ms.

martin-schulze-vireso:
Okay, I tested this with a file of 1 MLOC and the performance difference between master and this was negligible. I rebased and signed this locally but could not push into this PR. @lucastheisen do you mind allowing changes by the owner? Else you would need to do the rebase and sign the commits yourself or I would have to open another PR with the rebased code.

lucastheisen:
@martin-schulze-vireso , I am not sure how to

> ... allowing changes by the owner?

How would I do that?

martin-schulze-vireso:
Well, that is strange. In this case I will open another PR with the rebased code.

martin-schulze-vireso:
I am closing this now. @lucastheisen Please review #378 to see if I messed something up.

PR #196, opened 2019-03-05
bats: add support for deteriministic --shuffle

cyphar:
There are many cases where tests accidentially depend on one another,
and bats users might want to ensure their test suite does not have this
behaviour. While --jobs gives you some randomness in the execution,
--shuffle allows for deterministic ordering in order to re-test a
particular case (by setting BATS_SHUFFLE_SEED) and also shuffles all
tests entirely. And it only requires GNU coreutils (awk and shuf).

This is a follow-up of #172.


libexec/bats-core/bats-exec-suite
```diff
@@ -1,7 +1,25 @@
...
+}
+
+random_number() {
+  random_stream "$RANDOM$RANDOM" | tr -cd '0-9' | head -c 16
+}
+
+random_shuffle() {
+  shuf="$(type -p gshuf shuf | head -1)"
+  "$shuf" --random-source <(random_stream "$BATS_SHUFFLE_SEED")
+}
```
cyphar: To explain what is going on here, this is basically a way of creating a random stream with a seed value. As far as I'm aware this is the nicest way of doing it such that you can pretend `random_stream` is like `/dev/urandom`.

cyphar:
This still needs some tests -- though I'm a little cautious about how we're going to test the seed values in a way that doesn't cause us problems in the future. I will work on that when I have some time.

cyphar:
As an update, it looks like old bash versions don't like the `arrays_equal` code. I will need to take a look at this...

jasonkarns:
I'm 👍 on the feature. Haven't looked at the implementation, though. But before we settle too hard on the interface (the flag and var names), can we put together a quick survey of what other tools are using for their CLI flags/options/var names? That way we can follow any existing norms, however, slight they may be?

cyphar:
@jasonkarns As far as I'm aware there aren't many other testing tools which provide something like this. @sublimino [mentioned this](https://github.com/bats-core/bats-core/issues/171#issuecomment-462130265) in #171. Ruby has `--seed` and random ordering is the default, so I could add another command-line flag if you like (but the default ordering is going to be in-order to avoid breaking people's tests).

jasonkarns:
@cyphar Virtually every testing tool that I'm aware of, which supports random execution, also supports re-running a prior random order deterministically.

RSpec:

```
--order defined
--order rand
--order rand:123
--seed 123 # same as --order rand:123
```

Minitest is random by default and accepts seed via `-s`, `--seed` or `SEED` env var.
(with an in-test helper to force non-random order: `i_suck_and_my_tests_are_order_dependent!`)

Cucumber uses `--order random` and `--order random:SEED`

In JavaScript/node land:

Mochajs doesn't officially support random order, but the "heavily used" patch uses `--random SEED`.

Jasmine has `--random=true|false` and `--seed=SEED`

Tap, Jest, Ava, Cypress don't seem to support true randomization nor determinism with a seed; though they support parallelization of the suite so there's some implicit randomicity but without the ability to re-run that particular order. (And they can both force serial runs, which of course isn't the same as deterministically re-running a random order.)

The rest of the common JS tools (tape, teenytest, nightwatch, qunit) don't seem to support randomization at all.

I would be curious to see what features the top 2 or 3 testing tools of python, php, perl, go, java, .net support. At the very least `--seed` so far seems to be a de facto standard for the seed.

jasonkarns:
Personally, I do like cucumber, rspec, and mochajs' dual use of a single flag for both enabling randomization and providing the seed (since a seed without random ordering is meaningless).

`--order rand[:seed]` is a bit heavy on the parsing. `--random [SEED]` is a simple compromise (the seed argument is optional). 

`--seed SEED` on its own implying randomization seems to be the most straightforward of all the options. It follows the "convention" that all of these tools use for `--seed`, and since --seed without randomization is meaningless, it makes sense to me that the presence of the flag also triggers the shuffling. The downside, is that we would still need a separate flag to trigger shuffling without a seed.

cyphar:
I will go with `--random [SEED]` then.

test/test_helper.bash
```diff
@@ -14,6 +14,40 @@ make_bats_test_suite_tmpdir() {
   mkdir -p "$BATS_TEST_SUITE_TMPDIR"
 }
 
+arrays_equal() {
```
henning-schild, line 17: might be worth putting into a seperate commit, maybe with dedicated tests

test/bats.bats
```diff
@@ -483,3 +483,77 @@ END_OF_ERR_MSG
   # In theory it should take 3s, but let's give it bit of extra time instead.
   [[ "$duration" -lt 20 ]]
 }
+
+@test "randomise test execution order with --shuffle" {
+  # We can't be sure what the order will be -- but it should be different
+  # between different runs. We ignore all #-lines because they always include
+  # BATS_SHUFFLE_SEED which should always be different anyway.
+
```
henning-schild, line 491: Not sure how it is made sure that the several runs will not accidentally produce the same order. I guess that even different seeds could result in the same order.

If there is a slight chance of a collision (same seed or same order with different seed) i suggest calling with fixed seeds that are known to be free of collisions.

henning-schild (commented):
Cool feature!

Issue #202, opened 2019-03-26
ensure unofficial bash script mode test fails when run out of the tree


ycdtosa:
I am trying to pass the source tree tests out of place, running an installed bats binary.
This is the failing test from the bats.bats file.
```
@test 'ensure compatibility with unofficial Bash strict mode' {
  local expected='ok 1 unofficial Bash strict mode conditions met'

  # Run Bats under `set -u` to catch as many unset variable accesses as
  # possible.
  run bash -u "${BATS_TEST_DIRNAME%/*}/bin/bats" \
    "$FIXTURE_ROOT/unofficial_bash_strict_mode.bats"
  if [[ "$status" -ne 0 || "${lines[1]}" != "$expected" ]]; then
    cat <<END_OF_ERR_MSG
          ...blah blah blah
END_OF_ERR_MSG
    emit_debug_output && return 1
  fi
}
```

However it is failing because there is no "${BATS_TEST_DIRNAME%/*}/bin/bats" since i copied the tests out of the tree to a different location.

I want to test that the installed (/usr/bin/bats) bats is working!

The test tries to run bats with
```
run -u "${BATS_TEST_DIRNAME%/*}/bin/bats" \
   "$FIXTURE/unofficial_bash_script_mode.bats"
```

Is there any reason why i can not just use bats like most other tests at bats.bats
without the relative path trough the tests dirname? 

```
run -u bats \
   "$FIXTURE/unofficial_bash_script_mode.bats"

```

I have changed it and the test passes in both locations, in the source tree and in the isolated tree.

ycdtosa:
to fully reproduce the issue...

```
cp -r bats-core/test ~/test
# bats has been installed, and bats is in the path.
# I want to test if the bats binary in the path works.
bats ~/test
```

the test will fail, because there is no `~/test/../bin/bats` to run

ycdtosa:
I would say this is related

> To close the loop on this: I found and explained the breakage in #93, which actually wasn't due to #92. The "unofficial Bash strict mode" test failed because it was looking for `bats` relative to `BATS_TEST_DIRNAME`, which it couldn't because the `test` directory was mounted into the container in isolation. Mounting the root of the Bats project resolved the issue.

_Originally posted by @mbland in https://github.com/bats-core/bats-core/pull/91_

martin-schulze-vireso:
@ycdtosa  Can you verify that this is fixed now?

PR #203, opened 2019-03-26
fix bats.bats test about strict mode

ycdtosa:
Run bats, instead of hardcoding a relative path to a binary which might not be there.

This helps to run the tests out of source and to test an installed binary.

This refs #202

### DIFF
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -322,12 +322,12 @@ fixtures bats
   [ "${lines[3]}" = "ok 3 unquoted name" ]
 }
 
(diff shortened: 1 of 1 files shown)


ycdtosa:
wow. how broken is that!... sorry. will try again :-)

ycdtosa:
Ok. Still i am not convinced this is the proper way to call the same bats binary that started the test.
Is it?

Please review.

martin-schulze-vireso:
Thanks for your contribution. The same fix has already been applied in another PR.

PR #208, opened 2019-03-30
before_run/after_run hooks

joshrabinowitz:
Adds before_run/after_run hooks.

- [ x ] I have reviewed the [Contributor Guidelines][contributor].
- [ x ] I have reviewed the [Code of Conduct][coc] and agree to abide by it

### DIFF
--- a/libexec/bats-core/bats-exec-test
+++ b/libexec/bats-core/bats-exec-test
@@ -65,11 +65,16 @@ run() {
   local origFlags="$-"
   set +eET
   local origIFS="$IFS"
+
+  before_run
+
   output="$("$@" 2>&1)"
   status="$?"
   IFS=$'\n' lines=($output)
   IFS="$origIFS"
   set "-$origFlags"
+
+  after_run
 }
 
 setup() {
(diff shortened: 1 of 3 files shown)


joshrabinowitz:
I'm not sure what's wrong, in my self-tests the `after_run` hook I added doesn't get called. 
See here  => https://github.com/bats-core/bats-core/pull/208/files#diff-a962dbbde32925cd2df8fb361b7583a5R3

martin-schulze-vireso:
Well you injected your run hooks into the run command but you never use it.
Adding a line like `run true` to your tests triggers the output in the after_run.log.

I think you might have confused run which runs a single command inside a test with the running of a `@test` function itself.
The actual execution of the test is triggered [here](https://github.com/bats-core/bats-core/blob/1c83a1b1d743075ed8e505ff94e548701f545b73/libexec/bats-core/bats-exec-test#L337) so you would have to put your hooks around that and also ensure after_run is called in the error/teardown trap, like the teardown function.

joshrabinowitz:
@martin-schulze-vireso thank you for the response.

Closing this PR, I'll re-open it if I continue development on it.

PR #214, opened 2019-04-17
readme update to explain coreutils requirement on Darwin

NorseGaud:
- [*] I have reviewed the [Contributor Guidelines][contributor].
- [*] I have reviewed the [Code of Conduct][coc] and agree to abide by it


cyphar:
Yes, the code for it is in #196. The main blocker at the moment is that older bash versions are crashing when running the new tests for `--shuffle` (I think it's because of the `declare -n` in `arrays_equal`). I could split out just the `paste` code if you prefer.

PR #215, opened 2019-04-18
Always use upper case signal names to avoid locale dependent errors

roktas:
Lower case signal names cause errors under Turkish locale (see [Turkish dotless i](https://en.wikipedia.org/wiki/Dotted_and_dotless_I) for the reasons).  Reproduce the issue:

```sh
$ locale-gen tr_TR.UTF-8 # Ubuntu derivatives
$ LC_CTYPE=tr_TR.UTF-8 LC_MESSAGES=C bash -c 'trap - int'
bash: line 0: trap: int: invalid signal specification
```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -336,6 +336,21 @@ The following are intended to prevent too-compact code:
 
 [printf-vs-echo]: https://unix.stackexchange.com/a/65819
 
+### Signal names
+
+Always use upper case signal names (e.g. `trap - INT EXIT`) to avoid locale 
+dependent errors. In some locales (for example Turkish, see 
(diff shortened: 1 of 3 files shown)


sublimino:
Thanks @roktas! I've added this to the contributor documentation too.

It would be great to have a test to prevent regressions in https://github.com/bats-core/bats-core/blob/master/.travis.yml if you have time ,but I'll merge this as-is because it's been open so long :+1:

dimo414:
Do we actually know that upper-case signal names don't also exhibit this same issue in certain locales? If this behavior can happen for lower-case letters, why wouldn't it similarly be possible with upper-case letters?

I don't fully understand what's happening in this specific case, but it seems like lower-case ASCII `i` (U+0069) is being replaced with a different codepoint? Are we sure this isn't a bug in Bash or another layer of the stack?

Issue #197, opened 2019-03-09
bats no longer works with programs that read stdin


natefoo:
As of 17e2cf35ba4eb44a25f47f31f079809c0f111338 from #172 you can no longer use `run` in conjunction with programs that read from stdin, because the program will consume the remaining test files/names waiting on stdin [here](https://github.com/bats-core/bats-core/blob/8789f910812afbf6b87dd371ee5ae30592f1423f/libexec/bats-core/bats-exec-suite#L99).

A workaround is to redirect `/dev/null` to stdin, e.g. `run foo </dev/null`.

ycdtosa:
can you please provide an example?

natefoo:
@ycdtosa sure:

`read.sh`:
```sh
#!/bin/sh

[ ! -t 0 ] || return 0
[ -n "$DO_READ" ] || return 0

while read foo; do
    echo "$foo" >&3
done

return 0
```

`test.bats`:
```sh
#!/usr/bin/env bats

@test "test 1" {
    run ./read.sh
}

@test "test 2" {
    run ./read.sh
}

@test "test 3" {
    run ./read.sh
}
```

example run:
```console
% ./test.bats
 ✓ test 1
 ✓ test 2
 ✓ test 3

3 tests, 0 failures

% DO_READ=1 ./test.bats
 ✓ test 1

3 tests, 0 failures

% DO_READ=1 ./test.bats -t
1..3
/home/nate/test/bats/test.bats	test_test_2
/home/nate/test/bats/test.bats	test_test_3
ok 1 test 1
```

And closing stdin (updated `test.bats`, same `read.sh`):
```sh
#!/usr/bin/env bats

@test "test 1" {
    run ./read.sh </dev/null
}

@test "test 2" {
    run ./read.sh </dev/null
}

@test "test 3" {
    run ./read.sh </dev/null
}
```

example run:

```console
% DO_READ=1 ./test.bats
 ✓ test 1
 ✓ test 2
 ✓ test 3

3 tests, 0 failures

```

cyphar:
Damn. A simple solution might be to use a different fd for the `read`.

The primary issue is that bash doesn't allow you to use pipelines while also changing bash variables (because pipelines all execute in subshells) -- and iterating over an array would be slightly nasty to get right. So we end up with the `while foo; do ; done < <()` trick.

But I'll take a look at this.

Flamefire:
[...]

PR #227, opened 2019-06-12
Fix for tests reading from stdin

Flamefire:
As discussed in #197 the PR #172 introduced a change which broke the test runner when the test consumed stdin (containing remaining tests) by

- Iterating over the `all_tests` array previously created and split at the **last** TAB found. By construction of the test lines this should work in all cases (test function names cannot contain a TAB)
- count number of executed tests and fail if it doesn't match number of tests that were scheduled
  - avoid silent failure as on current master
- show number of not-run tests in the pretty-printer summary by counting passed tests


Flamefire:
Alternative would be e.g. redirecting /dev/null do stdin for the test commands, see e.g. https://stackoverflow.com/questions/13800225/shell-script-while-read-line-loop-stops-after-the-first-line

But I think this solution is better as it only relies on "simply" bashisms like iterating over a previously filled array. It also avoids an issue if the path to the test file contains a TAB

sublimino:
Thanks @Flamefire, this LGTM. Agreed on "simply Bashisms".

Anybody else from @bats-core/bats-core or @cyphar want to have a look?

cyphar:
LGTM from me. Thanks @Flamefire!

test/fixtures/suite/read_from_stdin/test.bats
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bats
+
+function cmd_using_stdin {
+    # Just reading from stdin
+    while read -r -t 1 foo; do
```
sublimino: Each test here blocks for a second - could this be reduced to `read -r -t 0.001 foo`?

test/fixtures/suite/read_from_stdin/test.bats
```diff
@@ -0,0 +1,30 @@
...
+        if [ "$foo" == "EXIT" ]; then
```
sublimino: This line is never reached. Maybe `test 3` could be something like
```bash
export -f cmd_using_stdin
run bash -c "echo 'EXIT' | cmd_using_stdin"
[ "$status" -eq 1 ]
```

Otherwise perhaps remove this block?

test/fixtures/suite/read_from_stdin/test.bats
Flamefire: I tried but a float as timeout is not supported on all Bash versions. The older ones only accept integers

test/fixtures/suite/read_from_stdin/test.bats
Flamefire: Good point: A test where something is actually read from stdin was missing.

- Extracted command into own file to use with `bash -c`
- Change tests to use stdin and not use stdin
- Move from fixtures/suite to fixtures/bats where it is better suited
- Rebased on master

test/fixtures/suite/read_from_stdin/test.bats
sublimino: Argh, I wondered if that was the case. From my rudimentary testing looks like Bash 3 is the problem? As this adds three seconds I'm wondering if there's any way to avoid adding that time to every test run. We could set by `${BASH_VERSINFO}` perhaps?

Flamefire: Good idea: bcce611

sublimino:
Awesome! Fast and clean. Cheers @Flamefire : ] I'll keep on with your changes in other PRs, sorry it's a bit slow -- we'll get there :pray:

PR #229, opened 2019-06-14
Fix wrong line numbers of errors in bash < 4.4

### LINKED ISSUE #228: Wrong failing line shown on bash < 4.4
TLDR: Doing a `return 1` in a function messes up the stacktrace of bash and hence bats which will show a wrong failing line.

Example:

`assert.bash`:   
```
function fail {
  return 1
}
```
`test.bats`:
```
load assert
@test "test" {
  echo "Prev"
  # Increase linecount
  # Now fail
  fail
}
```

I traced this down to the point where bash deletes a stack entry when returning from `fail`. It (correctly) deletes the top-most entry from `$BASH_SOURCE` but the 2nd entry from `$BASH_LINENO`. This looks like this (printing the stack trace array from the debug trap):

```
[D] 18 fail assert.bash 64 test_test /tmp/bats.10668.src <- Before return: Correct
[D]18 test_test /tmp/bats.10668.src <- After return: Wrong
```
This will show a failure in the test file line 18, instead of line 64 (or line 18 in assert.bash).

I created test cases to reproduce this but was unable so far to create a fix, especially as this is fixed in bash 4.4. But this is quite surprising and very annoying to handle.

Flamefire:
The last entry in the stack trace is unreliable, so don't use the last
stack trace at all similar to the LINENO used before.

Fixes #228

### DIFF
--- a/libexec/bats-core/bats-exec-test
+++ b/libexec/bats-core/bats-exec-test
@@ -214,8 +214,9 @@ bats_debug_trap() {
   if [[ "$BASH_SOURCE" != "$1" ]]; then
     # The last entry in the stack trace is not useful when en error occured:
     # It is either duplicated (kinda correct) or has wrong line number (Bash < 4.4)
-    # Therefore we use capture the stacktrace but use it only after the next debug
-    # trap fired. Expansion is required for empty arrays otherwise erroring
+    # Therefore we capture the stacktrace but use it only after the next debug
+    # trap fired.
+    # Expansion is required for empty arrays which otherwise error
     BATS_CURRENT_STACK_TRACE=( "${BATS_STACK_TRACE[@]+"${BATS_STACK_TRACE[@]}"}" )
     bats_capture_stack_trace
   fi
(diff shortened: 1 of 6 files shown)


libexec/bats-core/bats-exec-test
```diff
@@ -317,18 +319,17 @@ bats_perform_test() {
...
-  BATS_LINENO=0
   BATS_STACK_TRACE=()
+  BATS_CURRENT_STACK_TRACE=()
 
   BATS_TEST_COMPLETED=
   BATS_TEST_SKIPPED=
   BATS_TEARDOWN_COMPLETED=
   BATS_ERROR_STATUS=
-  trap "bats_debug_trap \"\$BASH_SOURCE\"" debug
+  trap 'bats_debug_trap "$BASH_SOURCE"' debug
```
josegonzalez: This change results in the value `BASH_SOURCE` being printed as is.

Flamefire: No, it's the same as before just less escapes and hence less confusing

josegonzalez: Oh you are right and I am an idiot :D

sublimino:
This seems fairly reasonable, we'll get it merged next :+1:

sublimino:
@Flamefire I've made some minor changes during review and PR'd them to your branch instead of reviews here.

https://github.com/Flamefire/bats-core/pull/1

If you're happy with them please merge on your branch, and I'll merge here. Otherwise I'm happy to open them for discussion here instead 🙏

Flamefire:
LGTM merged. Thanks!

sublimino:
Many thanks once more @Flamefire! :+1:

PR #238, opened 2019-07-24
Added Shellcheck script and CI config. Closes #14.

### LINKED ISSUE #14: Implement shellcheck in CI
We should integrate https://github.com/koalaman/shellcheck in our CI process.

In the meanwhile, we can encourage the use of tools like https://github.com/vim-syntastic/syntastic, which has saved my life as a Vim user.

Fun fact: I haven't yet implemented shellcheck for mbland/go-script-bash yet because, last time I checked, it didn't recognize `.` as a valid function name character. (I should check again, and submit a PR if that still isn't the case.)

alexdesignworks:

This PR adds Shellcheck support to CI through a `shellcheck.sh` script that may also be ran locally with `./shellcheck.sh`.

The check is ran in CI for Linux only - this is faster then installing Shellcheck for different OSes and is sufficient enough (Shellcheck will produce the same results independently of the environment it ran).

Please let me know about this approach and I can adjust the code. We really need coding standards check in PRs :))

Also, please note that the check is always passing in CI - this is to resolve in another PR that would be solely for fixing coding standards.

## Screenshot
This is a screenshot of CI job from this PR that passes the check, but some errors are still need to be fixed (or excluded).

Job__588_1_-_bats-core_bats-core_-_Travis_CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,9 @@ matrix:
 services:
   - docker
 
+before_script:
+ - sudo apt-get install shellcheck
+
 script:
 - |
   if [[ "${TRAVIS_OS_NAME:-}" == 'linux' && -n "${BASHVER}" ]]; then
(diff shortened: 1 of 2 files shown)


sublimino:
This seems well thought out : ] Thanks @alexdesignworks.

I'll merge on the weekend unless anybody thinks otherwise /cc @jasonkarns @bats-core/bats-core

sublimino:
@alexdesignworks thanks for the contribution, all working as expected : ]

https://travis-ci.org/bats-core/bats-core/builds/564380711

PR #239, opened 2019-08-04
Need ncurses for pretty printing to work

dlorch:

As the title says. Without ncurses, the output defaults to "TAP", even on an interactive console:

```
$ docker run -it bats/bats:latest /opt/bats/test
1..78
ok 4 -h and --help print help
ok 5 invalid filename prints an error
ok 6 empty test file runs zero tests
ok 7 one passing test
ok 8 summary passing tests
...
```

with ncurses:

```
$ docker run -it bats/bats:latest /opt/bats/test
 ✓ no arguments prints message and usage instructions
 ✓ invalid option prints message and usage instructions
 ✓ -v and --version print version number
 ✓ -h and --help print help
 ✓ invalid filename prints an error
 ✓ empty test file runs zero tests
 ✓ one passing test
 ✓ summary passing tests
...
```

Note that -p and --pretty do not work without ncurses.

I don't know if ncurses is actually needed or not, but bats checks for it [here]:

`command -v tput`

[here]: https://github.com/dlorch/bats-core/blob/master/libexec/bats-core/bats#L94

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM bash:${bashver}
 
 # Install parallel and accept the citation notice (we aren't using this in a
 # context where it make sense to cite GNU Parallel).
-RUN apk add --no-cache parallel && \
+RUN apk add --no-cache parallel ncurses && \
     mkdir -p ~/.parallel && touch ~/.parallel/will-cite
 
 RUN ln -s /opt/bats/bin/bats /usr/sbin/bats


Dockerfile
```diff
@@ -6,6 +6,7 @@ FROM bash:${bashver}
 # context where it make sense to cite GNU Parallel).
 RUN apk add --no-cache parallel && \
```
sublimino: Please add `ncurses` after `parallel` so it's all in the same Docker image layer and remove the separate line 🙏

dlorch: Fixed in last commit.

sublimino:
Thanks for the contribution @dlorch! :+1:

sublimino:
It occurs to me now this is merged that this is redefining accidental behaviour that CI suites may have been relying on: implicit default in Docker was `--tap`, it's not `--pretty`

If this will cause anybody problems please LMK here.

PR #243, opened 2019-08-09
Add sample .bats files into the examples/ dir

amilkh:
This will help beginners integrate bats into their workflow more easily.

Authored-by: Amil Khanzada <akhanzada@pivotal.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,8 +158,8 @@ Check out a copy of the Bats repository, then build a container image:
     $ docker build --tag bats/bats:latest .
 
 This creates a local Docker image called `bats/bats:latest` based on [Alpine
-Linux](https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md) 
-(to push to private registries, tag it with another organisation, e.g. 
+Linux](https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md)
+(to push to private registries, tag it with another organisation, e.g.
 `my-org/bats:latest`).
 
 To run Bats' internal test suite (which is in the container image at
@@ -172,8 +172,8 @@ to its path inside the container:
 
     $ docker run -it -v "$(pwd):/code" bats/bats:latest /code/test
 
-This is a minimal Docker image. If more tools are required this can be used as a 
-base image in a Dockerfile using `FROM <Docker image>`.  In the future there may 
+This is a minimal Docker image. If more tools are required this can be used as a
+base image in a Dockerfile using `FROM <Docker image>`.  In the future there may
 be images based on Debian, and/or with more tools installed (`curl` and `openssl`,
 for example). If you require a specific configuration please search and +1 an
 issue or [raise a new issue](https://github.com/bats-core/bats-core/issues).
@@ -263,6 +263,8 @@ process.
 
 For more details about how Bats evaluates test files, see [Bats Evaluation
 Process][bats-eval] on the wiki.
+
+For sample test files, see [examples](/docs/examples).
 
 [bats-eval]: https://github.com/bats-core/bats-core/wiki/Bats-Evaluation-Process
 
--- a/docs/examples/README.md
+++ b/docs/examples/README.md
@@ -1,0 +1,6 @@
+# Examples
+
+This directory contains example .bats files.
+See the [bats-core wiki][examples] for more details.
+
+[examples]: (/bats-core/bats-core/wiki/Examples)
--- a/docs/examples/package-tarball
+++ b/docs/examples/package-tarball
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+# "unofficial" bash strict mode
+# See: http://redsymbol.net/articles/unofficial-bash-strict-mode
+set -o errexit  # Exit when simple command fails               'set -e'
+set -o errtrace # Exit on error inside any functions or subshells.
+set -o nounset  # Trigger error when expanding unset variables 'set -u'
+set -o pipefail # Do not hide errors within pipes              'set -o pipefail'
+set -o xtrace   # Display expanded command and arguments       'set -x'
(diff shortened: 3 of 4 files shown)


examples/if_ls_glob.bats
```diff
@@ -0,0 +1,49 @@
...
+
+@test "it works when there are two .bats file" {
+  touch {a,b}.bats
+
+  run main
+  [ "${status}" -eq 0 ]
+}
+
+@test "it works when there are three .bats file" {
+  touch {a,b,c}.bats
```
alexdesignworks: is it possible to not use `.bats` extension as for the beginners it is not clear why `.bats` was used specifically. Using `.txt` would be more straightforward.

examples/if_ls_glob.bats
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bats
+
```
alexdesignworks: could you please add a description that this is an example file

examples/if_ls_glob.bats
```diff
@@ -0,0 +1,49 @@
...
+TEST_DIR="test_dir"
```
alexdesignworks: it is not clear what this variable for.

here and below - can you please add comments for every part that is not documented, so that newcomers could read the file without guessing what every line means, thanks

examples/if_ls_glob.bats
```diff
@@ -0,0 +1,49 @@
...
+}
+
+teardown() {
+  cd ..
+  rm -rf "${TEST_DIR}"
+}
+
+main() {
+  if ls *.bats >/dev/null 2>&1; then
+    bats -t *.bats
```
alexdesignworks: could you please explain why does the test function need to run tests. this is inception. very confusing.

can this simply create a set of fixture files?

or maybe I'm missing the point about this test example (hence the file comment would help to set the context)

amilkh:
Thanks for the feedback @alexdesignworks ! I'll address your comments on Friday.

examples/if_ls_glob.bats
amilkh: thanks commend added!

examples/if_ls_glob.bats
amilkh: hmm... I think the file should be self-documenting after the latest refactors. I try to avoid comments when the code can self-document.

examples/if_ls_glob.bats
amilkh: have removed the `ls` script for now because it is confusing!

amilkh:
Have made some updates, please let me know your thoughts!

alexskrypnyk:
@amilkh re p.3:
1. Latest version of shellcheck supports `.bats` linting
2. There is a PR for the CI of this project to check all files.

amilkh:
> 1. Latest version of shellcheck supports `.bats` linting
> 2. There is a PR for the CI of this project to check all files.

Thanks! I ran shellcheck.
also added note to README.md


if this PR looks ok, can we get it merged?

virgilwashere:
> Questions:
> 
> 1. From the code guidelines it says:
>> Declare temporary file-level variables using declare. Use unset to remove them when finished.
> should this apply to `$TEST_DIR`?

Yes it should. However ...(see next)

> 2. I think there may be a better pattern than `$TEST_DIR`.

Use `$BATS_TMPDIR` as defined in [Special Variables](/bats-core/bats-core/blob/master/README.md#special-variables)

- [x] pre-declared for you

- [x] existing standard

And I would suggest using `$BATS_TEST_DIRNAME` in your `main()` function. 


```bash
main() {
  "${BATS_TEST_DIRNAME}/package_tarball" "$@"
}
```

README.md
```diff
@@ -264,6 +264,8 @@ process.
 For more details about how Bats evaluates test files, see [Bats Evaluation
 Process][bats-eval] on the wiki.
 
+For sample test files, see [examples](/examples).
```
virgilwashere: The directory `/docs/examples` seems like a better home for this content.

```suggestion
For sample test files, see [examples](/docs/examples).
```

Include `examples/README.md` which at the very least should contain a link to the [Wiki Examples page](https://github.com/bats-core/bats-core/wiki/Examples).

```markdown
# Examples

See the [bats-core wiki][examples] for more details

[examples]: (/bats-core/bats-core/wiki/Examples)
```

examples/package-tarball
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -eux
```
virgilwashere: If it's going to be an example, let's follow good principals.

```suggestion
# "unofficial" bash strict mode
# See: http://redsymbol.net/articles/unofficial-bash-strict-mode
set -o errexit  # Exit when simple command fails               'set -e'
set -o errtrace # Exit on error inside any functions or subshells.
set -o nounset  # Trigger error when expanding unset variables 'set -u'
set -o pipefail # Do not hide errors within pipes              'set -o pipefail'
set -o xtrace   # Display expanded command and arguments       'set -x'
IFS=$'\n\t'
```

amilkh:
Thanks so much for the detailed review @virgilwashere !!
I've addressed all your comments, I think, so please let me know any other feedback or go ahead an merge!

AlexSkrypnyk:
@amilkh thank you for working on this. Looks great!

One of the CI jobs has a false positive fail - need maintainers to restart it manually.

amilkh:
Thanks guys for supporting me on this!! Have rebased everything into one commit, so please go ahead and merge when you're ready.

sublimino:
Hello @amilkh (and @virgilwashere and @AlexSkrypnyk!) good teamwork on this one : ] Just FYI -- I'll [add this to the next release](https://github.com/bats-core/bats-core/projects/4) and ensure it's merged after the next refactors are done. 

Thanks all for your hard work, this is a nice clean PR with lots of great conversation. Thank you! 🎉

PR #244, opened 2019-08-21
Feature/add once per file setup and teardown

martin-schulze-vireso:

This is a partial solution to the ongoing discussion in #39 
One issue remains: failing teardown_file will add a fake failed test case report for teardown_file which duplicates the number of the next test after the file. I think this is okay but I'd like to have feedback on that.

Goals of this PR:
- running setup_file/teardown_file regardless of filters
- running teardown_file even on user abort
- tests see variables that are set in setup_file
- skipping tests if setup_file failed

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,7 +572,7 @@ There was an initial [call for maintainers][call-maintain] for the original Bats
 
 ## Copyright
 
-© 2017-2018 bats-core organization
+© 2017-2020 bats-core organization
 
 © 2011-2016 Sam Stephenson
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM bash:${bashver}
 
 # Install parallel and accept the citation notice (we aren't using this in a
 # context where it make sense to cite GNU Parallel).
-RUN apk add --no-cache parallel && \
+RUN apk add --no-cache parallel ncurses && \
     mkdir -p ~/.parallel && touch ~/.parallel/will-cite
 
 RUN ln -s /opt/bats/bin/bats /usr/sbin/bats
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -224,6 +224,14 @@ if [[ "${#arguments[@]}" -eq 0 ]]; then
   abort 'Must specify at least one <test>'
 fi
 
+
+if [[ -n "$output" ]]; then
+  if [[ ! -w "${output}" ]]; then
+    abort "Output path ${output} is not writeable"
+  fi
+  export BATS_REPORT_OUTPUT_PATH="$output"
+fi
+
 filenames=()
 for filename in "${arguments[@]}"; do
   expand_path "$filename" 'filename'
@@ -247,14 +255,6 @@ done
 
 # shellcheck source=lib/bats-core/validator.bash
 source "$BATS_ROOT/lib/bats-core/validator.bash"
-
-if [[ -n "$output" ]]; then
-  if [[ ! -w "${output}" ]]; then
-    printf "Error: output path %s is not writeable\n" "${output}" >&2
-    exit 1
-  fi
- export BATS_REPORT_OUTPUT_PATH="$output"
-fi
 
 set -o pipefail execfail
 exec bats-exec-suite "${flags[@]}" "${filenames[@]}" | bats_test_count_validator | "bats-format-${formatter}" "${formatter_flags[@]}"
--- a/shellcheck.sh
+++ b/shellcheck.sh
@@ -16,3 +16,5 @@ done < <(
   )
 
 LC_ALL=C.UTF-8 shellcheck "${targets[@]}"
+
+exit $?
--- a/test/parallell.bats
+++ b/test/parallell.bats
@@ -51,10 +51,11 @@ setup() {
       [[ "${lines[$i]}" == "ok $i slow test $t" ]]
     done
   done
-  # In theory it should take 3s, but let's give it bit of extra time instead.
-  # also check that parallelization happens accross all files instead of 
+  # In theory it should take 3s, but let's give it bit of extra time for load tolerance.
+  # (Since there is no limit to load, we cannot totally avoid erronous failures by limited tolerance.)
+  # Also check that parallelization happens accross all files instead of
   # linearizing between files, which requires at least 12s
-  [[ "$duration" -lt 12 ]]
+  [[ "$duration" -lt 12 ]] || (echo "If this fails on Travis, make sure the failure is repeatable and not due to heavy load."; false)
 }
 
 @test "setup_file is not over parallelized" {


martin-schulze-vireso:
Parallel jobs might be possible on a file level as we need the setup_file to run before all tests of the file. Suite wide parallelization would only be possible if there are no setup_file functions. Detecting that might be tricky because they don't need to be defined in the .bats file itself.

martin-schulze-vireso:
> It was an explicit design goal of --jobs for parallelism to not be tied in any way to how tests are organised 

Okay, I was not aware of this.

> I  think a much better solution would be to make bats-exec-test aware of whether this is the first (or last) time that a test within a file will be executed ...

I think parallel's semaphore mode might allow for running the tests in parallel across files while still starting them per file.

> (I will admit to not fully understanding why this feature is necessary -- surely most setup and teardown code could just do its own tracking of "is this already set up"? But either way, I think the above proposal should still work for this use-case.)

My rationale for this PR is to achieve reliable per file setup/teardowns even if the test run is aborted or not all tests are run due to filters.

cyphar:
> Okay, I was not aware of this.



> I think parallel's semaphore mode might allow for running the tests in parallel across files while still starting them per file.

Neat -- I didn't know parallel had a semaphore mode!

martin-schulze-vireso:
So I looked into implementing parallelization as described above. I got a nearly working solution now but I stumbled upon something along the way:
GNU parallel does not retain the environment by default. There are ways to achieve this (using `env_parallel` or `parallel --env _`) but they obviously come with performance impacts. 

One of my design goals for this PR was to get the environment passed through from setup_file which is at odds with parallel's default behavior.

I tried to use the bats test suite itself as a benchmark by running it in parallel but that gave me a lot of additional headaches as bats seems not designed to be reentrant or concurrent. However, some of the problems still stem from the environment being culled by parallel.

So this is more of a request for comments: 
Do we want the environment to persist?  I certainly do here and have a hunch that others do as well.
What about the performance impact? Since this only happens with parallelization, should we offer an explicit switch like `--parallel-preserve-environment` for cases where it is needed?

Using `--env _` requires a "manual" setup step of running `parallel --record-env` to populate a list of environment variables that should be ignored so the user should be made aware of this or the missing initialization should be detected (parallel will exit with 255 if `--env _` is used without initialization)

martin-schulze-vireso:
Okay, I think this is working as intended now. @cyphar I made the timing tests stricter so they ensure that running in parallel is really test layout independent. 

One minor problem is, that setup_file and teardown_file are not parallelized via semaphores: the semaphores would get n processes and the file-wide operations too, giving us 2n. However, I don't see this as a common or important enough problem to be worth the hassle of forcing a semaphore solution. A possible implementation would need to run setup_file in the bats-exec-file shell but also in a `parallel --semaphore` to block a semaphore slot. The only solution I can see for this: run a dummy semaphore process that reports when it starts and blocks until it gets notified that it can end. This would then be used as follows:

1. start dummy semaphore process
2. wait for it to report that it got a semaphore slot blocked
3. run setup_ file
4. signal the dummy semaphore process to finish, thereby releasing the blocked slot

However, implementing this seemed to complex to me. The signalling could work via pipes but I am not sure which features are available on all platforms. I nearly implemented a `coprocess` solution only to notice that this requires Bash 4 or higher, which is not the default bash on Mac, so we would have no parallelism there anymore. I had bad experiences with simply using stdin/out because it interferes with the "normal" stdin/out of calling process. (I am also not sure how well this interacts with `parallel`'s semaphore mode). Hence my decision to just leave it restricted to n setup_file/teardown_file processes and n actual tests until this proves to be problematic.

Also note that I introduced another temporary directory `BATS_RUN_TEMPDIR` which is cleaned up after the test run and contains some of the helper files that are necessary to run bats. The cleanup was necessary for the AppVeyor CI builds because they routinely seem to reuse PIDs which leads to filename collisions without the cleanup. Maybe a check if the file/folder already exists should be added to warn of ill effects or abort before anything surprising happens.

README.md
```diff
@@ -200,6 +200,9 @@ Usage: bats [-cr] [-f <regex>] [-j <jobs>] [-p | -t] <test>...
   -f, --filter     Filter test cases by names matching the regular expression
   -h, --help       Display this help message
   -j, --jobs       Number of parallel jobs to run (requires GNU parallel)
+  --parallel-preserve-environment
```
sublimino: Perhaps this should be `--jobs-preseve-env` ? It relates to that flag rather than the implementation IMHO

README.md
```diff
@@ -252,6 +255,12 @@ with dependencies between tests (or tests that write to shared locations). When
 enabling `--jobs` for the first time be sure to re-run bats multiple times to
 identify any inter-test dependencies or non-deterministic test behaviour.
 
+If your code relies on variables from the environment, or from `setup_file()`, 
+you need to specify `--parallel-preserve-environment` as well. Note that this 
+requires running `parallel --record-env` first as a setup step as GNU Parallel
```
sublimino: Is this something we could automate?

README.md
```diff
@@ -370,6 +379,11 @@ You can define special `setup` and `teardown` functions, which run before and
 after each test case, respectively. Use these to load fixtures, set up your
 environment, and clean up when you're done.
 
+You can also define `setup_file` and `teardown_file`, which will run once per file,
+before the first and after the last test, respectively. 
+__WARNING__ these will not be run in parallel mode!
```
sublimino: Can we change "parallel mode" to "--jobs mode" or similar?

What does this mean for users? Could we add wording on the `--jobs` help text?

sublimino:
Apologies @martin-schulze-vireso, I didn't realise this was a blocker for #246. Let's get this moving then.

Could the commits here be squashed down to one (or a few) so they're more discrete please? I've had a first glance but will need some time to go deep on the code : ]

README.md
martin-schulze-vireso: I see where you are coming from and have no hard preference here. A hypothetical other implementation that would not "suffer" from parallel's lack of enviroment propagation would not need this flag though.

README.md
martin-schulze-vireso: I am afraid that a fully automatic detection would amount to executing the code to see what variables it does reference. A simpler solution could be to have a `BATS_PRESERVE_SETUP_FILE_VARIABLES` variable that takes all variable that should be preserved. This is more explicit and gives users the opportunity to make up their own mind instead of taking all variables.

Then there would be the additional requirement of environment variables of the bats run that we might want to forward but that could be solved in the same way.

README.md
martin-schulze-vireso: Good catch, that one is outdated, they also do work with parallel mode now.

martin-schulze-vireso:
> Could the commits here be squashed down to one (or a few) so they're more discrete please? I've had a first glance but will need some time to go deep on the code : ]

I'll see what I can do. I want to get this done this week.

README.md
martin-schulze-vireso: I just noticed that you referred to `parallel --record-env`. There will be an explicit error message if this has not been run before. I opted for explicit user action because the results might be non obvious: this command records the current environment as to ignore. This could result in unintentionally ignoring variables that the user explicitly wants to forward.

martin-schulze-vireso:
@sublimino Okay, I put everything into a single commit now.

sublimino:
@martin-schulze-vireso thank you, I'll continue to review this week :+1: 

@jasonkarns @mbland would appreciate any input here if you have it : ]

martin-schulze-vireso:
@sublimino Did not hear back from you last week, how is the review going?

sublimino:
@martin-schulze-vireso I'm really struggling for time at the moment, I could really do with assistance from other @bats-core maintainers but will persevere when I can find a moment. Thank you for bearing with me for so long 🙏

sublimino:
@btamayo @mbland @jasonkarns 

If this merge or #246 (substantial refactors) have side effects I'm concerned I won't be able to respond quickly enough to the amount of work required to keep master stable. 

As a temporary measure I propose we abort the [1.2 roadmap](https://github.com/bats-core/bats-core/projects/4#card-32105912 and just [cut a release of 1.2](https://github.com/bats-core/bats-core/issues/266) so we have some stability. 

Then we can then merge these two commits (this and #246) with some safety net.

Once that's done as I'm and not succeeding in merging a reasonable period of time, I propose another call for maintainers. 

And copious apologies for @martin-schulze-vireso for the time this has taken to merge, and thanks for his continued patience :pray:

sublimino:
How much is left here now #246 is merged @martin-schulze-vireso, should it be closed or rebased?

martin-schulze-vireso:
IIRC #246 was based on this, so your rebase should have pulled in all if this code anyways.

sublimino:
@martin-schulze-vireso I rebased to make sure, everything did get pulled in it seems, I've merged the minor changes left here for commit message completeness. Thank you! :pray:

PR #153, opened 2018-10-12
Add a JUnit output formatter compatible with Jenkins

josegonzalez:
Refs sstephenson/bats#161, this is a rebase of the original work onto this repository. I'll try and fix any issues that come up until its in a good place.

My reasoning for wanting this PR is to get timing support for Dokku's CircleCI runs. CircleCI only supports junit timing information, and this isn't possible via [tap-xunit](https://github.com/aghassemi/tap-xunit) as there is no timing support (see #107 for more details). Getting timing information would allow us to intelligently speed up our builds without this being a manual guess-and-check process every few months.

Hopefully closes #107.

- [x] I have reviewed the [Contributor Guidelines][contributor].
- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md
[coc]:         https://github.com/bats-core/bats-core/blob/master/docs/CODE_OF_CONDUCT.md

Issue #192, opened 2019-02-27
Provide a bash compatible syntax for bats to allow for other tools to work


lucastheisen:
Right now, the `.bats` files are not valid shell syntax so external tools, [like shellcheck, fail when they try to parse the script](https://github.com/koalaman/shellcheck/issues/709).  This makes it difficult to leverage the benefits of static analysis, while still using this library.  We tried a few workarounds, but all had problems.  It is a simple matter [to support an _additional_ syntax](https://github.com/bats-core/bats-core/pull/191) which is valid bash, allowing for shellcheck to succeed.

sublimino:
Some thoughts:
- are there other syntax types that are more preferable? Options include function names pre/suffixed with strings such as `test_`, annotations/comments, or other magic strings
- could treating comments as code break existing well-known Bats test suites (investigation required)
- what is the rationale for the existing syntax (anything other than parsability?)
- are other linters a target (e.g. shfmt)?

Thanks for starting the conversation @lucastheisen, I agree existing syntax isn't compatible with established tooling. This is a significant enough change that it may warrant a major release anyway. Thoughts welcome!

waterkip:
With Perl one has [Test::Class::Moose](https://metacpan.org/pod/Test::Class::Moose#Declare-a-test-method) where you can have `sub test_foo { ... }` or `sub foo : Test { ... }` or `sub foo : Tests { ... }`. Support for all three of them is very nice. 

I'm in favor of this change, my linter also complains within vim with .bats files.

ko1nksm:
How about this? It is similar to the current syntax. (Add a semicolon after the description)

```sh
@test "addition using bc"; {
  result="$(echo 2+2 | bc)"
  [ "$result" -eq 4 ]
}
```

I think also good to be able to use new line instead of a semicolon.

```sh
@test "addition using bc"
{
  result="$(echo 2+2 | bc)"
  [ "$result" -eq 4 ]
}
```

spl:
I found this issue after encountering the error below when running [`shfmt`](https://github.com/mvdan/sh) on a `.bats` file.

```
test.bats:23:1: "}" can only be used to close a block
```

Are there any known workarounds for using `shfmt` on Bats files?

jasonkarns:
The bats preprocessor generates valid bash. If we support an additional syntax that is valid bash from the get-go, it seems that using the same syntax bats itself generates would be preferable?

spl:
For anyone who might be interested, this is what I did.

[...]

PR #246, opened 2019-10-01
JUnit output resurrected

martin-schulze-vireso:

This is a merge of PR #153 on top of #244. Let's see if we can finally get this landed!

Some remaining todos:

- [x] Validate the JUnit output did not get broken by the merge
- [x] use all combinations of timing and formatter
- [x] shorten `--formatter` values
- [x] document `-T` on bats 
- [x] remove `--extra-flags` (and `--extra-flag` with missing s)
- [ ] credit all original authors


lib/bats-core/tracing.bash
```diff
@@ -0,0 +1,152 @@
+bats_capture_stack_trace() {
```
sublimino: indentation broken in this file

libexec/bats-core/bats
```diff
@@ -26,14 +26,17 @@ Usage: $cmd [-cr] [-f <regex>] [-j <jobs>] [-p | -t] <test>...
...
-  -h, --help       Display this help message
-  -j, --jobs       Number of parallel jobs to run (requires GNU parallel)
-  -p, --pretty     Show results in pretty format (default for terminals)
-  -r, --recursive  Include tests in subdirectories
-  -t, --tap        Show results in TAP format
-  -v, --version    Display the version number
+  -c, --count               Count the number of test cases without running any tests
+  -e, --extra-flags         Extra flags to pass to the bats executor (Options: -T)
+  -f, --filter              Filter test cases by names matching the regular expression
+  -F, --formatter           Switch between formatters (Default: bats-format-tap, Options: bats-format-junit, bats-format-pretty, bats-format-tap)
```
sublimino: Going from `--tap` to `-F bats-format-tap` introduces a lot of keystrokes for a new option. Could the `bats-format-` prefix be dropped?

libexec/bats-core/bats
```diff
@@ -26,14 +26,17 @@ Usage: $cmd [-cr] [-f <regex>] [-j <jobs>] [-p | -t] <test>...
...
+  -h, --help                Display this help message
+  -j, --jobs                Number of parallel jobs to run (requires GNU parallel)
+  -o, --output              A directory to output reports to
+  -p, --pretty              [DEPRECATED] Show results in pretty format (default for terminals), use "--formatter bats-format-pretty" instead
```
sublimino: This is a pretty major flag to deprecate, and introduces something that will force a breaking change to remove. I err on the side of the sysadmin, and would prefer to not deprecate flags that will break scripts in the future when removed.

libexec/bats-core/bats
```diff
@@ -26,14 +26,17 @@ Usage: $cmd [-cr] [-f <regex>] [-j <jobs>] [-p | -t] <test>...
...
+  -r, --recursive           Include tests in subdirectories
+  -t, --tap                 [DEPRECATED] Show results in TAP format, use "--formatter bats-format-tap" instead
```
sublimino: Same deprecation question

README.md
```diff
@@ -196,14 +196,17 @@ Usage: bats [-cr] [-f <regex>] [-j <jobs>] [-p | -t] <test>...
...
+  -e, --extra-flags         Extra flags to pass to the bats executor (Options: -T)
+  -f, --filter              Filter test cases by names matching the regular expression
+  -F, --formatter           Switch between formatters (Default: bats-format-tap, Options: bats-format-junit, bats-format-pretty, bats-format-tap)
+  -h, --help                Display this help message
+  -j, --jobs                Number of parallel jobs to run (requires GNU parallel)
+  -o, --output              A directory to output reports to
+  -p, --pretty              [DEPRECATED] Show results in pretty format (default for terminals), use "--formatter bats-format-pretty" instead
+  -r, --recursive           Include tests in subdirectories
+  -t, --tap                 [DEPRECATED] Show results in TAP format, use "--formatter bats-format-tap" instead
+  -v, --version             Display the version number
```
sublimino: `-T` timing option missing : ]

README.md
```diff
@@ -196,14 +196,17 @@ Usage: bats [-cr] [-f <regex>] [-j <jobs>] [-p | -t] <test>...
...
-  -c, --count      Count the number of test cases without running any tests
-  -f, --filter     Filter test cases by names matching the regular expression
-  -h, --help       Display this help message
-  -j, --jobs       Number of parallel jobs to run (requires GNU parallel)
-  -p, --pretty     Show results in pretty format (default for terminals)
-  -r, --recursive  Include tests in subdirectories
-  -t, --tap        Show results in TAP format
-  -v, --version    Display the version number
+  -c, --count               Count the number of test cases without running any tests
+  -e, --extra-flags         Extra flags to pass to the bats executor (Options: -T)
```
sublimino: This seems a bit cumbersome, can this be replaced by the top-level `-T` (which works fine in `bats -T test/`)

sublimino:
This is awesome @martin-schulze-vireso, thank you for all your hard work on this 🎉 !!

I've done a first pass with some API questions: 
- Can we just pass `-T` for timing? 
- Can we reduce the new flag lengths?
- Do we ever want to deprecate any flags, considering how backwards-compatible we try to be as a project?

You could run shfmt over the code (not the tests yet, notes from @spl in https://github.com/bats-core/bats-core/issues/192#issuecomment-528315083 on why, but that's not adopted in core yet) to squash a few of the whitespace issues. 

I'll need a few days to parse and test this, and would welcome anybody else in @bats-core/bats-core -- this is a sizeable refactor : ]

martin-schulze-vireso:
@sublimino I am still in the process of cleaning up, so not much done in lieu of improvements on top of #153. Yet, here are my 2ct:

> - Can we just pass -T for timing?

Yes, that was my intention.

> - Can we reduce the new flag lengths?

I hope so, they are quite a mouth full. :) 

> - Do we ever want to deprecate any flags, considering how backwards-compatible we try to be as a project?

I think it is not without merit to unify the interface. I would propose to phase out the old switches in the documentation but retain their implementation. Maybe, one day we can remove the old ones because no one is using them anymore.

lib/bats-core/tracing.bash
martin-schulze-vireso: Yeah, my editor had a mind of its own and I noticed too late. I'll look into shfmt as you advised above.

libexec/bats-core/bats
martin-schulze-vireso: Yes, I set that up as a todo in the PR text.

libexec/bats-core/bats-format-junit
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env bash
```
sublimino: I don't know if it's in here, but I'm seeing 25-30s pauses in this formatter when tailing the output file. Can you reproduce?

```bash
$ time bats --formatter bats-format-junit --filter "errors are caught" test
1..2
ok 1 halfway setup_file errors are caught and reported
ok 2 halfway teardown_file errors are caught and reported

real    0m31.577s
user    0m22.649s
sys     0m12.468s

$ time bats --filter "errors are caught" test
 ✓ halfway setup_file errors are caught and reported
 ✓ halfway teardown_file errors are caught and reported

2 tests, 0 failures


real    0m1.116s
user    0m1.056s
sys     0m0.226s
```

README.md
martin-schulze-vireso: I think that is an artifact of the multiple rewrites this has gone through. I believe we can do away with it now.

sublimino:
> @sublimino I am still in the process of cleaning up, so not much done in lieu of improvements on top of #153. Yet, here are my 2ct:
Ah I see, cool -- I got eager! 

> > * Do we ever want to deprecate any flags, considering how backwards-compatible we try to be as a project?
> 
> I think it is not without merit to unify the interface. I would propose to phase out the old switches in the documentation but retain their implementation. Maybe, one day we can remove the old ones because no one is using them anymore.

That could be a decent compromise.

libexec/bats-core/bats-format-junit
martin-schulze-vireso: Nope:
```
$ time ./bin/bats --formatter bats-format-junit --filter "errors are caught" test
1..2
ok 1 halfway setup_file errors are caught and reported
ok 2 halfway teardown_file errors are caught and reported

real    0m0,853s
user    0m0,828s
sys     0m0,180s

$ time ./bin/bats --filter "errors are caught" test
 ✓ halfway setup_file errors are caught and reported 
 ✓ halfway teardown_file errors are caught and reported 

2 tests, 0 failures


real    0m0,482s
user    0m0,444s
sys     0m0,128s
```

martin-schulze-vireso: Just to make sure there was no misunderstanding: By 

> when tailing the output file

you mean you did `tail -f TestReport-.xml` while the above command ran, right?

martin-schulze-vireso: I fixed a bug where not adding `-T` when using `--formatter junit` would break the formatter due to missing timing outputs. Maybe that resolved your issue too?

martin-schulze-vireso:
I somwhat dislike the fact that the junit formatter is actually outputting tap and passing junit to the side. Why couldn't there be pretty output and junit? This hints at the fact, that this is not a mere formatter. I feel like a cleaner solution would be possible.

sublimino:
> I somwhat dislike the fact that the junit formatter is actually outputting tap and passing junit to the side.

If you're suggesting junit to stdout instead of TAP, I was surprised at the behaviour too. However it makes sense in CI -- user has immediate feedback in the build, the server can parse the junit afterwards. I'm not quite sure of the best user experience here.

Are there other machine-readable output formats we may want? Is the concept of a "file logger" (in addition to the standard out TAP/pretty) desirable?

martin-schulze-vireso:
> If you're suggesting junit to stdout instead of TAP, I was surprised at the behaviour too.

No, my suggestion was more in the direction of an orthogonal feature: why not have --formatter pretty in combination with junit? However, this is low priority.

> However it makes sense in CI ...

I am not sure about the details of the JUnit xml format but the formatter generates one .xml report per .bats file in a suite. This would not be possible when redirecting output.

> Are there other machine-readable output formats we may want? Is the concept of a "file logger" (in addition to the standard out TAP/pretty) desirable?

Well, I don't have another format on my wishlist and I did not notice one in the issues. I just wanted to point out that the junit output does not fit the current formatter concept, which may lead to friction down the line.

martin-schulze-vireso:
Is there anything more I can do to facilitate the review process?

sublimino:
@martin-schulze-vireso apologies, I just need to find the time! It's on my list.

sublimino:
Has anybody tested this PR in anger on their test suites?

@martin-schulze-vireso I've got time ahead and I'll be on this during this weekend with a view to merging within the next 5 or so days. Let this comment hold me accountable, thank you for your patience! Any notes of thoughts you might want to add here/DM are appreciated but unnecessary, I'll come back to you if I need anything.

martin-schulze-vireso:
I'd second the request for independent tests, this needs some real workload. @josegonzalez you probably know best how this should behave and it sounded like you have your original code in use. Can you take this on a test ride and see if I broke something?

sublimino:
Another call for independent testing -- this is a major and often requested change. 

CC-ing individuals involved in the original PR in case anybody has spare cycles :pray:  @mmazur @patthiel @stackfull @nkakouros @makefu

makefu:
awesome, will definitly test in the next days

makefu:
Update: i am using the following test file with your PR:
```bats
#!/usr/bin/env bats

@test "skippy" {
  skip
  sleep 1
  echo "skip"
  echo "skip-stderr" >&2
}
@test "success" {
  sleep 1
  echo "success"
  echo "success-stderr" >&2
  true
}
@test "failure" {
  sleep 2
  echo "failure"
  echo "failure-stderr" >&2
  false
}
```

Output of TestReport-lol.bats.xml:
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="lol.bats" tests="3" failures="2" errors="0" skipped="0" time="3" timestamp="2019-12-05T13:33:48" hostname="obby">
  <properties>
    <property name="BATS_ROOT" value="/nix/store/r39idxgq65pmb88i036v89dvdaswiwfa-bats-1.1.0"/>
    <property name="BATS_CWD" value="/home/felix/pki/pki-release-testing/tests"/>
    <property name="BATS_TEST_PATTERN" value="^[[:blank:]]*@test[[:blank:]]+(.*[^[:blank:]])[[:blank:]]+\{(.*)$"/>
    <property name="BATS_VERSION" value="1.2.0-dev"/>
    <property name="BATS_TEST_FILTER" value=""/>
  </properties>
    <testcase classname="lol.bats" name="skippy" time="">
        <failure type="unknown"><![CDATA[
Wrong output format: ok 1 skippy # skip]]></failure>
    </testcase>
    <testcase classname="lol.bats" name="success" time="1"/>
    <testcase classname="lol.bats" name="failure" time="2">
        <failure type="unknown"><![CDATA[
(in test file lol.bats, line 19)
  `false' failed
failure
failure-stderr]]></failure>
    </testcase>
  <system-out></system-out>
  <system-err></system-err>
</testsuite>
```

I would expect the test `skippy` would result in "success" or "skip" as result and `failure` results in failure type "failed" (or what ever is the correct thing here.

Cheers

martin-schulze-vireso:
@makefu Thanks for your input. The skippy error was due to a parsing error when no skip reason was given. I fixed that and adde a regression test.

Regarding the failure type: I could not find some official stand on what should go in there so I just changed it to failure as you suggested.

cyphar:
It looks like you included the commits from #244 in this PR -- can you separate them so they can be reviewed separately?

makefu:
@cyphar sure do, can you give me a commit rev and repo i should use?
Regards

martin-schulze-vireso:
@cyphar I assume you meant me with your comment. I'll try to makie the reply short, since there is some history scattered around the previous PRs: I prepared a separate very intrusive PR (#244) and volunteered to get the JUnit issue going again as both sat idle for a while and since I was on speed with the internals anyways. However, I did not want to invalidate all my work on #244 so I asked if I could bundle them.

Obviously, this creates a rather unwieldy diff here. The only solution that I can see right now is to split them again and merge #244 first, then rebase this PR on master, which should be trivial  since the JUnit part was already rebased on the #244 parts anyways. @sublimino What is your take on this?

sublimino:
@cyphar I found it easier to review this from a diff to master rather than per-commit :/ 

@martin-schulze-vireso my preference would actually be to clean the commit history as there's a lot of experimentation that was started in the previous PR, rolled back, then updated in this one. That said I was looking to get to a well-tested point on this PR before worrying about that. 

Now we have some traction with testing this could be a good time to squash the commit history into something more easily readable. 

Would you be happy to squash this patchset down to a few commits @martin-schulze-vireso? If not I'll look at what I can do, time is short for me right now though.

josegonzalez:
I can try this on Dokku once it is in a good place to review.

martin-schulze-vireso:
@sublimino I just had a look into it and it is a bit more complex than I expected. I forgot that I did not in fact rebase the junit code on top of my setup_file code but merged that. This makes rebasing much harder. I am not sure if we can do that now, since that is not the only merge that happened. I tried to use `git rebase --rebase-merges` but it pulls even more commits up because the junit branch is so old.

I can try anyways but before I'd like to work out what exactly we want to achieve. My understanding is that you want to be able to do more of a commit by commit review instead of a whole PR review, did I get that right?

sublimino:
@martin-schulze-vireso correct. My goals:

1. a clean commit history so a future maintainer can understand the feature by looking at 3..n commits that each deliver something distinct (**IF** that's possible, not a red line)
1. not committing code that is then removed in a later commit in the same PR (similar to 1 really)

That could be a rebase that squashes the whole PR into n commits (which I appreciate can be difficult and may not be worth the time/effort), or worst-case scenario a single commit with a lot of comments and context is the body of the commit message.

Happy for anybody else to chime in if they have opinions here too :pray:

martin-schulze-vireso:
Okay, I lumped together everything into one commit, which is comparatively small. If need be I can try to separate it into several functional parts, e.g. timing information, junit parsing...

I created a diff and then applied it as patch to create this commit, because found no way to rebase them without handling each intermediate commit's merge conflicts. This discarded the original authorship data so we need to put that back in.

sublimino:
@martin-schulze-vireso I don't see the single commit in this PR, could you rebase back to https://github.com/bats-core/bats-core/pull/246/commits/2bec4779b9839be9ac85218ce1fd474c104a621c and force push over the top?

It's possible to run those travis tests locally using the docker container https://travis-ci.org/bats-core/bats-core/jobs/637166600#L194 -- is that not reproducible for you locally?

martin-schulze-vireso:
@sublimino 
> I don't see the single commit in this PR
We are still in the conundrum with  2 PR's in one, the other commits you are seeing are from #244 which is not yet landed on master, so they turn up in this PR too.

> It's possible to run those travis tests locally using the docker container https://travis-ci.org/bats-core/bats-core/jobs/637166600#L194 -- is that not reproducible for you locally?

I did that and got no problems. I also had a travis run on the same commit on my repository and it simply worked. You could try to restart this job and see if the problem remains.

martin-schulze-vireso:
> I don't know what caused the failed test

I looked again and see the error now. The test fails because it is taking to long, which trips up my harsher timing test. I did not see that because the error message for some reason reports the wrong failing line but the debug output clearly shows the test having taken 13 seconds but we require below 12.

martin-schulze-vireso:
@sublimino Seeing that you referenced this issue again as a blocker, I want to get this done ASAP. Is there anything you expect me to do right now or are you reviewing? 

The test failure above won't easily be solvable by adjusting the timing because it is totally dependent on Travis' load. So I am afraid the best we can do is to add a warning to the test, that Travis might trip it up and if it goes away during restarts then there is no problem.

You complained about the other commits, can we get #244 done to simplify this PR?

sublimino:
@martin-schulze-vireso I've rebased all your commits down to one, made some changes in d28a0baac9cdadbf519f5ec4082374c09cd653ab and I think we're ready to merge! Please double check and LMK :pray:

martin-schulze-vireso:
I just reviewed your follow up commit, which consisted mostly of formatting changes. I assume the other one to be unchanged apart from being a squash? If so, I im fine with that.

man/bats.1
```diff
@@ -1,115 +1,94 @@
...
+\fB\-F\fR, \fB\-\-formatter <type>\fR
+Switch between formatters: pretty (default), tap (default w/o term), junit
 .TP
 \fB\-h\fR, \fB\-\-help\fR
-Display help message
-.
+Display this help message
+.TP
+\fB\-j\fR, \fB\-\-jobs <jobs>\fR
+Display this help message
```
martin-schulze-vireso, line 37: This is the wrong help text for this option.

man/bats.1.ronn
```diff
@@ -49,16 +49,28 @@ OPTIONS
...
-  * `-f`, `--filter`:
+  * `-f`, `--filter <regex>`:
     Filter test cases by names matching the regular expression
+  * `-F`, `--formatter <type>`:
+    Switch between formatters: pretty (default), tap (default w/o term), junit
   * `-h`, `--help`:
-    Display help message
+    Display this help message
+  * `-j`, `--jobs <jobs>`:
+    Display this help message
```
martin-schulze-vireso, line 59: Wrong help message

man/bats.1
sublimino: Well spotted, thank you!

sublimino:
Thanks for all your hard work on getting this merged @martin-schulze-vireso, and to @josegonzalez for everything in https://github.com/bats-core/bats-core/pull/153 :tada:

makefu:
wooo! i love to see this feature finally be merged into bats :+1:

andrzejtp:
I'm looking into LAVA integration.

It is somewhat difficult to follow how to add a new formatter.  The 2 patches comprising this PR do much more than just adding JUnit formatter. I wish these changes were more granular. I'll give it a try, though.

martin-schulze-vireso:
@andrzejtp Yes, the PR got quite large. I am not sure what the exact requirements for LAVA are but the JUnit output extended the internal TAP output by some timing and suite information and works as a filter on the TAP output.

In essence you get an extended TAP stream and write a formatter that ingests this stream and creates the corresponding output for your format.

I think the JUnit change would not have been that invasive if more of the necessary information (like the current test file and timing information) were already in place. Now that we have them it might be possible to just write the formatter as a filter on the (extended) TAP outptut.

PR #255, opened 2019-11-30
Expose new env variable to tests: BATS_JOB_SLOT

2opremio:
This new variable is set to the job slot number (`{%}` replacement string from GNU
Parallel) when running tests in parallel (`--jobs` flag). Otherwise it's set to
1 for consistency.

In situations in which tests have resource conflicts when running in parallel
(i.e. tests need dedicated resources) BATS_JOB_SLOT can be used to establish
what resources the test should use.

Dedicated resources will be created per slot beforehand, and the tests will know
which resource to use based on their slot number, guaranteeing no clashes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,6 +459,10 @@ There are several global variables you can use to introspect on Bats tests:
 * `$BATS_TEST_DESCRIPTION` is the description of the current test case.
 * `$BATS_TEST_NUMBER` is the (1-based) index of the current test case in the
   test file.
+* `$BATS_JOB_SLOT` is the (1-based) job slot number being used. Only set
+  relevant when running tests in parallel (`--jobs` flag), otherwise set to 1. 
+  See GNU Parallel's documentation on the `{%}` replacement string 
+  for more details.
 * `$BATS_TMPDIR` is the location to a directory that may be used to store
   temporary files.
 
--- a/libexec/bats-core/bats-exec-suite
+++ b/libexec/bats-core/bats-exec-suite
@@ -93,7 +93,7 @@ if [[ "$num_jobs" != 1 ]]; then
   # amount of overhead using it over a simple loop in the serial case.
   set -o pipefail
   printf '%s\n' "${all_tests[@]}" | grep -v '^$' | \
-    parallel -qk -j "$num_jobs" --colsep="\t" -- bats-exec-test "${flags[@]}" '{1}' '{2}' '{#}' || status=1
+    parallel -qk -j "$num_jobs" --colsep="\t" -- bats-exec-test "${flags[@]}" '{1}' '{2}' '{#}' '{%}' || status=1
 else
   # Just do it serially.
   test_number=0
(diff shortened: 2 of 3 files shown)


2opremio:
This PR illustrates the use of `BATS_JOB_SLOT` https://github.com/fluxcd/flux/pull/2647

In particular, here https://github.com/fluxcd/flux/pull/2647/files#diff-bee7c8cb80c7fa8c7e459049cba62b71R49 and here https://github.com/fluxcd/flux/pull/2647/files#diff-83128850f9d1bfdaf7af07a69281e981R16

stefanprodan:
@sublimino can you please take a look at this when you have some time? Thanks

cyphar:
LGTM, though it wouldn't hurt (in the future) to make `bats-exec-test`'s calling conventions less confusing.

2opremio:
@cyphar Can we please get it merged? Thanks!

cyphar:
I'm not one of the maintainers. @sublimino and the others are the ones who can actually get it merged.

sublimino:
@2opremio I'm waiting to merge #246 first as it's big, I'll get this merged for the next release though

dholbach:
Looks like #246 is merged, but there's apparently a conflict in `libexec/bats-core/bats-exec-suite` - @2opremio do you have time to take a look?

2opremio:
Unfortunately I don't have time anymore.

martin-schulze-vireso:
I am not sure how well this interacts with the semaphore mode as I believe that always returns 1 on '{%}'.

cyphar:
Yeah this will be significantly more complicated because we now run parallel-in-parallel... :thinking:

martin-schulze-vireso:
I am thinking about how to implement this usecase with the new parallelization infrastructure.

@2opremio Could you give a short example? I'm not sure if I understand what you want to achieve.

What I got was:

1. You have resources that must not be shared -> each job needs its own
2. You want to allocate as few resources as possible -> at most n resources if there are n jobs in parallel
3. you want to reuse the resource of another test on the same job slot -> the first test on the slot must allocate the resource

What I don't know/understand:

- What kind of resources are we talking about?
- In which pattern are they used?

2opremio:
> @2opremio Could you give a short example? I'm not sure if I understand what you want to achieve.

See https://github.com/bats-core/bats-core/pull/255#issuecomment-559972456

This PR is 8 months old now, and I have since moved on to other work. Good luck!

martin-schulze-vireso:
@2opremio 

> See #255 (comment)

Yes, I already looked into that but I don't use Kubernetes, so I don't understand what is going on there. Hence my questions what you wanted to achieve.

2opremio:
Map tests to resources, so that multiple parallel tests don't share the same resource (in the example, the resources are Kubernetes clusters in which to run the tests).

martin-schulze-vireso:
After mulling over an implementation for this I came to following conclusion:

1. mimicking the intended behavior of BATS_JOB_SLOT from this PR would entail an own implementation because parallel's semaphore mode does not help us here
2. the implementation would likely entail some kind of resource pool (e.g. the kubernetes cluster names?) from which resources can be taken and given back

This approach has the following advatnages:
1. it is more general than BATS_JOB_SLOT
2. it is not necessarily part of bats-core

However, to function correctly, we must ensure that resources are given always back when not needed anymore. The teardown function is probably the best place for that.

martin-schulze-vireso:
Seeing that this PR would need substantial rewrites to be mergable again, I will close this now. If this is still a required feature I would suggest to open an issue for further discussion. The changes from #339 would allow for a relatively straightforward implementation. However, I still think that a more general solution that does not require changes to bats-core would be better.

PR #260, opened 2020-01-21
Add a default workdir to the Docker image

felipecassiors:
This way we don't need to use the full path of the mounted folder
as long as we mount to the default working directory which is
`/code/` according to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ To run Bats' internal test suite (which is in the container image at
 To run a test suite from your local machine, mount in a volume and direct Bats
 to its path inside the container:
 
-    $ docker run -it -v "$(pwd):/code" bats/bats:latest /code/test
+    $ docker run -it -v "$(pwd):/code" bats/bats:latest test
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -170,7 +170,7 @@ To run Bats' internal test suite (which is in the container image at
 To run a test suite from your local machine, mount in a volume and direct Bats
 to its path inside the container:
 
-    $ docker run -it -v "$(pwd):/code" bats/bats:latest /code/test
+    $ docker run -it -v "$(pwd):/code" bats/bats:latest /test
```
sublimino: Should this be `./test` instead of `/code/test`?

felipecassiors: Oh, yeah! Actually just `test` would make the trick, but `/test` is wrong indeed.

sublimino:
Thanks for your contribution @felipecassiors, I've rebased this on to https://github.com/bats-core/bats-core/pull/274 and will merge `/code` as the WORKDIR. Thank you :D :100:

PR #274, opened 2020-04-12
Update Docker infrastructure for bats

waterkip:
Bats is no longer installed in `/usr/sbin`, bats isn't a sbin candidate. It is
moved to `/usr/local/bin`. This should not have any side effects as
`/usr/local/bin` should be in ones PATH.

Have a default docker-compose.yml for users that they can use to easily play
with docker-compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,10 +169,14 @@ To run Bats' internal test suite (which is in the container image at
 
     $ docker run -it bats/bats:latest /opt/bats/test
 
-To run a test suite from your local machine, mount in a volume and direct Bats
-to its path inside the container:
+To run a test suite from a directory called `test` in the current directory of
+your local machine, mount in a volume and direct Bats to its path inside the
+container:
 
-    $ docker run -it -v "$(pwd):/opt/bats" bats/bats:latest /opt/bats/test
+    $ docker run -it -v "${PWD}:/code" bats/bats:latest test
+
+> `/code` is the working directory of the Docker image. "${PWD}/test" is the
+> location of the test directory on the local machine.
 
 This is a minimal Docker image. If more tools are required this can be used as a
 base image in a Dockerfile using `FROM <Docker image>`.  In the future there may
@@ -180,7 +184,8 @@ be images based on Debian, and/or with more tools installed (`curl` and `openssl
 for example). If you require a specific configuration please search and +1 an
 issue or [raise a new issue](https://github.com/bats-core/bats-core/issues).
 
-Further usage examples are in [the wiki](https://github.com/bats-core/bats-core/wiki/Docker-Usage-Examples).
+Further usage examples are in
+[the wiki](https://github.com/bats-core/bats-core/wiki/Docker-Usage-Examples).
 
 ## Usage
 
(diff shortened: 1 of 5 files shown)


waterkip:
There is another PR that changes the WORKDIR to `/code`, please have a look and maybe we can incorporate that into here or add that one on top of this one:
https://github.com/bats-core/bats-core/pull/260

sublimino:
@waterkip I've rebased the `/code` change onto this repo, I think we're good to go. This now:

- move `bats` from `/usr/sbin` to `/usr/local/bin` in Docker image 
- set WORKDIR to `/code` in Docker image
- add docker-compose example file

I'm a little concerned about people relying on the previous behaviour with bats' path and workdir, but there's a release cut for `1.2.0` so I'm more inclined to ship changes now!

Please check this over and LMK what you think.

waterkip:
According the the FHS[1] /usr/sbin is exclusivly used for things used by the system admin, bats is used by a user. And thus violates the FHS by having it in /usr/sbin. Users have been relying on a bug if they rely on it. I suppose they mostly relied on it being in PATH and this change doesn't affect that reliance.

As to the WORKDIR, it used to be `/`, you cannot really mount stuff nicely in there because docker mounts are recursive so you need to mount stuff. So I think one tends to mount the project dir in a seperate directory and call bats with that separate directory. So yeah.. It could potentially break if people mounted stuff in `/`, mounted with -v $(pwd)/projects:/projects' and called the docker with `bats ./projects/tests`. That said, when I rely on a another Docker project either by consuming it in a custom Dockerfile by having my own WORKDIR or if I use docker-compose I tend to make sure it fits my needs and I use a own project directory and call that thing by its absolute path.

I cannot say that their isn't any impact with these changes. There is a chance that a consumer doesn't do the things I think are logical. If that is an acceptable risk to you, go for it, if it isn't then don't. I've proposed changes to the WORKDIR and so did someone else, so having something else that `/` seems wanted. I just don't know by how many and how people are going to respond to the change.

1: http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s10.html

sublimino:
Thanks for the detailed reply @waterkip : ] All agreed. I've rebased everything and added it to this PR in https://github.com/bats-core/bats-core/pull/274/files — now we have a release out I'll merge this and see if it causes anybody problems, then decide how to proceed as appropriate as (as you correctly state) the previous incarnation still had problems. Thanks again! :tada:

README.md
```diff
@@ -169,18 +169,23 @@ To run Bats' internal test suite (which is in the container image at
 
     $ docker run -it bats/bats:latest /opt/bats/test
 
-To run a test suite from your local machine, mount in a volume and direct Bats
-to its path inside the container:
+To run a test suite from a directory called `test` in the current directory of
+your local machine, mount in a volume and direct Bats to its path inside the
+container:
 
-    $ docker run -it -v "$(pwd):/opt/bats" bats/bats:latest /opt/bats/test
```
dimo414, line 175: I believe the intent of this line was actually to help users run Bats' own tests, not to run their own Bats tests (note the "To run Bats' internal test suite" line above).

It's certainly helpful to provide an example for running your own Bats tests, but this section still reads strangely.

sublimino: The code still works to test `bats-core`, but I take your point on the phrasing changes. 

Perhaps the sentence could be prefixed with: 

> To run arbitrary tests (this example still works for the `bats-core` directory, but mounted into the container instead of reading the files from the container image)

Wordier, but perhaps clearer?

dimo414: I'd suggest splitting the two purposes apart more fully; a section on "how to run your own tests" and a section on "how to test bats itself".

Issue #253, opened 2019-11-18
misidentified failing line


stevenjswanson:
I've found an instance where bats reports the failure on an incorrect line:

With this in `foo.bats`
```
@test "good" {
    function foo () {
	true;
	return 0;
    }
    #FOO=yes
    foo;
    ! [ -v FOO ]
}

@test "bad" {
    function foo () {
	true;
	return 0;
    }
    FOO=yes
    foo;
    ! [ -v FOO ]
}
```
I get: 

```
$ bats foo.bats | cat
1..2
ok 1 good
not ok 2 bad
# (from function `foo' in file foo.bats, line 15,
#  in test file foo.bats, line 18)
#   `foo;' failed
```
The `bad` test should fail, but the failure is on `! [ -v FOO ]` rather than `foo`.

I'm on `master` at commit `87b16eb8ec381beca7f05d028e4ad7cf3f647a0f`.

kgroat:
For your test cases, if you move the `!` inside, it should start reporting the correct line.  Like:
```bats
[ ! -v FOO ]
```

(Also of note -- the `-v` expression [isn't actually POSIX compatible](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html#tag_20_128_05) and some implementations don't have it, like BSD-based OS's.  If you need your bats scripts to be portable, you may want to keep that in mind.  If you're always going to be running on the same OS, then feel free to ignore this little tangent)

That being said, I'm seeing a similar issue when I use `[[` for an assertion:
```bats
@test "why" {
  echo "this is not the bad line"
  [[ "this" =~ "is the bad line" ]]
}
```
The report shows:
```diff
 ✗ why
   (in test file test/shared_functions.bats, line 26)
     `echo "this is not the bad line"' failed
```

My best guess is that this is because both `!` and `[[` are bash builtins, while `[` is actually an executable (usually found at `/bin/[` or `/usr/bin/[` but given *priority* like it's a builtin).

strager:
I encountered this issue as well. I am working on a fix right now.

martin-schulze-vireso:
Fixed since the PR above has been merged. However, you should be aware that using `!` on any but the last command will effectively disable failure exits. See https://bats-core.readthedocs.io/en/stable/gotchas.html#my-negated-statement-e-g-true-does-not-fail-the-test-even-when-it-should

PR #279, opened 2020-04-18
Fix incorrect line information on test failures

strager:

Bats tries hard to tell users the failing line when a test fails. For
example:

     ✗ run tests which consume stdin (see #197)
       (in test file test/bats.bats, line 516)
         `[ "$status" -eq 1 ]' failed

This line reporting is broken for a number of cases. For example, Bats
reports the wrong line in the following cases:

* With older versions of bash (such as the version deployed on macOS
  10.12.6), when a `[[` or `((` check fails.

* When a command succeeds but its success was turned into a failure
  using a `!` prefix.

* When a command inside a function fails, but the failure is handled by
  the caller using `||`, but the failure handler also fails.

Redesign Bats' interpretation of traps to fix these cases. The new
design is very fragile, but works for more cases than the old design.

This should fix issue #253.


strager:
I'm not picky about the variable names. I did find BATS_CURRENT_STACK_TRACE vs BATS_STACK_TRACE (from the original code) confusing. I don't like BATS_DEBUG_LAST_STACK_TRACE vs BATS_DEBUG_LASTLAST_STACK_TRACE, but I couldn't think of anything better. I'd love to hear your suggestions.

strager:
ping. This is ready for review.

sublimino:
Thanks for the effort and documentation you've put into this PR @strager, it looks excellent. Very easy to parse, thank you :tada: 

An initial thought is that the actual line number that's error-ing in the test cases https://github.com/bats-core/bats-core/pull/279/files#diff-345d20dffe74505b7fee3680715e50f5R134-R160 could be changed so the tests assert for different numbers of lines and different error-ing lines, unless that's intentional? 

I'd like more eyes on this than just me (which is all the project has at the moment). Would anybody else that's been active recently like to take a look at this with me? /cc @jasonkarns @martin-schulze-vireso @dimo414 @Vampire @waterkip

lib/bats-core/tracing.bash
```diff
@@ -117,15 +127,61 @@ bats_trim_filename() {
 	printf -v "$2" '%s' "${1#$BATS_CWD/}"
 }
 
+# bats_debug_trap tracks the last line of code executed within a test. This is
```
dimo414, line 130: This analysis is really helpful, maybe it should be in the wiki rather than inline?

lib/bats-core/tracing.bash
```diff
@@ -117,15 +127,61 @@ bats_trim_filename() {
...
 	# don't update the trace within library functions or we get backtraces from inside traps
 	if [[ "$1" != $BATS_ROOT/lib/* && "$1" != $BATS_ROOT/libexec/* ]]; then
-		# The last entry in the stack trace is not useful when en error occured:
-		# It is either duplicated (kinda correct) or has wrong line number (Bash < 4.4)
-		# Therefore we capture the stacktrace but use it only after the next debug
-		# trap fired.
-		# Expansion is required for empty arrays which otherwise error
-		BATS_CURRENT_STACK_TRACE=("${BATS_STACK_TRACE[@]+"${BATS_STACK_TRACE[@]}"}")
+		BATS_DEBUG_LASTLAST_STACK_TRACE=(
+			"${BATS_DEBUG_LAST_STACK_TRACE[@]+"${BATS_DEBUG_LAST_STACK_TRACE[@]}"}"
```
dimo414, line 180: You may just be emulating the same pattern used elsewhere in this project, but the outer-quotes in these `${arr[@]+"${arr[@]}"}` expansions should be removed. I dug into this a few weeks ago because I couldn't find a convincing breakdown: https://stackoverflow.com/a/61551944/113632

lib/bats-core/tracing.bash
```diff
@@ -23,6 +23,15 @@ bats_capture_stack_trace() {
 	done
 }
 
+bats_get_failure_stack_trace() {
+	# See bats_debug_trap for details.
+	if [[ -n "${BATS_DEBUG_LAST_STACK_TRACE_IS_VALID}" ]]; then
+		eval "${1}"'=("${BATS_DEBUG_LAST_STACK_TRACE[@]}")'
```
dimo414, line 29: You may need the same `${arr[@]+"${arr[@]}"}` expansion as below (maybe not, just calling it out).

dimo414, line 29: I like combining `eval` with `printf` to make it easier to see what's being `eval`-ed. You could write this as:

    eval "$(printf '%s=("${BATS_DEBUG_LAST_STACK_TRACE[@]}")' "$1")"

Which makes it (more) obvious the only dynamic behavior is the variable name being assigned.

lib/bats-core/tracing.bash
```diff
@@ -117,15 +127,61 @@ bats_trim_filename() {
...
+# Below are tables describing different command failure scenarios and the
+# reliability of $BASH_LINENO within different the executed DEBUG, ERR, and EXIT
+# trap handlers. Naturally, the behaviors change between versions of Bash.
+#
+# Table rows should be read left to right. For example, on bash version
+# 4.0.44(2)-release, if a test executes `false` (or any other failing external
+# command), bash will do the following in order:
+# 1. Call the DEBUG trap handler (bats_debug_trap) with $BASH_LINENO referring
+#    to the source line containing the `false` command, then
+# 2. Call the DEBUG trap handler again, but with an incorrect $BASH_LINENO, then
```
dimo414, line 143: I don't know that it's fair to call these "incorrect"; the DEBUG trap is firing prior to three different commands; the user's command, the ERR trap command, and the EXIT trap command. Are the line numbers actually incorrect, or just not what we care about?

lib/bats-core/tracing.bash
dimo414: It'd be great if you could share your methodology too, e.g. in a Gist, so we can reproduce what you're seeing.

lib/bats-core/tracing.bash
```diff
@@ -117,15 +127,61 @@ bats_trim_filename() {
...
-		# It is either duplicated (kinda correct) or has wrong line number (Bash < 4.4)
-		# Therefore we capture the stacktrace but use it only after the next debug
-		# trap fired.
-		# Expansion is required for empty arrays which otherwise error
-		BATS_CURRENT_STACK_TRACE=("${BATS_STACK_TRACE[@]+"${BATS_STACK_TRACE[@]}"}")
+		BATS_DEBUG_LASTLAST_STACK_TRACE=(
+			"${BATS_DEBUG_LAST_STACK_TRACE[@]+"${BATS_DEBUG_LAST_STACK_TRACE[@]}"}"
+		)
+
+		BATS_DEBUG_LAST_LINENO=("${BASH_LINENO[@]+"${BASH_LINENO[@]}"}")
```
dimo414, line 183: Is it worth capturing `${FUNCNAME[@]}` too? `BASH_SOURCE`, `BASH_LINENO`, and `FUNCNAME` are all connected, so if we're capturing two we'd probably want to capture all three.

dimo414:
Left some comments, mostly just nits. At a high-level this approach makes sense, but I'm curious if there's a cleaner way it could be implemented. As-is we're basically "guessing" that the `LASTLAST` entry will be right, instead of "knowing" that we've captured the right lines.

From the inline table it seems like if we can properly detect the three different situations where the DEBUG trap fires (the current command, the ERR trap, and the EXIT trap) we could skip capturing a trace from the latter two events and therefore always have the right line numbers. I don't know how feasible that is, but I think it would be a clearer and more reliable approach if it's possible.

@rcaloras might have some suggestions too, since bash-preexec has had to do plenty of DEBUG-trap wrestling already (sorry to rope you into this Ryan 😉).

rcaloras:
@dimo414, think your comments make sense 👍Nothing else on my end.

martin-schulze-vireso:
Without having looked to deep into this I want to mention one thing I already noticed before while doing the separation into the lib files: There is a lot of code that tries to create backwards compatibility and I wonder at what cost it comes.

For one, the workarounds might penalize systems that don't need them, by requiring far more operations. Secondly, we riddle our code with switches that should handle complex exceptional cases that only arise on some systems, which makes the code harder to maintain due to higher complexity and interdependencies. 

I would propose to separate version specific fixes into version specific library files that are only loaded on that platform. I'd hope to get some improved readability and performance out of that. However, I don't know if this makes implementations for the exceptional systems harder.

sublimino:
@martin-schulze-vireso this makes good sense, I fear we do not have enough test coverage in Travis for this yet though. We should expand CI to cover our [supported versions](https://github.com/bats-core/bats-core/#supported-bash-versions) whilst trying to keep cycle time reasonable, then we can refactor complex code to library functions in relative safety.

ItsCam57:
penis @strager

lib/bats-core/tracing.bash
strager: > It'd be great if you could share your methodology too, e.g. in a Gist, so we can reproduce what you're seeing.

Sure! Here's the script I used to inspect Bash's behavior: https://gist.github.com/strager/dd5370f27e96c0d66deebc6ab051d421

lib/bats-core/tracing.bash
strager: > I don't know that it's fair to call these "incorrect"

The behavior changed between Bash 4.0.44 and Bash 4.4.20, so my assumption is that Bash fixed a bug (or unintentionally changed behavior? unsure). I'm calling the Bash 4.0.44 behavior incorrect, compared to Bash 4.4.20's behavior.

lib/bats-core/tracing.bash
strager: > Is it worth capturing `${FUNCNAME[@]}` too? `BASH_SOURCE`, `BASH_LINENO`, and `FUNCNAME` are all connected, so if we're capturing two we'd probably want to capture all three.

Perhaps. My assumption was that BASH_SOURCE+BASH_LINENO uniquely identifies a line of code. Do you think those two are insufficient and that FUNCNAME would be a good tie-breaker?

strager:
> An initial thought is that the actual line number that's error-ing in the test cases https://github.com/bats-core/bats-core/pull/279/files#diff-345d20dffe74505b7fee3680715e50f5R134-R160 could be changed so the tests assert for different numbers of lines and different error-ing lines, unless that's intentional?

I don't know what you mean. Are you saying that, for example, failing_with_bash_cond should fail on line 4, failing_with_bash_expression should fail on line 5, and failing_with_negated_command should fail on line 6?

> Left some comments, mostly just nits. At a high-level this approach makes sense, but I'm curious if there's a cleaner way it could be implemented. As-is we're basically "guessing" that the LASTLAST entry will be right, instead of "knowing" that we've captured the right lines.
>
> From the inline table it seems like if we can properly detect the three different situations where the DEBUG trap fires (the current command, the ERR trap, and the EXIT trap) we could skip capturing a trace from the latter two events and therefore always have the right line numbers. I don't know how feasible that is, but I think it would be a clearer and more reliable approach if it's possible.

I tried! My original solution was to ask Bash for the command which failed. Unfortunately, I couldn't get it working for all the cases. For example, for all versions of Bash that I tried, `${BASH_COMMAND}` does not tell me whether the command was executed with `!` or not, so we cannot distinguish between `false` and `! true`.

> Without having looked to deep into this I want to mention one thing I already noticed before while doing the separation into the lib files: There is a lot of code that tries to create backwards compatibility and I wonder at what cost it comes.
>
> For one, the workarounds might penalize systems that don't need them

It's not just for backwards compatibility. This patch fixes some issues with Bash 4.0.44, yes, but it also fixes some issues with newer versions of Bash too. The wonkiness needed to fix line numbers for Bash 4.0.44 also fixes line numbers for Bash 4.4.20. In other words, all versions of Bash that I tested need these "workarounds".

dimo414:
> I tried! My original solution was to ask Bash for the command which failed. Unfortunately, I couldn't get it working for all the cases.

Can you share what you tried so we can reproduce what you saw? Maybe as a separate (throwaway) PR, if that would be easiest. It certainly wouldn't surprise me if this approach isn't feasible, but it would be good to validate that is the case (and ideally capture why it's not feasible, in case the situation improves in the future).

martin-schulze-vireso:
@dimo414 I extracted the commands from the tests and wrote up my own script to test here https://gist.github.com/martin-schulze-vireso/05e5ba81b8700af960558e32bab60c1e when calling it without parameters it calls itself with each of the commands at the beginning.

`./test_bash_quirks.bash` gives me the following output for bash 5.0:
```
Correctly failed COMMAND: (( 1 == 0 ))
DEBUG ?64: (( 1 == 0 ))
DEBUG ?64: (( 1 == 0 ))
FAILED ?64: (( 1 == 0 ))
DEBUG ?1: (( 1 == 0 ))
EXITED ?1: (( 1 == 0 ))

Correctly failed COMMAND: [[ 1 -eq 0 ]]
DEBUG ?64: [[ 1 -eq 0 ]]
DEBUG ?64: [[ 1 -eq 0 ]]
FAILED ?64: [[ 1 -eq 0 ]]
DEBUG ?1: [[ 1 -eq 0 ]]
EXITED ?1: [[ 1 -eq 0 ]]

Correctly failed COMMAND: false
DEBUG ?64: false
DEBUG ?64: false
FAILED ?64: false
DEBUG ?1: false
EXITED ?1: false

Correctly failed COMMAND: true_function && false
DEBUG ?64: true_function
DEBUG ?64: false
DEBUG ?64: false
FAILED ?64: false
DEBUG ?1: false
EXITED ?1: false

Incorrectly passed COMMAND: ! true_function
DEBUG ?64: true_function
DEBUG ?65: true
DEBUG ?66: trap - ERR EXIT DEBUG

Incorrectly passed COMMAND: false_function
DEBUG ?64: false_function
EXITED 64?1: false

Correctly failed COMMAND: false_function || false
DEBUG ?64: false_function
DEBUG ?64: false
DEBUG ?64: false
FAILED ?64: false
DEBUG ?1: false
EXITED ?1: false

Correctly failed COMMAND: f() { false_function || return 2; }; f
DEBUG ?64: f
DEBUG ?64: return 2
FAILED ?64: return 2
DEBUG ?1: return 2
EXITED ?1: return 2

Correctly failed COMMAND: return_0 && false
DEBUG ?64: return_0
DEBUG ?64: false
DEBUG ?64: false
FAILED ?64: false
DEBUG ?1: false
EXITED ?1: false

Correctly failed COMMAND: f() { return_0 && return 2; }; f
DEBUG ?64: f
DEBUG ?64: return 2
FAILED ?64: return 2
DEBUG ?1: return 2
EXITED ?1: return 2

Incorrectly passed COMMAND: ! return_0
DEBUG ?64: return_0
DEBUG ?65: true
DEBUG ?66: trap - ERR EXIT DEBUG

Correctly failed COMMAND: return_1 || false
DEBUG ?64: return_1
DEBUG ?64: false
DEBUG ?64: false
FAILED ?64: false
DEBUG ?1: false
EXITED ?1: false

Correctly failed COMMAND: f() { return_1 || return 2;  }; f
DEBUG ?64: f
DEBUG ?64: return 2
FAILED ?64: return 2
DEBUG ?1: return 2
EXITED ?1: return 2

Correctly failed COMMAND: return_1
DEBUG ?64: return_1
DEBUG ?64: return 1
FAILED ?64: return 1
DEBUG ?1: return 1
EXITED ?1: return 1

```

One obvious problem with relying on `$BASH_COMMAND` is with compounds like `return_1 || return 2` where only part of the compound will be shown. We also get some bad behavior on failures within functions and the reported line is always lower by one than the actual.

Petemtl (approved):

martin-schulze-vireso:
I finally took the time to dive a bit deeper into this and rebase it onto current master. Unfortunately, there is some bad interaction between the CTRL-C mid test aborts and this PRs efforts that break the reports but I think that will be fixable. More to the point of the review:

https://github.com/bats-core/bats-core/pull/279#issuecomment-631230105
> Can you share what you tried so we can reproduce what you saw? Maybe as a separate (throwaway) PR, if that would be easiest. It certainly wouldn't surprise me if this approach isn't feasible, but it would be good to validate that is the case (and ideally capture why it's not feasible, in case the situation improves in the future).

I tried to reproduce with the linked trapee scripts and this is what I saw:
1. the BASH_COMMAND variable will remain the same between the first and second call to debug_trap. However, I did not find an easy solution to distinguish an error from simply calling the same command twice in a row 
2. I can corroborate that it does not contain the `!`

My summary is: The current approach of storing two traces and choosing in the err trap is actually the most robust.

Now to more pressing issues with the tests that I noticed just recently in the PR for #462: `!` related "failures" don't trigger the ERR trap. This means they can only fail a test if they are the last command before returning from a function, which makes the function return with an error that then gets caught by the ERR trap. The test cases in trapee and in the test suite play heavily on that by having all commands be the "returning" command in a function. This PR actually "fixes" this by reporting the provoking command as failure instead of the returning function. While I think that this is an improvement in readability, it directs unaware users into a dangerous corner: it is not obvious why this fails correctly:

```
@test test() {
    true
    ! true
}
```

while this does not:

```
@test test () {
    ! true
    true
}
```

for Bash 3 you can do this with `(())` and `[[]]` failures instead of `!`, too.

Overall, I think this PR is a net positive and we have to deal with the `!` (and `(())`/`[[]]`) issue via shellcheck.

strager:
I don't have time to rebase or champion this PR myself. Feel free to take it over.

martin-schulze-vireso:
I rebased and made some fixes. However, I can't Push these changes here. @strager die you allow maintainers to push?

strager:
> I can't Push these changes here. @strager die you allow maintainers to push?

No.

I enabled the option just now. You can try again.

martin-schulze-vireso:
Thanks, I'll try again later.

PR #282, opened 2020-04-23
Reimplement load to be more useable and consistent

### LINKED ISSUE #79: load: allow loading relative paths of arbitrary extension
I'm not sure what the goal of expecting a helper script to be an extension of `.bash` is, but it makes absolutely zero sense.

Let me specify a path and load that file. Don't force the extension on me.

dimo414:
The existing behavior of `load` accepts absolute paths or relative paths - without a `.bash` suffix. This is inconsistent and confusing, both because the `.bash` suffix is arbitrary but also because absolute paths don't have the same constraint.

`load` is re-implemented to behave as a more consistent shorthand for `source`-ing files relative to `BATS_TEST_DIRNAME`, while falling back to `source` the argument as-is if the relative path does not resolve. The `.bash` suffix behavior is preserved for backwards compatibility, but any relative path is now supported. In addition (since `source` supports this natively) scripts on the system `PATH` can also be `load`-ed.

Fixes #79

---

### DIFF
--- a/lib/bats-core/test_functions.bash
+++ b/lib/bats-core/test_functions.bash
@@ -3,24 +3,30 @@
 BATS_TEST_DIRNAME="${BATS_TEST_FILENAME%/*}"
 BATS_TEST_NAMES=()
 
+# Shorthand for source-ing files relative to the BATS_TEST_DIRNAME,
+# optionally with a .bash suffix appended. If the argument doesn't
+# resolve relative to BATS_TEST_DIRNAME it is sourced as-is.
 load() {
-  local name="$1"
-  local filename
+  local file="${1:?}"
 
-  if [[ "${name:0:1}" == '/' ]]; then
-    filename="${name}"
-  else
-    filename="$BATS_TEST_DIRNAME/${name}.bash"
+  # For backwards-compatibility first look for a .bash-suffixed file.
+  # TODO consider flipping the order here; it would be more consistent
+  # and less surprising to look for an exact-match first.
+  if [[ -f "${BATS_TEST_DIRNAME}/${file}.bash" ]]; then
+    file="${BATS_TEST_DIRNAME}/${file}.bash"
+  elif [[ -f "${BATS_TEST_DIRNAME}/${file}" ]]; then
+    file="${BATS_TEST_DIRNAME}/${file}"
   fi
 
-  if [[ ! -f "$filename" ]]; then
-    printf 'bats: %s does not exist\n' "$filename" >&2
+  if [[ ! -f "$file" ]] && ! type -P "$file" >/dev/null; then
+    printf 'bats: %s does not exist\n' "$file" >&2
     exit 1
   fi
 
-  # Dynamically loaded user files provided outside of Bats.
+  # Dynamically loaded user file provided outside of Bats.
+  # Note: 'source "$file" || exit' doesn't work on bash3.2.
   # shellcheck disable=SC1090
-  source "${filename}"
+  source "${file}"
 }
 
 run() {
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -196,6 +196,11 @@ fixtures bats
   [ $status -eq 0 ]
 }
 
+@test "load sources relative scripts with filename extension" {
+  HELPER_NAME="test_helper.bash" run bats "$FIXTURE_ROOT/load.bats"
+  [ $status -eq 0 ]
+}
+
 @test "load aborts if the specified script does not exist" {
   HELPER_NAME="nonexistent" run bats "$FIXTURE_ROOT/load.bats"
   [ $status -eq 1 ]
@@ -209,6 +214,19 @@ fixtures bats
 @test "load aborts if the script, specified by an absolute path, does not exist" {
   HELPER_NAME="${FIXTURE_ROOT}/nonexistent" run bats "$FIXTURE_ROOT/load.bats"
   [ $status -eq 1 ]
+}
+
+@test "load relative script with ambiguous name" {
+  HELPER_NAME="ambiguous" run bats "$FIXTURE_ROOT/load.bats"
+  [ $status -eq 0 ]
+}
+
+@test "load supports scripts on the PATH" {
+  path_dir="$BATS_TMPNAME/path"
+  mkdir -p "$path_dir"
+  cp "${FIXTURE_ROOT}/test_helper.bash" "${path_dir}/on_path"
+  PATH="${path_dir}:$PATH"  HELPER_NAME="on_path" run bats "$FIXTURE_ROOT/load.bats"
+  [ $status -eq 0 ]
 }
 
 @test "output is discarded for passing tests and printed for failing tests" {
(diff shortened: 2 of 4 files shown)


libexec/bats-core/bats-exec-test
```diff
@@ -40,24 +40,21 @@ fi
...
-  if [[ "${name:0:1}" == '/' ]]; then
-    filename="${name}"
-  else
-    filename="$BATS_TEST_DIRNAME/${name}.bash"
-  fi
+  local file="${1:?name}"
 
-  if [[ ! -f "$filename" ]]; then
-    printf 'bats: %s does not exist\n' "$filename" >&2
-    exit 1
```
sublimino, line 55: Could we re-add this error message please? So users don't notice the change : ]

libexec/bats-core/bats-exec-test
```diff
@@ -40,24 +40,21 @@ fi
 BATS_TEST_DIRNAME="${BATS_TEST_FILENAME%/*}"
 BATS_TEST_NAMES=()
 
+# Shorthand for source-ing files relative to the BATS_TEST_DIRNAME,
+# optionally with a .bash suffix appended. If the argument doesn't
+# resolve relative to BATS_TEST_DIRNAME it is sourced as-is.
 load() {
-  local name="$1"
```
sublimino, line 44: Apologies for my naivety  — I usually use `:-`, can you explain the behaviour here for me please?

libexec/bats-core/bats-exec-test
```diff
@@ -40,24 +40,21 @@ fi
...
+    file="${BATS_TEST_DIRNAME}/${file}"
+  elif [[ -f "${BATS_TEST_DIRNAME}/${file}.bash" ]]; then
+    file="${BATS_TEST_DIRNAME}/${file}.bash"
   fi
 
-  # Dynamically loaded user files provided outside of Bats.
+  # Dynamically loaded user file provided outside of BATS.
   # shellcheck disable=SC1090
-  source "${filename}"
+  source "${file}" || exit
```
sublimino, line 57: What was the previous behaviour here, silent fail? If so perhaps this could be something like

```bash
if ! source "${file}"; then
  printf error message
fi
```

dimo414:
Thanks for the quick review! Looks like there's some existing tests that are affected, I'll tweak them and update the PR.

libexec/bats-core/bats-exec-test
dimo414: It would return a non-zero exit code (and print an error), but allow the script to proceed. I imagine the prior assumption was that sourcing would always succeed if the file existed, which isn't necessarily true (e.g. lacking read permission or syntax errors in the file).

That said, this points out another behavior change which may be important. `source` will also return a non-zero exit code if the sourced script's final command does so. If that's a problem we can't actually rely on `source` for error-handling after all (though I doubt users _intend_ for sourced scripts to have non-zero exit codes).

libexec/bats-core/bats-exec-test
dimo414: `:?` fails the script if the variable is "null or not set", so this just ensures the user has in fact passed an argument to `load`.

libexec/bats-core/bats-exec-test
dimo414: Do we actually expect any users to notice this message change? 😄 You'd typically only see it immediately before your test `exit`s due to a user error (such as assuming relative paths work...), I'm not sure anyone will miss this message. `source` prints an error when it fails to source a file.

That said, the motivation for removing it is to allow `source` to access files on the `PATH`, which `-f` precludes. I believe [something like](https://stackoverflow.com/a/6569837/113632):

    if [[ ! -e "$file" ]] && ! type -P "$file" >/dev/null; then

would suffice, but I was hoping to simplify the implementation by just deferring error-handling to `source`.

dimo414: It looks like the decision was made for us :) [Bash 3.2 seems unhappy with the `|| exit` syntax](https://travis-ci.org/github/bats-core/bats-core/jobs/678477480) (though I'm not exactly sure why), so I went with the approach mentioned above, which preserves the error message.

libexec/bats-core/bats-exec-test
sublimino: What's the significance of the `name` value in here, a default if `$1` is an empty string?

libexec/bats-core/bats-exec-test
```diff
@@ -40,24 +40,27 @@ fi
...
-  local filename
+  local file="${1:?name}"
 
-  if [[ "${name:0:1}" == '/' ]]; then
-    filename="${name}"
-  else
-    filename="$BATS_TEST_DIRNAME/${name}.bash"
+  if [[ -f "${BATS_TEST_DIRNAME}/${file}" ]]; then
+    file="${BATS_TEST_DIRNAME}/${file}"
+  elif [[ -f "${BATS_TEST_DIRNAME}/${file}.bash" ]]; then
```
sublimino, line 51: I'm wondering now if preserving existing behaviour means searching for the `.bash` version first. In the (perhaps unlikely!) event of both `my_include` and `my_include.bash` existing, would the current implementation default to `my_include` instead of the previous behaviour of `my_include.bash`? Might be easiest-verified with a dedicated test case.

dimo414: I've added a test for the ambiguous behavior, thanks for the suggestion. Given that this change will be held until the 1.3 release how do you feel about keeping the current proposed ordering (exact-match is preferred to `.bash`-suffixed)? If `.bash`-suffixed is preferred `ambiguous` is not reachable, whereas exact-match allows both `ambiguous` and `ambiguous.bash` to be loaded.

If you'd prefer to keep `.bash`-suffix prioritized LMK and I'll flip it and update the test.

---

WRT `${1:?name}`, it is the error that gets reported, and looking at it again I think the default error message is preferable so I've removed `name`. You can experiment with this syntax, try `echo ${foo:?}`, see also [Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html).

libexec/bats-core/bats-exec-test
dimo414: The text after `:?` is the error that gets reported, and looking at it again I think the default error message is preferable so I've removed it. You can experiment with this syntax, try `echo ${foo:?}`, see also [Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html).

libexec/bats-core/bats-exec-test
sublimino: I would prefer the `.bash`-suffixed please, it's an edge case and a minor nit but I worry about breaking any backwards compatibility :pray: Then we should be good to go : ]

dimo414: Sounds good; I left a TODO to consider revisiting this in the future (e.g. if there's a major release).

sublimino:
Thanks for turning this around so quickly @dimo414! :tada:

jasonkarns:
Were docs updated to match new behavior?

sublimino:
@jasonkarns good catch, https://github.com/bats-core/bats-core#load-share-common-code isn't updated.

https://github.com/bats-core/bats-core/blob/master/docs/CHANGELOG.md was updated

PR #286, opened 2020-04-27
Make duplicate test names a hard error

### LINKED ISSUE #278: Fail on duplicate test names
I fell into the trap that I accidentally had same-named tests.
This caused one of tests being run twice and the other not at all.
I consider this rather critical, as it can hide failures and you might wonder what goes on while all tests are green.

I've seen that there is already a detection for duplicate test names and a warning printed at the start of the run.
But if you start the tests, then wait until they are finished and then look at the result, this warning can easily be missed (I did so).
I'd like to suggest to change this to a hard error, as it hides potential test failures by not executing the code that should be executed.

dimo414:

Fixes #278.

### DIFF
--- a/libexec/bats-core/bats-exec-suite
+++ b/libexec/bats-core/bats-exec-suite
@@ -8,6 +8,12 @@ have_gnu_parallel=
 bats_parallel_args=()
 flags=()
 
+abort() {
+  printf 'Error: %s\n' "$1" >&2
+  usage >&2
+  exit 1
+}
+
 while [[ "$#" -ne 0 ]]; do
   case "$1" in
   -c)
@@ -43,7 +49,7 @@ if (type -p parallel &>/dev/null); then
   # shellcheck disable=SC2034
   have_gnu_parallel=1
 elif [[ "$num_jobs" != 1 ]]; then
-  printf 'bats: cannot execute "%s" jobs without GNU parallel\n' "$num_jobs" >&2
+  abort "Cannot execute \"${num_jobs}\" jobs without GNU parallel"
   exit 1
 fi
 
@@ -55,8 +61,7 @@ TESTS_LIST_FILE="${BATS_RUN_TMPDIR}/test_list_file.txt"
 all_tests=()
 for filename in "$@"; do
   if [[ ! -f "$filename" ]]; then
-    printf 'bats: %s does not exist\n' "$filename" >&2
-    exit 1
+    abort "Test file \"${filename}\" does not exist"
   fi
 
   test_names=()
@@ -78,7 +83,7 @@ for filename in "$@"; do
   done < <(BATS_TEST_FILTER="$filter" bats-preprocess "$filename")
 
   if [[ "${#test_dupes[@]}" -ne 0 ]]; then
-    printf 'bats warning: duplicate test name(s) in %s: %s\n' "$filename" "${test_dupes[*]}" >&2
+    abort "Duplicate test name(s) in file \"${filename}\": ${test_dupes[*]}"
   fi
 done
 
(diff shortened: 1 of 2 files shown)


libexec/bats-core/bats-exec-suite
```diff
@@ -43,7 +49,7 @@ if (type -p parallel &>/dev/null); then
   # shellcheck disable=SC2034
   have_gnu_parallel=1
 elif [[ "$num_jobs" != 1 ]]; then
-  printf 'bats: cannot execute "%s" jobs without GNU parallel\n' "$num_jobs" >&2
+  abort "Cannot execute \"${num_jobs}\" jobs without GNU parallel"
   exit 1
```
Vampire, line 53: Is the `exit 1` still necessary here?

libexec/bats-core/bats-exec-suite
```diff
@@ -78,7 +83,7 @@ for filename in "$@"; do
   done < <(BATS_TEST_FILTER="$filter" bats-preprocess "$filename")
 
   if [[ "${#test_dupes[@]}" -ne 0 ]]; then
-    printf 'bats warning: duplicate test name(s) in %s: %s\n' "$filename" "${test_dupes[*]}" >&2
+    abort "Duplicate test name(s) in file \"${filename}\": ${test_dupes[*]}"
```
Vampire, line 86: This will immediately stop execution, right?
Did you consider only failing in the end including the failure message as I suggested in the issue?

jasonkarns: I agree with @vampire. The full test run is still valuable feedback. Marking just the duplicated test as failed is a useful alternative. Though we probably want to distinguish that failure somehow that it's different from a normal test failure.

sublimino: I did — my reasoning was that it's less confusing to error immediately and fail fast, otherwise if all tests pass there's a passing suite but a failing exit code and possibly a missed error.

libexec/bats-core/bats-exec-suite
sublimino: Thanks @Vampire, good spot :pray:

libexec/bats-core/bats-exec-suite
Vampire: That's why I said the error should also be printed last.
I agree that printing the error above all tests and then having a negative exit code is too quiet, especially as during manual invocation you will easily miss that there was a non-0 exit code.

Either @jasonkarns suggestion to actually fail the tests in question would be a good idea, or printing the errors in the end accompanied with the non-0 exit code, like I did here: https://github.com/svn-all-fast-export/svn2git/commit/ece0088e8f88561c45838ab3b4dfb3d21baa9fa3

sublimino: It feels non-obvious to fail after passing test suite, for a check done in pre-processing. Users should rename the test anyway.

I wonder if there's also a risk of nondeterminism here with parallel `--job`s, the "second" duplicate suite may sometimes run as the first. This wouldn't break anything (we've already got a breaking change for the build failing) but if they give different results is a misdirecting user experience. Needs a test.

It seems faster, simpler, and more obvious to fail when the duplicate is detected. That said we can discuss on a PR if preferred :pray:

Vampire: No, it's fine, if it is a reasoned decision rather than a read over of my comment, it's fine.
Especially as it fails at the very beginning.

dimo414: I'm not too bothered by the early failure, because it denotes a mistake in the test setup itself (as sublimino said), but if we really wanted to be more graceful I think a better fix would be to not key off (only) the user-provided test description, and instead ensure that the generated function name _is_ unique.

Vampire: Which would still make it hard to connect failure to test then, I like it failing for duplicate test names. :-)
While then there could be a configuration option on whether to fail or not if there are duplicate names, while either way at least the tests would run correctly.

PR #288, opened 2020-04-27
Reintroduce readlink greadlink usage

waterkip:
In gh #217 the helper to pick readlink or greadlink was removed and soley relies
on readlink. This will break on systems such as MacOS. Reintroducing the helper
and using the result of whatever came out of it.

Signed-off-by: Wesley Schwengle <wesley@opndev.io>

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -49,6 +49,7 @@ HELP_TEXT_BODY
 }
 
 expand_link() {
+  readlink="$(command -v greadlink readlink | head -1)"
   IFS='/' read -ra parts <<<"$1"
 
   # Starting point.
@@ -61,7 +62,7 @@ expand_link() {
     [ -e "$path/$part" ] || return 1
 
     # If it's a link, we need to resolve it first.
-    [ -L "$path/$part" ] && part="$(readlink "$path/$part")"
+    [ -L "$path/$part" ] && part="$("$readlink" "$path/$part")"
 
     # If the component starts with "/" we stomp the existing $path.
     [[ "$part" =~ ^/ ]] && path=""


sublimino:
Thanks @waterkip. There's also code in `bin/bats` that uses readlink, I wonder which approach (`command -v` vs `type -p`) is preferable?

Also can you provide a failing test case for this please? We run MacOS in CI and I don't know why this wasn't caught :(

waterkip:
I would think on just a bit of fiddling on my cli that `command -v` is the better option. `type -p` shows output on STDOUT when it cannot find something:

```
$ type -p foo 2>/dev/null
foo not found
```

```

**Comment by sublimino** [comment]
My digging same to the same conclusion, `command -v` is preferable.

https://unix.stackexchange.com/a/85250/2512

Do you have any idea what a test case could look like? Then we'll fix this ASAP :pray:

**Comment by waterkip** [comment]
ATM I don't have a good idea how to test this.. Do you know a Unix based OS (preferably with a Dockerfile) that uses greadlink and not readlink?

We could opt for finding out what you use and than use that, eg, call it BATS_READLINK and use that everywhere. Although I have the feeling that this expand link is a bit the same as my link resolution fix for starting bats. So maybe unify that to a path resolution scheme for bats.. Is a bit more work tho.

Testcase.. I need a box with greadlink or maybe copy readlink to greadlink and set a custom path so it uses greadlink and not readlink.. but that might be cumberstone with all the other tools found in /usr/bin or elsewhere.

**Comment by sublimino** [comment]
Fair enough, thank you for the detailed response! Let's get this patched then, and I'll open an issue for this so it's visible. Thanks again @waterkip! :100:

**Comment by sublimino** [comment]
@waterkip rebased to GPG sign the commit and switch to `command -v`

**Comment by sublimino** [comment]
I will admit to being a bit confused now, https://github.com/bats-core/bats-core/pull/217/files removes the `readlink -f` dependency, this PR re-introduces `greadlink` (without `-f`).

In the original issue/README, there's the text https://github.com/bats-core/bats-core/pull/214/files

> Darwin's readlink does not include the -f option

So I assume all is well and this follows all the orginal authors' intentions, but without a failing test for either of these PRs (and no Mac to try this on) I'm a little stuck! 

I'll merge this to fix master but expect follow-up issues.

**Comment by waterkip** [comment]
The `readlink -f` removal still requires `readlink` somewhere. So `readlink -f` is gone and a call to just `readlink` still exists.

In the refactor the `$readlink` was removed and replaced with `readlink`, `$readlink` was either `greadlink` or `readlink`.

I'm not sure what the function exactly does, but I saw it skimming over de PR. Note that the manpage of `readlink` says

```
Note realpath(1) is the preferred command to use for canonicalization functionality.

Print value of a symbolic link or canonical file name

-f, --canonicalize
canonicalize  by  following every symlink in every component of the given name recursively; all but the last component must exist
```

So we might even need realpath here.

I have the feeling we are having the same issue here as with #183: Get the location of a path with the existing tools on a system that vary in usage and availability.

sublimino:
@waterkip for the README, is this correct?

> **Mac OSX/Darwin Warning:** If you're executing bats directly (`bin/bats`) 
> you may need to `brew install coreutils` to obtain `greadlink`. Darwin's 
> readlink does not include the -f option. This may be fixed 
> [by this PR](https://github.com/bats-core/bats-core/pull/217), feedback  welcome there.

waterkip:
I think so, yes.

Issue #249, opened 2019-10-22
Line endings in ${lines[@]} not sanitized


thorstenkampe:
Default line endings on different platforms are not stripped - making it hard to interact with native applications on Cygwin.

Consider the following test suite
```
@test 'Cygwin Python' {
    run python3.8 -m this
    [[ ${lines[0]} == 'The Zen of Python, by Tim Peters' ]]
}

@test 'Windows Python' {
    run python -m this
    [[ ${lines[0]} == 'The Zen of Python, by Tim Peters' ]]
}
```
Result:
```
 ✓ Cygwin Python
 ✗ Windows Python
   (in test file test.bats, line 8)
     `[[ ${lines[0]} == 'The Zen of Python, by Tim Peters' ]]' failed

2 tests, 1 failure
```
By the way: unless you already know what the issue is, `bats` gives you no aide for debugging. I suspected `\r\n` line endings from Windows Python. Let's sanitize [explicitly](https://superuser.com/a/1215968)
```
@test 'Windows Python' {
    run python -m this
    [[ ${lines[0]%[[:cntrl:]]} == 'The Zen of Python, by Tim Peters' ]]
}
```
...works...
```
 ✓ Cygwin Python
 ✓ Windows Python

2 tests, 0 failures
```

sublimino:
Hello @thorstenkampe, thanks for the code in this report. I think stripping line endings would be a very wide breaking change to existing test suites. While this is annoying to debug (as are all control characters!) I don't know how we can address this. 

What would be your ideal resolution here, a change to Bats' behaviour under Windows to strip trailing `\r\n`?

thorstenkampe:
@sublimino from what I can see, bats is already stripping `\n` line endings. Otherwise testing...
```
[[ ${lines[0]} == text ]]
```
...could never succeed. In my opinion, Bats should be platform agnostic and strip all kinds of line endings (`\r\n`, `\n`, `\r`). The code for all output lines could be as simple as
```bash
lines=("${lines[@]%[[:cntrl:]]}")
```

rolweber:
Maybe define a flag to switch between current and future behavior?
That should address compatibility concerns.

dimo414:
Can you see if #289 helps? I haven't tried it with `\r\n` but this is yet another example of problems with the existing `run` implementation.

Perhaps it would be worth simply scrapping `run` (leaving it as-is and deprecating it) and introducing a new function with cleaner, more usable semantics.

[...]

PR #158, opened 2018-11-07
Don't glob output

vonbrand:
If the output contains shell metacharacters, like '*', it can get
mangled by shell globbing. Turn it off for test runs.

Signed-off-by: Horst H. von Brand <vonbrand@inf.utfsm.cl>

fixes #156, fixes #281

- [X] I have reviewed the [Contributor Guidelines][contributor].
- [X] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md
[coc]:         https://github.com/bats-core/bats-core/blob/master/docs/CODE_OF_CONDUCT.md

PR #289, opened 2020-04-27
Fix the implementation of the run helper function to properly parse lines of output

### LINKED ISSUE #281: run function is susceptible to globbing
[This line](https://github.com/bats-core/bats-core/blob/118391d/libexec/bats-core/bats-exec-test#L73) is not a safe way to split arbitrary data into lines (as the suppressed [SC2206](https://github.com/koalaman/shellcheck/wiki/SC2206) notes), as it relies on Bash's wordsplitting logic which also expands globs.

MCVE:

```
$ bats --version
Bats 1.1.0

$ ls
bats-bug.bats  some-file

$ cat bats-bug.bats
#!/usr/bin/env bats

@test "bats bug" {
  run printf 'foo\n*\nbar\n'
  printf 'Output:\n%s\n\n' "$output"
  echo "Lines:"
  printf '  %s\n' "${lines[@]}"
  [[ "${lines[1]}" == "*" ]]
}

$ bats bats-bug.bats
 ✗ bats bug
   (in test file bats-bug.bats, line 7)
     `printf '  %s\n' "${lines[@]}"' failed
   Output:
   foo
   *
   bar

   Lines:
     foo
     bats-bug.bats
     some-file
     bar

1 test, 1 failure
```

I can send a PR to use `while read` instead, assuming there's no reason this behavior is intended (e.g. backwards compatibility).

dimo414:
The existing behavior is broken, as it relies on the shell to [word-split unquoted variables](https://github.com/koalaman/shellcheck/wiki/SC2206). This breaks down in several ways, notably that empty lines are silently dropped, but also because words-splitting also triggers globbing, so any line that happens to match a shell glob will be expanded in the resulting array.

**This is a breaking change**, but the existing behavior is fundamentally broken and does not work as documented.

Fixes #281 and #284 

### DIFF
--- a/lib/bats-core/test_functions.bash
+++ b/lib/bats-core/test_functions.bash
@@ -30,17 +30,18 @@ load() {
 }
 
 run() {
-  local origFlags="$-"
-  set +eET
-  local origIFS="$IFS"
   # 'output', 'status', 'lines' are global variables available to tests.
-  # shellcheck disable=SC2034
+  local line origFlags="$-"
+  set +eET
   output="$("$@" 2>&1)"
-  # shellcheck disable=SC2034
   status="$?"
-  # shellcheck disable=SC2034,SC2206
-  IFS=$'\n' lines=($output)
-  IFS="$origIFS"
+  lines=()
+  while IFS= read -r line; do
+    lines+=("$line")
+  done <<<"$output"
+  if [[ -n "$line" ]]; then # if there's any content after the last newline
+    lines+=("$line")
+  fi
   set "-$origFlags"
 }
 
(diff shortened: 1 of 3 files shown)


sublimino:
Thanks @dimo414, give me some time to parse and test this. As it's a breaking change we may want to release 1.4.0 first too, but will get that done sooner rather than later (haven't build a roadmap yet).

dimo414:
Another option, as mentioned in #249, would be to move this behavior to a new function and deprecate `run`. That would avoid the backwards-incompatibility concerns and allow us to create a more robust utility.

martin-schulze-vireso:
Closing this in favor of #158

PR #291, opened 2020-04-27
Rephrase the load docs

dimo414:
Followup to #282

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,18 +351,17 @@ location of the current test file. For example, if you have a Bats test in
 `test/foo.bats`, the command
 
 ```bash
-load test_helper
+load test_helper.bash
 ```
 
 will source the script `test/test_helper.bash` in your test file. This can be
 useful for sharing functions to set up your environment or load fixtures.
+`load` delegates to Bash's `source` command after resolving relative paths.
 
-If you want to source a file using an absolute file path then the file extension
-must be included. For example
-
-```bash
-load /test_helpers/test_helper.bash
-```
+> For backwards compatibility `load` first searches for a file ending in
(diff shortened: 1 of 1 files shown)


sublimino:
@dimo414 thanks! I've added some wording, can you verify it's correct please?

dimo414:
I don't think either of those changes are really necessary to include, but no objections if you feel they're important.

* `PATH` support is a nice-to-have (since it's supported by `source`, and because I suspect it can be used [to improve installing libraries](https://github.com/ztombol/bats-docs/issues/15)), but I don't know that it _really_ needs to be called out here, maybe just say something like "`load` delegates to `source` after resolving relative paths"?
* I intentionally didn't mention the ordering of `.bash` vs. exact-match, because I don't think it should be documented behavior that users rely on. Really, I'd rather see people always just use the exact file name (are the extra five characters really saving anyone time?), but that's a question for another time.

sublimino:
@dimo414 thanks, incorporated your suggestions

README.md
```diff
@@ -351,18 +351,17 @@ location of the current test file. For example, if you have a Bats test in
...
+`load` delegates to Bash's `source` command after resolving relative paths.
 
-If you want to source a file using an absolute file path then the file extension
-must be included. For example
-
-```bash
-load /test_helpers/test_helper.bash
-```
+> For backwards compatibility `load` first searches for a file ending in
+> `.bash` (e.g. `load test_helper`searches for `test_helper.bash` before
```
dimo414, line 362: Add a space before "searches"

dimo414:
LGTM

sublimino:
Thanks for your help @dimo414! :tada:

PR #300, opened 2020-05-01
feat: Add lava support

andrzejtp:
Support lava formatter.

This requires including test end marker in extended syntax output, because lava expects such markers (signals, in lava parlance). This change could perhaps be factored out to a separate commit, consisting of a definition of bats_test_end and its invocation in bats_exit_trap().

Lava assumes that "measurements" can be passed to the output signals (markers), so a -M option is added to support that e.g. in tap output, if the user wants to see the measurements but prefers tap output for human reading when working on the tests. That same option is implied if the user selects
the lava formatter.

The formatter itself is pretty straightforward: it translates begin, end, ok (including skipped) and not ok to lava signals.

The measurements are passed from an individual test to the framework like in the example below, with lava_measurement and lava_unit variables.

Invocation: bats -F lava file.bats

or

bats -M -F tap file.bats

An example file.bats:

#!/usr/bin/env bats

setup()
{
	true
}

@test "Passing testcase" {
	true
}

@test "Passing testcase with measuerement" {
	lava_measurement="abc"
	true
}

@test "Passing testcase with measuerement and units" {
	lava_measurement="123"
	lava_units="kg"
	true
}

@test "Failing testcase" {
	false
}

@test "Skip me" {
	skip
}

Signed-off-by: Andrzej Pietrasiewicz <andrzej.p@collabora.com>

### DIFF
--- a/lib/bats-core/test_functions.bash
+++ b/lib/bats-core/test_functions.bash
@@ -85,6 +85,14 @@ bats_test_begin() {
   setup
 }
 
+bats_test_end() {
+  BATS_TEST_DESCRIPTION="$1"
+
+  if [[ -n "$BATS_EXTENDED_SYNTAX" ]]; then
+    printf 'end %d %s\n' "$BATS_TEST_NUMBER" "$BATS_TEST_DESCRIPTION" >&3
+  fi
+}
+
 bats_test_function() {
   local test_name="$1"
   BATS_TEST_NAMES+=("$test_name")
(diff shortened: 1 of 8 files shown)


martin-schulze-vireso:
Thanks for your work on this. Unfortunately, the current code has moved on and this likely entails a larger rewrite. I put this on the list in #541 to add as an "external" formatter, which won't be included directly with bats-core but maybe gets supported nevertheless. However, for now we have to delay this. I am closing this to avoid clogging up the list with stale PRs.

Issue #290, opened 2020-04-27
odd failure when using --jobs


haircommander:
hey folks

We are running bats in a debian container in circle CI. We've run into this problem that started this morning:
`cat: /tmp/bats-run-28189/parallel_output/1/stdout: No such file or directory`
on any bats job that runs using parallel.

Has anyone seen this error/have any idea what to do to mitigate it?

haircommander:
confirmed with bats version v1.2.0 btw

sublimino:
@haircommander thanks for the report, if I understand correctly this is only on `v1.2.0` as it started this morning? What's the docker invocation, `docker run -it ...` ?

haircommander:
hard to tell exactly. We *were* building off of master, then it started failing. We attempted to switch to v1.2.0, and it continued to fail. A fun addition is that on other branches of CI, this problem does not occur :upside_down_face: . 

Unfortunately, the set up is pretty specific to our circle instance. I couldn't find a simple reproducer starting from the container and a `docker run` command. I can, however, get someone access to a shell where the problem is occurring if they want to poke around.

GabyCT:
I am seeing the same error with the latest version

saschagrunert:
Might the issue be somewhere between those changes, since we see it since last Friday (April 24th)?
https://github.com/bats-core/bats-core/compare/433db67f0081de7f0924ca74fe4afced9df5cdfb...master

sublimino:
1. Can anybody who is building from master try `v1.2.0` please? It was released before the latest set of merges https://github.com/bats-core/bats-core/commit/73b8d2f95513207b319efe34685553b75c0b214e

1. Can anybody provide a reproduction case here please? @haircommander @GabyCT is there any way you can share the bats CLI invocation and any env/docker you are using?

1. If anybody has the time, running `git bisect` should find the source of the error ([simple guide](https://dockyard.com/blog/2016/07/26/how-to-bisect-ember))

1. > Might the issue be somewhere between [433db67...master](https://github.com/bats-core/bats-core/compare/433db67f0081de7f0924ca74fe4afced9df5cdfb...master), since we see it since last Friday (April 24th)?

    It might be, but I can't see anything obvious in there. A larger search ([1.2.0 to now](https://github.com/bats-core/bats-core/compare/73b8d2f95513207b319efe34685553b75c0b214e..master?diff=unified)) will take me a bit longer to work though.

[...]

PR #302, opened 2020-05-02
Fix parallel invocation for debian

martin-schulze-vireso:
parallel's --semaphore mode does not seem to mesh well with --results.
We emulate its behavior by creating the output ourselves.

This is intended to replace #298 

### DIFF
--- a/libexec/bats-core/bats-exec-file
+++ b/libexec/bats-core/bats-exec-file
@@ -167,7 +167,7 @@ bats_run_tests() {
         # use --results to gather output in separate files
         # use --fg mode to retrieve exit codes (all non zero exit codes seem to mapped to 1), but which requires us to suppress its output
         # shellcheck disable=SC2154 #bats_parallel_args is inherited from exec-suite
-        parallel --semaphore -j "$num_jobs" "${bats_parallel_args[@]}" --fg --results "$test_output_dir/{}/" -- bats-exec-test "${flags[@]}" "$filename" "$test_name" "$test_number" >>"$parallel_log_file" 2>&1 &
+        parallel --semaphore -j "$num_jobs" "${bats_parallel_args[@]}" --fg -- bats-exec-test "${flags[@]}" "$filename" "$test_name" "$test_number" >"$test_output_dir/stdout" 2>"$test_output_dir/stderr" &
         pid=$!
         test_pids_and_numbers+=("$pid,$test_number")
       fi
--- a/test/parallell.bats
+++ b/test/parallell.bats
@@ -37,7 +37,8 @@ setup() {
 
 @test "parallel suite execution with --jobs" {
   SECONDS=0
-  run bats --jobs 40 "$FIXTURE_ROOT/suite/"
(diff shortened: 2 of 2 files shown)


martin-schulze-vireso:
Travis seems to be under too much load to succesfully run the strict timing tests.

sublimino:
@martin-schulze-vireso I see the same problem — this works for me locally, but the errors on the build apparently can't be suppressed

```
parallel: Warning: ...
```

Does this mean we have to `grep -v` them out of the parallel output? :/ Is it clear to you why this wasn't a problem with the intial `parallel` patchset? Would like to get this fix out with the `readlink` changes in `1.2.1` at some point soon-ish : ]

martin-schulze-vireso:
> Is it clear to you why this wasn't a problem with the intial parallel patchset?

The original code used `--results` to output stdout and stderr to files. This presumably separates the stderr of the task from parallels own stderr which solves this problem.

However, `--results` uses the parallel arguments (those after ::: ) to label the path. E.g.: `parallel ::: 1 2 ::: 3 4` would generate something like:
```
1/
  3/
    stdout
    stderr
  4/
    stdout
    stderr
2/
  3/
    stdout
    stderr
  4/
    stdout
    stderr
```

However, in semaphore mode we don't have those arguments so I assume that the debian version of parallel does some reads out of bounds that lead to the garbled output.

Then again this is a very specific problem: We overload the machine with too many parallel jobs and this output only makes bats fail because we check for exact output lines. I am not sure if it is a "normal" usecase to check the output of `bats --jobs` runs.

sublimino:
@martin-schulze-vireso I've added a `grep` to the output stream to test CI in this PR. Not my finest hour (feels so cludgy) but if it works to get this feature fixed up I'm not so bothered. 

That said, happy to modify if we can find a better solution.

(I've also rebased your commit to sign it, if you are able to GPG sign I won't do this in future, but I don't mind doing it if not)

martin-schulze-vireso:
@sublimino I'd advise to just do the warning stripping inside the test output and leave the core unchanged. If someone really has a problem with the warnings, they should tell us here. As I said above, the warnings only become errors because we test for `bats --jobs`' output line by line. Aside from internal tests I don't see a usecase for this.

sublimino:
Ah apologies @martin-schulze-vireso I misunderstood that.

While it instinctively feels nasty to have this output in the test run, I agree the previous hack wasn't nice. I've reverted the previous and patched it in to the tests as you suggested, is this what you had in mind?

martin-schulze-vireso:
Yes, I am still mulling over a real fix but I haven't found the time to try that yet. Just in case you want to try: Maybe we can give the semaphore mode some dummy parameters after `:::` so the path becomes more reliable.

sublimino:
Thanks @martin-schulze-vireso, I'm on the path to releasing `1.2.1` but can loop back afterwards.

It reproduces reliably in a Debian container FWIW, I'll put this example in CI at some point https://github.com/bats-core/bats-core/issues/290#issuecomment-621292715

martin-schulze-vireso:
Just FYI: The "real" fix I proposed above does not work. We might use a separate pipe for direct communication with the parallelized process but I don't think that this would gain much over the current solution.

PR #309, opened 2020-05-11
Add @include directive

jccolson:
The files included aren't sourced, so it's possible to have tests in the included files.

Samples :
@test "addition using bc" {
result="$(echo 2+2 | bc)"
[ "$result" -eq 4 ]
}
@include "Tests/*.bats"

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -93,6 +93,7 @@ expand_path() {
 BATS_LIBEXEC="$(dirname "$(expand_link "${BASH_SOURCE[0]}")")"
 export BATS_CWD="$PWD"
 export BATS_TEST_PATTERN="^[[:blank:]]*@test[[:blank:]]+(.*[^[:blank:]])[[:blank:]]+\{(.*)\$"
+export BATS_TEST_INCLUDE="^[[:blank:]]*@include[[:blank:]]+(.*)\$"
 export BATS_TEST_FILTER=
 export PATH="$BATS_LIBEXEC:$PATH"
 export BATS_ROOT_PID=$$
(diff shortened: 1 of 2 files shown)


martin-schulze-vireso:
While the code seems simple, its ramifications might not be. I fear that there might be many unintended sideeffects of including in this fashion. To name a few:

1. Which {setup,teardown}[_file] method will be run if the includer and the includee both provide one?
2. Where (line and filename?) will errors be reported from the includer and the includee?
3. How do we prevent include loops?
4. Can includes lead to unparsable files when both original files were okay? (Even without assuming some strange include places like within functions)

martin-schulze-vireso:
To answer some of the questions above:

> Which {setup,teardown}[_file] method will be run if the includer and the includee both provide one?

The last one that was defined. However, that might not be easy to discern because they might be hidden in a long include chain. However, we already have that problem when `source`ing or `load`ing helpers, so I wont hold that against this PR.

> Where (line and filename?) will errors be reported from the includer and the includee?

As it stands, the resulting preprocessed file will have all includes baked in. This means the line numbers after the first (non empty) include are wrong and the associated file will also be wrong for all included files, when reporting errors. This is a deal breaker to me.

> How do we prevent include loops?

We might not need to. `source` does not care either. However, it might not be obvious to users what the issue is. I think if we would implement this, we should keep track to detect these loops or limit the recursion depth.

> Can includes lead to unparsable files when both original files were okay?

I think the obvious problems arise when using include within functions or similar (multiline) scopes.


All in all I still do not buy into this solution for the given use case and see the implementation as not sufficient. @jccolson I still invite you to discuss the use case in more detail, so we can see if there is a more fitting mechanic than includes to solve the problem you are facing.

PR #312, opened 2020-05-13
Fallback to posix version when (g)readlink -f is not available

### LINKED ISSUE #307: realpath not available on MacOS
Hi,

since https://github.com/bats-core/bats-core/commit/433db67f0081de7f0924ca74fe4afced9df5cdfb bats-core is using the command realpath which is not available on MacOS.

While [travis](https://travis-ci.org/github/bats-core/bats-core/jobs/682808214) reports OK for MacOS running bats inside github actions on MacOS reports some "not found" error. I have the same behavior on my local machine, too.

But all test running fine: https://github.com/jkroepke/helm-secrets/runs/639356455#step:8:6

Exception: If bats-core available via git but not installed via `install.sh`, tests are not running:

```
% ./tests/bats/core/bin/bats -r tests/unit 
./tests/bats/core/bin/bats: line 34: realpath: command not found
./tests/bats/core/bin/bats: line 41: realpath: command not found
env: ./libexec/bats-core/bats: No such file or directory

% ls tests/bats/core/libexec/bats-core/bats
tests/bats/core/libexec/bats-core/bats
```

ko1nksm:
Fallback to [posix version](https://github.com/ko1nksm/readlinkf) instead of using `realpath` when `(g)readlink -f` is not available. Use `(g) readlink -f` when it is available, which gives slightly better performance than before.

Fixes #307

### DIFF
--- a/bin/bats
+++ b/bin/bats
@@ -2,68 +2,58 @@
 
 set -euo pipefail
 
-BATS_READLINK='true'
-BATS_REALPATH=''
-if command -v 'greadlink' >/dev/null; then
-  BATS_READLINK='greadlink'
-elif command -v 'readlink' >/dev/null; then
-  BATS_READLINK='readlink'
+if command -v greadlink >/dev/null; then
+  bats_readlinkf() {
+    greadlink -f "$1"
+  }
+else
+  bats_readlinkf() {
+    readlink -f "$1"
+  }
 fi
 
-if ! "${BATS_READLINK}" -m "$0" &>/dev/null; then
-  BATS_REALPATH='realpath'
+fallback_to_readlinkf_posix() {
+  bats_readlinkf() {
+    [ "${1:-}" ] || return 1
+    max_symlinks=40
+    CDPATH='' # to avoid changing to an unexpected directory
+
+    target=$1
+    [ -e "${target%/}" ] || target=${1%"${1##*[!/]}"} # trim trailing slashes
+    [ -d "${target:-/}" ] && target="$target/"
+
+    cd -P . 2>/dev/null || return 1
+    while [ "$max_symlinks" -ge 0 ] && max_symlinks=$((max_symlinks - 1)); do
+      if [ ! "$target" = "${target%/*}" ]; then
+        case $target in
+          /*) cd -P "${target%/*}/" 2>/dev/null || break ;;
+          *) cd -P "./${target%/*}" 2>/dev/null || break ;;
+        esac
+        target=${target##*/}
+      fi
+
+      if [ ! -L "$target" ]; then
+        target="${PWD%/}${target:+/}${target}"
+        printf '%s\n' "${target:-/}"
+        return 0
+      fi
+
+      # `ls -dl` format: "%s %u %s %s %u %s %s -> %s\n",
+      #   <file mode>, <number of links>, <owner name>, <group name>,
+      #   <size>, <date and time>, <pathname of link>, <contents of link>
+      # https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ls.html
+      link=$(ls -dl -- "$target" 2>/dev/null) || break
+      target=${link#*" $target -> "}
+    done
+    return 1
+  }
+}
+
+if ! BATS_PATH=$(bats_readlinkf "${BASH_SOURCE[0]}" 2>/dev/null); then
+  fallback_to_readlinkf_posix
+  BATS_PATH=$(bats_readlinkf "${BASH_SOURCE[0]}")
 fi
 
-resolve_path() {
-  local path_to_bats=$1
-  local target_path
-
-  if [[ -L "${path_to_bats}" ]]; then
-    # Busybox has no readlink from coreutils
-    if [[ -n "$BATS_REALPATH" ]]; then
-      target_path="$("$BATS_REALPATH" "$path_to_bats")"
-      # we need to traverse the paths
-      target_path="$(dirname "$target_path")" # strip bats
-      target_path="$(dirname "$target_path")" # strip bin
-    else
-      target_path="$("$BATS_READLINK" -m "$path_to_bats/../../")"
-    fi
-  elif [[ -f "${path_to_bats}" ]]; then
-    target_path=$(dirname "$path_to_bats")
-    if [[ -n "$BATS_REALPATH" ]]; then
-      target_path="$("$BATS_REALPATH" "$target_path")"
-    else
-      target_path="$("$BATS_READLINK" -m "$target_path")"
-    fi
-    target_path="$(dirname "$target_path")" # strip bin
-  elif [[ -d "${path_to_bats}" ]]; then
-    if [[ -n "$BATS_REALPATH" ]]; then
-      target_path="$("$BATS_REALPATH" "$path_to_bats")"
-      target_path="$(dirname "$target_path")"
-    else
-      target_path="$("$BATS_READLINK" -m "$path_to_bats/../")"
-    fi
-  fi
-
-  echo "$target_path"
-}
-
-bats_resolve_absolute_root_dir() {
-  local path_to_bats="$1"
-  local result="$2"
-  local target_path
-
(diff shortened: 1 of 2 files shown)


sublimino:
Thank you @ko1nksm :pray: 

> In order not to lengthen YOUR script, I write it code-golf-like style. Should I change to a readable style?

Yes please, that would be deeply appreciated. Do you know if the code-golf increases performance in your library?

> This code is originally written for a POSIX compliant shell, so it contains workarounds for other shells (ksh, etc.). Should I remove unnecessary code and dedicate it to bash only?

That's a great question. I think it will be easier to maintain if it doesn't diverge from your codebase, at the cost of some obviousness to a reader of the code who may wonder why `ksh` fixes are in `bats` :upside_down_face: 
- Do you intend to make upstream long-form (e.g. non-code-golf), or just this PR?  If this code diverges for that reason perhaps it does makes sense to remove `ksh` fixes too. 
- Does anybody else have an opinion here?

> It changes global variable p (local is not POSIX compliant). There is no problem if the function is called by "command substitution" (subshell). Should I change it to the local variable?

We try to avoid subshells for performance reasons if possible, a `local` would be best, but depends on the above questions (i.e. the level of change from your upstream repo you anticipate). If we're trying to prevent divergence we could leave this as a global possibly. 

You've turned this around so quickly, thank you for your efforts! :tada:

ko1nksm:
Thank you reply @sublimino

I will rewrite it non-code-golf style and remove workaround for other shells.

> Do you know if the code-golf increases performance in your library?

The code-golf is meant to shorten the code, not performance. It's neither fast nor slow.

Compatibility with `readlink -f` is often a problem in shell scripts world. I did research before implementing this, but I couldn't find any reliable code.

I want to get rid of this problem entirely in shell scripts world, so I'll add code upstream to make it widely used. I will rewrite it based the current code, but it is a new code and needs testing. Please give me a few days.

ko1nksm:
I refactored upstream [readlinkf](https://github.com/ko1nksm/readlinkf) and rebased this commit. This is still for POSIX shell. I will happily change it to bats coding style if you want.

Note:

- The `bats_readlinkf` is intended for use via command substitution (or subshell). This is because it to match the interface with the `(g)readlink` version and preserve the current directory and variables that are implicitly changed like `PWD` and `OLDPWD`, not only variables used.
- Export the `bats_readlinkf` function to not define the same function twice. I believe `libexec/bats-core/bats` does not run directly.
- When run `bash bats`, `$0` is "bash" not absolute path. Therefore I changed it to `${BASH_SOURCE[0]}`.

bin/bats
```diff
@@ -2,68 +2,53 @@
...
-    else
-      target_path="$("$BATS_READLINK" -m "$path_to_bats/../")"
-    fi
-  fi
-
-  echo "$target_path"
+fallback_to_readlinkf_posix() {
+  bats_readlinkf() {
+    [ "${1:-}" ] || return 1
+    target=$1 max_symlinks=10 CDPATH=''
```
sublimino, line 18: Each declaration on its own line please

bin/bats
```diff
@@ -2,68 +2,53 @@
...
+
+    [ -e "${target%/}" ] || target=${1%"${1##*[!/]}"} # trim trailing slashes
```
sublimino, line 20: My instinct suggests this should be quoted, I guess there's a reason it's not?

bin/bats
```diff
@@ -2,68 +2,53 @@
...
+    [ -d "${target:-/}" ] && target="$target/"
+
+    cd -P . 2>/dev/null || return 1
+    while [ "$max_symlinks" -gt 0 ] && max_symlinks=$((max_symlinks - 1)); do
+      if [ ! "$target" = "${target%/*}" ]; then
+        cd -P "${target%/*}/" 2>/dev/null || break
```
sublimino, line 26: I wonder if this should be `cd -P -- ...` in case of directory names like `-Lblah`

bin/bats
```diff
@@ -2,68 +2,53 @@
...
+        target=${target##*/}
```
sublimino, line 27: Do these direct-to-variable assignments need quoting? We tend to quote by default out of paranoia!

bin/bats
```diff
@@ -2,68 +2,53 @@
...
+        target="${PWD%/}${target:+/}$target"
+        printf '%s\n' "${target:-/}"
+        return 0
+      fi
+
+      # `ls -dl` format: "%s %u %s %s %u %s %s -> %s\n",
+      #   <file mode>, <number of links>, <owner name>, <group name>,
+      #   <size>, <date and time>, <pathname of link>, <contents of link>
+      # https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ls.html
+      link=$(ls -dl "$target" 2>/dev/null) || break
```
sublimino, line 40: Perhaps again `ls -dl -- ...`?

bin/bats
```diff
@@ -2,68 +2,53 @@
...
+      target=${link#*" $target -> "}
```
sublimino, line 41: Another quote query

bin/bats
ko1nksm: OK, Fixed.

bin/bats
ko1nksm: Yes! Thank you! However, in very old zsh, `cd` could not accept `--`. Another way I implemented is prepend `./` except it starts with `/`.

bin/bats
ko1nksm: Use `--` this time.

ko1nksm:
**About quoting**

My coding style is "Do not quote in simple assignment". Quoting isn't necessary here. And another reason, This is workarounds for some shells.

The posh has many bugs even the latest version (0.13.2).

```sh
a='foobar' b='bar'
c=${a%"$b"*} # => c is "foo" (correct)
d="${a%"$b"*}" # => d is "foobar" (incorrect)
```

The bosh has bug expect the latest version (2020-05-11).

```sh
x="a*b"
y="a[*]b"
echo ${x#"$y"} # => "a*b" (correct)
echo "${x#"$y"}" # => empty (incorrect)
```

You can only change this PR if you wish. In that case, separate a commit that you can easily see the differences.

> do you think there's an advantage to using native readlink/greadlink

Depends on symbolic link depth and environment. There is no difference in the current expected use of bats, IMO. 

CPU Core i7 3.4Gzh, PCI express SSD.

| on Linux                             | real     | user     | sys      |
| ------------------------------------ | -------- | -------- | -------- |
| `readlink -f`, symlink depth: 40     | 0m0.002s | 0m0.000s | 0m0.002s |
| `readlinkf_posix`, symlink depth: 3  | 0m0.007s | 0m0.004s | 0m0.003s |
| `readlinkf_posix`, symlink depth: 40 | 0m0.096s | 0m0.069s | 0m0.031s |

| on WSL1                              | real     | user     | sys      |
| ------------------------------------ | -------- | -------- | -------- |
| `readlink -f`, symlink depth: 3      | 0m0.009s | 0m0.000s | 0m0.016s |
| `readlink -f`, symlink depth: 40     | 0m0.018s | 0m0.000s | 0m0.031s |
| `readlinkf_posix`, symlink depth: 3  | 0m0.037s | 0m0.004s | 0m0.016s |
| `readlinkf_posix`, symlink depth: 40 | 0m0.476s | 0m0.016s | 0m0.328s |

And, in rare cases, GNU `readlink`, Busybox `readlink` and FreeBSD `readlink` may have [some different behaviors](https://github.com/ko1nksm/readlinkf/blob/master/NOTE.md).

It's okay to use bats, but if you want full compatibility then fallback should be remove.

**Other changes**

Increased max symlinks, 10 to 40. it is a same value as [MAXSYMLINKS][MAXSYMLINKS] of linux kernel.

[MAXSYMLINKS]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/namei.h?h=v5.6&id=7111951b8d4973bda27ff663f2cf18b663d15b48#n13

**Comments**

I added comments to the code, but I am not a native English speaker. Please correct it if you found weird my English.

**Note**

Currently the upstream CI is failing, but this seems to be a problem with the msys2 package repository.

PR #335, opened 2020-07-25
Improve performance: Don't preprocess per test

martin-schulze-vireso:
This should be done only once per file.

I noticed that each test will generate its own preprocessed file. This seems to have been broken a while back already since there is a comment that explicitly says we want to cache the result.

With this change the bats test suite ran 20% faster on my machine.

I'd like to have a regression test in place for this but so far I have no idea how to achieve this in a sensible manner.

### DIFF
--- a/lib/bats-core/preprocessing.bash
+++ b/lib/bats-core/preprocessing.bash
@@ -12,7 +12,9 @@ BATS_PARENT_TMPNAME="$BATS_RUN_TMPDIR/bats.$PPID"
 BATS_OUT="${BATS_TMPNAME}.out" # used in bats-exec-file
 
 bats_preprocess_source() {
-	BATS_TEST_SOURCE="${BATS_TMPNAME}.src"
+	# export to make it visible to bats_evaluate_preprocessed_source
+	# since the latter runs in bats-exec-test's bash while this runs in bats-exec-file's
+	export BATS_TEST_SOURCE="${BATS_TMPNAME}.src"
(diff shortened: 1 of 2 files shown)


sublimino:
I have been doing rudimentary benchmarking with https://github.com/sharkdp/hyperfine and Google Cloud Build but don't have a regression test suite yet, just using

```bash
for ((i = 0; i < 200; i++)); do
  cat <<EOF@test "a passing test #${i}" {
  true
}
EOF
done | tee test/fixtures/bats/regression.bats >/dev/null

hyperfine 'bin/bats test/fixtures/bats/regression.bats'
```

And mixing in `--jobs`. Not particularly empirical yet!

This makes a noticeable difference using a single file with 200 no-op tests:

```
# this version
hyperfine 'bin/bats test/fixtures/bats/regression.bats'                                                                                       
Benchmark #1: bin/bats test/fixtures/bats/regression.bats                                                       
  Time (mean ± σ):      1.468 s ±  0.011 s    [User: 1.367 s, System: 0.154 s]                                  
  Range (min … max):    1.453 s …  1.483 s    10 runs                                 

# previous commit
hyperfine 'bin/bats test/fixtures/bats/regression.bats'                                                                                      
Benchmark #1: bin/bats test/fixtures/bats/regression.bats                                                       
  Time (mean ± σ):     18.669 s ±  0.194 s    [User: 18.027 s, System: 0.703 s]                                 
  Range (min … max):   18.379 s … 19.006 s    10 runs

Issue #311, opened 2020-05-12
ANSI coded output causes Junit formatter to create invalid XML


bioball:
If outputting text with ANSI controls, the Junit formatter will create invalid XML. For instance:

```bats
@test "my-test" {
  printf "\e[1mhello\e[0m"
  return 1
}
```

Produces this output (with properties redacted)

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="my_test.bats" tests="1" failures="1" errors="0" skipped="0" time="0" timestamp="2020-05-12T15:01:25">
    <testcase classname="my_test.bats" name="my-test" time="0">
        <failure type="failure"><![CDATA[
(in test file test/my_test.bats, line 106)
  `printf "\e[1mhello\e[0m"' failed
[1mhello[0m]]></failure>
    </testcase>
  <system-out></system-out>
  <system-err></system-err>
</testsuite>
```

The CDATA section includes the control character literal `` (Unicode: 0x1b), which are invalid XML characters. This is causing some XML parsers to fail parsing the resulting file.

I think maybe there's two ways to handle this:
1. Encode the escape character (as `\e` or `\033` or something)
2. Strip all control characters from the captured output

sublimino:
Thanks for the report @bioball, and for kicking the tyres on this one : ]

cc @martin-schulze-vireso WDYT? Option 1 may add complexity when escaping for different targets, or might be as simple as a one-time escape before writing CDATA.

If the second option, I've been using a sed control-character replacer for years that may be suitable:

```
colour-strip () { 
    sed "s,\x1B\[[0-9;]*[a-zA-Z],,g"
}
```

martin-schulze-vireso:
>  Option 1 may add complexity when escaping for different targets

What do you mean by different targets? My understanding is that the XML definition clearly states which symbols are allowed, so there really should be only "one rule" for that.

I think stripping them is good enough until someone complains about missing them.

However, this issue raises the question: What if someone does `printf "]]>"`? IMHO this hints at a more general requirement to properly sanitize/escape the output for the CDATA area.

sublimino:
> What do you mean by different targets? 
Viewed on different operating systems and terminal types. Nothing specific, just general caution!

> I think stripping them is good enough until someone complains about missing them.
Agreed

> However, this issue raises the question: What if someone does printf "]]>"? IMHO this hints at a more general requirement to properly sanitize/escape the output for the CDATA area.

That's a good point, I don't know enough about XML to help here.

[...]

Issue #342, opened 2020-08-14
pretty output with junit formatter


jeff-cook:
I plan on running bats inside a CI pipeline. It would be great to be able to run bats with both the pretty console output while also saving the junit report. 

I tried and only got the junit output, or the pretty output.

```
bats --recursive --pretty --formatter junit .
bats --recursive --formatter junit --pretty .
```

Is there anyway to be able to get both without running the command twice?

martin-schulze-vireso:
Unfortunately, no. They are both formatters and there can only be one formatter. However, during implementation the discussion already came up, whether junit should be considered a formatter at all.

jeff-cook:
I could see not having junit being a formatter and have it a new `--report` output. That would be the best of both worlds to me.

martin-schulze-vireso:
> I could see not having junit being a formatter and have it a new --report output.

I was thinking along the same lines. @sublimino What do you think about this?

martin-schulze-vireso:
After having thought about this for a while I want to get into a more detailed discussion. The basic question is: What is the purpose of `--formatter` when we have `--report`?

Until the JUnit formatter, we only had the output on stdout in formats pretty and tap. Where tap is basically the internal output stream. Additionally, we have the lava (#300), CppUnit (#54) , and tap13 (#337) formats in PRs.

As far as I can see, bats has two basic usage modes: Interactive, when developing locally in a terminal and batch when started within a CI script.

The consumption of bats output is also dependent on this: In interactive mode, the developer looks at the stdout and might sometimes use the return code. In batch mode, the return code is probably the most important feature, if the CI pipeline does not support report parsing, the user will look at the stdout. If report parsing is supported, then the report will be reprocessed for human consumption.

The current `--formatter` option sets the format of stdout, the `--report` option sets the format of the report file. I see the following matrix:

  Format  | `--formatter` | `--report`
------------ | -------------------- | --------------------------
TAP[13] | reduced readability | yes 
pretty | yes | no
junit | no | yes
CppUnit | no | yes
lava | no | yes

From this table I gather that we really need just one stdout format, which is `pretty` and all other formats are more or less for report mode. The only exception is TAP, which is still human readable enough. We might need to provide at least TAP for backwards compatibility and some tests might become easier if we just can work with the format on stdout but other than that I don't see compelling reasons for the other formatters to ever be printed on stdout.

What do you think about all this @sublimino ? Should we just default to pretty output (maybe with a non-interactive version for CI logs)?

[...]

Issue #341, opened 2020-08-14
recursive with junit formatter


jeff-cook:
When running recursive with junit formatter there are name conflicts with the reports that do not exist in other formats.

```
bats --recursive --formatter junit .
```

The report files are all created in one directory. 
The files are named based on the bats file name, however ignore the path to the bats file.
If there is 2 `test.bats` files in different directories there will only be one `TestReport-test.bats.xml` reports file.
Also within the report file the path is not used in the class name.

martin-schulze-vireso:
I'd like to separate this from the recursive mode. The problem comes from duplicate filenames and can be triggered via

```
bats --formatter junit subfolder1/test.bats subfolder2/test.bats
```

as well.

When using one output file per input file, we would have to either put the whole path into the resulting file name or mirror the input folder structure from a given anchor, to avoid the name clashes between duplicate file names in different folders.

I had a look again at the JUnit xml schemes and noticed that the current output somewhat deviates. It uses `<testsuite>` as root node whereas all XSDs I found would put `<testsuites>` as root with mutliple `<testsuite>`s as children. If this format is compatible with the desired targets, I would propose to switch to a single file output, which would eliminate the problem.

jeff-cook:
I would be good with creating a single xml file, or putting each xml file in the dir with the .bats file creating it.

Anything I can do to help?

martin-schulze-vireso:
I think the single xml file is the easiest to implement.

> Anything I can do to help?

You could venture your own PR for this or help on review/test a PR that was done by someone else.

PR #337, opened 2020-07-26
TAP13 compliant timing

martin-schulze-vireso:
This is a loos port of #127 to the current code layout.

Features:
- measuring tests in ms (instead of (integer!) seconds as before)
- adding a TAP13 formatter to avoid having to maintain multiple formats in the internals

- [x] I have reviewed the [Contributor Guidelines][contributor].
- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md
[coc]:         https://github.com/bats-core/bats-core/blob/master/docs/CODE_OF_CONDUCT.md

PR #345, opened 2020-08-21
Allow for output format to differ between stdout and report file

martin-schulze-vireso:
This mainly deals with #342 but might incorporate #341  as well.

- [x] document --report option
- [x] update to #337's timing information, as soon as it lands

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ Usage: bats [OPTIONS] <tests>
   --no-parallelize-within-files
                             Serialize test execution within files instead of
                             running them in parallel (requires --jobs >1)
+  --report-formatter <type> Switch between reporters (same options as --formatter)
   -o, --output <dir>        Directory to write report files
   -p, --pretty              Shorthand for "--formatter pretty"
   -r, --recursive           Include tests in subdirectories
--- a/lib/bats-core/formatter.bash
+++ b/lib/bats-core/formatter.bash
@@ -23,7 +23,7 @@ function bats_parse_internal_extended_tap() {
     fi
 
     ok_line_regexpr="ok ([0-9]+) (.*)"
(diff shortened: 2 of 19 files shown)


martin-schulze-vireso:
This should also fix #311 now. While working on the escaping of the CDATA sections I noticed the problem is more widespread. We also need to escape those parts where file names, test names, or skip messages are inserted, as they might contain characters that break the xml.

Hence, I opted for using xml escape sequences like `&lt;` instead of the CDATA only solution.

sublimino:
I've added a commit, this looks great.

One question is flag names, currently `--formatter` and `--report`. Should the latter be `--reporter`?

I also think the `-help` text has got messy, I'll get round to cleaning that up.

martin-schulze-vireso:
Yes, I also thought about that but reporter is a loaded word. If we are okay with a long name I would opt for `--report-format[ter]`.

martin-schulze-vireso:
I just noticed that there is a problem when we want to use a report formatter that uses the internal extended syntax and the tap formatter which simply forwards the input. We can either forbid this combination or we have to implement a filter function on the tap formatter or we live with the invalid tap output on stdout.

martin-schulze-vireso:
I made the TAP formatter compatible with extended syntax. For this I added `lib/bats-core/formatter.bash` to centralize the parsing logic and use callbacks for the individual formatters. I only implemented this for tap and pretty, we should move the others over to this interface as well.

In the long run this offers the opportunity to check if the formatters implement the interface correctly by at least having a definition of these callbacks, which might pave the way for a stable interface to move formatters out of bats-core.

sublimino:
Thanks, LMK if this ready to review and merge 🙏

martin-schulze-vireso:
@sublimino You can review now.

sublimino (approved):
Looks awesome, thank you @martin-schulze-vireso :100::tada:

PR #348, opened 2020-09-11
Small cleanups

martin-schulze-vireso:
- [x] I have reviewed the [Contributor Guidelines][contributor].
- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md
[coc]:         https://github.com/bats-core/bats-core/blob/master/docs/CODE_OF_CONDUCT.md

PR #349, opened 2020-09-11
dynamic test registration

### LINKED ISSUE #306: [wishlist] dynamic test registration
It would be nice to be able to register test functions at runtime, in addition to the `@test` preprocessor behavior. Something like:

```shell
bats::register_test is_prime 2 true
bats::register_test is_prime 3 true
bats::register_test is_prime 4 false

is_prime() { .... }
```

#241 is closely related; it's possible implementing that would address this need. Alternatively, implementing this could be sufficient for users to construct their own parameterization on top. I'm not sure we need both features or not, but I can see them both being useful.

martin-schulze-vireso:
I played a bit around with the framework to see if we could get parametrized tests working, so this is more of a proof of concept.
This needs #348 but I wanted to make those separate so we can play around here.

I see the following issues with this PR:

1. it adds another layer of execution to the test files, so we now run them n+2 times
2. with bats-exec-suite `source`ing the test file, there is even more collision surface between tests and the framework, this becomes especially nasty when there are syntax errors in the test files

Closes #306

### DIFF
--- a/libexec/bats-core/bats-gather-tests
+++ b/libexec/bats-core/bats-gather-tests
@@ -1,0 +1,299 @@
+#!/usr/bin/env bash
+set -eET
+
+args="$@"
+filter_tags_list=()
+
+# shellcheck source=lib/bats-core/common.bash disable=SC2153 
+source "$BATS_ROOT/lib/bats-core/common.bash"
+# shellcheck source=lib/bats-core/preprocessing.bash
+source "$BATS_ROOT/lib/bats-core/preprocessing.bash"
+
+abort() {
+  printf 'ERROR: '
+  printf "$@"
+  exit 1
+} >&2
+
+while [[ "$#" -ne 0 ]]; do
+  case "$1" in
+    --dummy-flag)
+      ;;
+    --filter-tags)
+      shift
+      IFS=, read -ra tags <<<"$1" || true
+      if ((${#tags[@]} > 0)); then
+        for ((i = 0; i < ${#tags[@]}; ++i)); do
+          bats_trim "tags[$i]" "${tags[$i]}"
+        done
+        bats_sort sorted_tags "${tags[@]}"
+        IFS=, filter_tags_list+=("${sorted_tags[*]}")
+      else
+        filter_tags_list+=("")
+      fi
+      ;;
+    --)
+      shift 1
+      break
+      ;;
+    *)
+      abort "Unknown flag %s in command:\nbats-gather-tests %s" "$1" "${args[*]}"
+      ;;
+  esac
+  shift 1
+done
+
+
+# shellcheck source=lib/bats-core/test_functions.bash disable=SC2153 
+# required to provide e.g. `load` for free code (some users rely on this)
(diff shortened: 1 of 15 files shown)


lucastheisen:
If i understand the code correctly, this basically allows you to define functions:

```bash
function assert_equals {
  if [[ "$1" != "$2" ]]; then
    echo "[$1] != [$2]"
  fi
}
```

Which can be run at any point later by invoking a another function:

```bash
bats_test_function --description "one is one" assert_equals "one" "one"
bats_test_function --description "two is two" assert_equals "two" "two"
bats_test_function --description "two is three" assert_equals "two" "three"
```

Then you would see a single set of results:

```bash
 ✓ one is one
 ✓ two is two
 ✗ two is three
   (from function `assert_equals' in file tests.bats, line ?)
     `?' failed

3 tests, 1 failure
```

This seems sufficiently dynamic for most use cases.

When you say

> there is even more collision surface between tests and the framework

is that just to say that bats will be less helpful in pointing out where your syntax errors are?  or that things might not actually work because your tests have vars that collide with bats internals?

martin-schulze-vireso:
> When you say
> 
> > there is even more collision surface between tests and the framework
> 
> is that just to say that bats will be less helpful in pointing out where your syntax errors are? or that things might not actually work because your tests have vars that collide with bats internals?

The main problem is: when we hook code into bats, it might mess with internal variables. Additionally, if there is a syntax error in the sourced file, bats will go down with it, potentially swallowing the error (or giving unhelpful filenames due to the preprocessing). I am thinking about jailing this `source` into an extra bash invocation to avoid this.

brokenpip3 (approved):
:muscle:

libexec/bats-core/bats-gather-tests
```diff
@@ -0,0 +1,306 @@
...
+  fi
+
+  test_names=()
+  test_dupes=()
+
+  BATS_TEST_FILENAME="$filename"
+  _bats_test_functions_setup -1 # invalid TEST_NUMBER, as this is not a test
+
+  BATS_TEST_NAME=source
+  BATS_TEST_FILTER="$BATS_TEST_FILTER" bats_preprocess_source # uses BATS_TEST_FILENAME
```
debarshiray, line 284: There doesn't seem to be anything left that's using `BATS_TEST_FILTER`.

debarshiray (commented):
I happened to notice something while reading the code:

debarshiray (commented):
This broke test suites with multiple files loading common constants.  I have tried to put together a possible fix with test cases to illustrate the problem in https://github.com/bats-core/bats-core/pull/904

libexec/bats-core/bats-gather-tests
debarshiray: I tried to remove `BATS_TEST_FILTER` in https://github.com/bats-core/bats-core/pull/907

joeymarianer-calendly:
Just curious, is there any specific reason this feature isn't in https://bats-core.readthedocs.io/en/stable/ (I searched for "dynamic" and "bats_test_function")

PR #363, opened 2020-10-03
fix: time is greater than 0ms for long test

sublimino:
- busybox doesn't support `date +%s%N` syntax
- returned time value judged to be in millis if more than 18 digits long

---


martin-schulze-vireso:
On Bash 5 we could also rely on `$EPOCHREALTIME`. It should be quite easy to transform it to milliseconds since epoch.

martin-schulze-vireso (approved):

Issue #283, opened 2020-04-25
Isolated temp dir per run


dimo414:
The README points to `BATS_TMPDIR` as a "location to a directory that may be used to store temporary files", but currently it's (typically) just [`/tmp` or `$TMPDIR`](https://github.com/bats-core/bats-core/blob/118391d/libexec/bats-core/bats-exec-test#L352), meaning that anything written to that directory will affect subsequent runs.

BATS itself appears to instead use [`BATS_TMPNAME`](https://github.com/bats-core/bats-core/blob/118391d/libexec/bats-core/bats-exec-test#L358) which is keyed off the PID (which still isn't safe, as the OS will reuse PIDs over time), but it's not clear whether users are supposed to use `BATS_TMPNAME` or if it's an implementation detail.

I would suggest providing a different variable, e.g. `BATS_TEST_TMP`, that is unique per-run, and documenting that in the README instead of `BATS_TMPDIR`. I'd also suggest changing `BATS_TMPNAME` to not (only) rely on the PID, but I'm not sure if there are additional constraints with that - [`BATS_PARENT_TMPNAME`](https://github.com/bats-core/bats-core/blob/118391d/libexec/bats-core/bats-exec-test#L359) makes me suspicious parts of BATS currently relies on this naming scheme.

rico-chet:
The problem of handling temporary directories is [solved](https://github.com/tralston/bats-file#working-with-temporary-directories), the question is whether it belongs to the core and the convenient answer is _no_. Which percentage of tests need this functionality? We probably don't know, and a smaller focused library is better than a bigger broader one.

So, I rather support #276.

dimo414:
"solved" is a bit strong given that this requires a separate library that hasn't even been pulled into the bats-core org yet.

It's reasonable to decide that file-system dependent assertions and test utilities should live in a separate library, but bats-core is responsible for test setup and teardown, and for the environment made available to tests. The existing documented behavior is error-prone, and will remain so even if people are *supposed* to use bats-file.

rico-chet:
> Bats is a TAP-compliant testing framework for Bash.

So `bats-core` is concerned about TAP-compliance and the framework side of things, batteries, canvas, paint and brushes not included.

We could debate endlessly. ;-)

dimo414:
Maybe we're talking past each other - are you suggesting the current behavior of `BATS_TMPDIR` (and `BATS_TMPNAME`) is a *feature* and not a bug?

[...]

PR #365, opened 2020-10-09
Improve BATS_RUN_TMPDIR documentation

martin-schulze-vireso:
Improve on the issues discussed in https://github.com/bats-core/bats-core/issues/283#issuecomment-629869040

I noticed that BATS_TMPDIR is unconditionally initialized from TMPDIR, so having set BATS_TMPDIR from outside will be overriden by TMPDIR. I think this might be confusing for new users. My proposal would be:

1. [ ] use BATS_TMPDIR as is, if it is set
2. [ ] use BATS_TMPDIR=$TMPDIR if BATS_TMPDIR is not set and TMPDIR is set
3. [ ] use BATS_TMPDIR=/tmp if neither BATS_TMPDIR nor TMPDIR are set
4. [ ] document this for new users

Hopefully, this does not break existing users as it only changes the case where an already set `BATS_TMPDIR` is (IMHO erroneously) overridden. WDYT @sublimino ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -594,8 +594,10 @@ There are several global variables you can use to introspect on Bats tests:
   test file.
 * `$BATS_SUITE_TEST_NUMBER` is the (1-based) index of the current test case in
  the test suite (over all files).
-* `$BATS_TMPDIR` is the location to a directory that may be used to store
-  temporary files.
+* `$BATS_RUN_TMPDIR` is the path to a directory that will be deleted after the run
+  unless `--no-tmpdir-cleanup` is specified (defaults to `$BATS_TMPDIR/bats-run-<PID>`)
+* `$BATS_TMPDIR` is the path to a directory that may be used to store
+  temporary files. (defaults to `/tmp`)
 
 ### Libraries and Add-ons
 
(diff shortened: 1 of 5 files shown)


rockandska:
Is there any particular reason to use BATS_TMPDIR ?
Why not only relay on TMPDIR ?
This is having BATS_TMPDIR who is the most confusing thing IMO

martin-schulze-vireso:
BATS_TMPDIR has been around for ages. Changing its meaning might break current code.

martin-schulze-vireso:
@rockandska 

Ah, I think I misunderstood what you were getting at. I still believe we have a problem with preexisting code.

I just did some detective work on where BATS_TMPDIR is coming from and why it was added.
Commit e9d3143fdbaa0fb867cf4be7bcb7226a2c50324f added it merely as a local helper. Then it was documented in b1eee9f455f677a9356495aae382d4879a6c5cc3 and now we are stuck with this mess, as there might be actual users.

I am fully with you @rockandska  that we should rely on `TMPDIR` for setting `BATS_TMPDIR`'s value. Additionally, I think new users should not be told about it. It should merely be used as an internal variable. As such, this PR does not make sense.

Issue #358, opened 2020-09-25
Add documentation for beginners


martin-schulze-vireso:
The current documentation of bats does not lend itself to an easy introduction. There are several resources out there that give an introduction (e.g. https://advancedweb.hu/testing-bash-scripts-with-the-bats-testing-framework/, https://medium.com/@pimterry/testing-your-shell-scripts-with-bats-abfca9bdc5b9, https://opensource.com/article/19/2/testing-bash-bats) but they are not official in the sense that they are not hosted here and some of them are quite aged, so they don't reflect the current state of bats anymore.

I think it would be worthwhile to have a short tutorial that introduces the fundamentals of bats and maybe a list of common mistakes and pitfalls that will get especially beginners confused.

To that end, I would propose we gather ideas and common mistakes here.

martin-schulze-vireso:
Some items from the gitter chat:

> https://gitter.im/bats-core/bats-core?at=5aec061e03cafa797b26c1a5
> How exactly do I load the additional BATS libs?

> https://gitter.im/bats-core/bats-core?at=5b2ce9edad21887018d4922c
>  I want to test all possible combination of command line options to a program. The commands can be easily generated using some nested for loops. Can I put a @test inside a bash for loop?
> https://gitter.im/bats-core/bats-core?at=5eb08c845cd4fe50a3ecbb19
> Is there a way to do table driven testing using bats? I see that @test does not like to appear inside a loop.

> https://gitter.im/bats-core/bats-core?at=5b3209c2479ca266898046c4
> how can i pass a parmter to a bats file?

> https://gitter.im/bats-core/bats-core?at=5b495c669a612333aa5f6501
> I'm wondering how can I run code within a .bats file *after* all the tests run, to clean up some things

> https://gitter.im/bats-core/bats-core?at=5b64dea1a9426c705d2343da
> How can I set the current directory from where BATS runs?

> https://gitter.im/bats-core/bats-core?at=5c38e2c76a394e282191abd1
> [...] i'm trying to test that an environment variable is set when a function is called - what is the recommended way to do that? 
> https://gitter.im/bats-core/bats-core?at=5cd38d036fd7c11cd88686be
>  I have a bash library that is sourced and has a function that sets a variable to be accessed from the calling script. How can I test this?
> https://gitter.im/bats-core/bats-core?at=5ec2d688b8869e38c49765c4
[...]

PR #366, opened 2020-10-09
README.md: more examples of (mis)using run

### LINKED ISSUE #377: Improve the run function
The current run functions has many shortcomings and there are several issues and PRs that deal with them. This issue tries to aggregate them to get a better overview of what needs to be done:

1. globbing of `$output` into `$lines` (see #281, #289)
   1. `$lines` picks up filenames from PWD when output contains `*` (see #151, #152, #156, and #158)
   1. this also skips empty lines (see #224), which might be a feature
   1. line endings are not sanitized #249
1. bad background task lifetime through FD3 handling within run #205
1. explicitly select (only) stdout/stderr for output (see #2, #47)
1. checking the exit code on the run command itself gives better error messages (see #367)
  - [ ] write up new documentation/tutorial section for changed run behavior -> the different use cases (check exit code, check output, ...) look very different (see #366) and should be achievable with the same command

My favorite solution would be to have an interface like this:

```
run command # works as before for backwards compatibility
# exit code checking
run ! <command> # check that command failed
run =3 <command> # check that command exited with 3
# output handling
--preserve-empty-lines # default in new mode
--skip-empty-lines # default in old mode
--separate-std-out-and-err # output contains only stdout and there is $stderr
-- # mark the end of options for commands that start with -
```

This should achieve all of the above goals. If possible I'd like to reuse the existing name `run`, but we need to evaluate if that would break existing users.

kolyshkin:
This is a followup to #343 and closes #377.

After fixing some more code that uses `run` left and right, I am coming back
to further improve its documentation.

In particular, I try to emphasize the only use case for run (checking the output
of a command that exits with non-zero status).

I also provide a non-working example of using run with pipes.


martin-schulze-vireso:
Nice. I think that ties in well with #358.

kolyshkin:
@sublimino PTAL

sublimino:
Really nice work! These clear, simply stated documentation changes make the basics much easier for newcomers to bats, thanks @kolyshkin! You mentioned above:

> In particular, I try to emphasize the only use case for run (checking the output
of a command that exits with non-zero status).

I use `run` with https://github.com/bats-core/bats-assert#assert_output a lot, so I've added a "subshell" method to the end of the list in https://github.com/bats-core/bats-core/pull/366/commits/e0e13c327129ac7001e6278da0759d3cb2e40bb2 too :pray:

martin-schulze-vireso:
Wow, I think this closes #1.

martin-schulze-vireso:
With the discussion going on in #367, I would want to wait for a decision there, to avoid landing this and making it obsolete in the next step.

sublimino:
@martin-schulze-vireso good point, I missed that one somehow! Will have a look over the weekend.

It could be worth merging this so there's a baseline to extend upon with #367, but don't mind either way.

martin-schulze-vireso:
Just noticed that the above PR already tried to document some of this. We should

- [ ] check that #33's ideas are not lost

martin-schulze-vireso:
> It could be worth merging this so there's a baseline to extend upon with #367, but don't mind either way.

I hope that when #367 is done, there is only one use case without `run`: Check success + get output unconditionally.

martin-schulze-vireso:
@kolyshkin Now that #367 has landed (although the exact syntax is not settled #479), how much of this PR should we retain?

kolyshkin:
I agree that (once we settle on "run with exit code check" syntax) this needs to be revised, as now there are more ways to achieve the same result.

Yet I think the documentation removal in #367 (e.g. https://github.com/bats-core/bats-core/pull/367/commits/ae2679709a608c3573b02b2f596616b2e063f632#diff-12cfeed9f9b1cc83679dd5e693fdf109f2080614ae71c90466cace54a187082dL54) is a mistake.

Instead, it should have been modified to show how to check exit codes.

For example, writing

```bash
run command args ...
echo "$output"
[ "$status" -eq 0 ]
```

(the first example from the removed doc) is OK but not a good style; either `command args ...` or `run 0 command args ...` is better.

martin-schulze-vireso:
@kolyshkin You're right. I removed that part and intended to add new documentation for it but never did. We'll let this reminder sit here until we settled on the syntax.

martin-schulze-vireso:
@kolyshkin Would you mind giving another stab at this? The `run -X`/`run !` syntax is released now, so we can move forward.

martin-schulze-vireso:
@kolyshkin I did not hear back from you. Do you plan to work on this again?

martin-schulze-vireso:
I reworked this according to the new `run` exit code check interface and added automatic output/stderr printing on exit code errors to `run`. This changed in the documentation in comparison to the old code:

1. "check the command succeeded": updated command to use `run`'s exit code check
2. "check the command failed": replaced by reference to `run !`
3. hide the command output: omitted, with new `run` exit code checking, the `run` command is actually shorter
4. make sure the command succeeded, and check its output: omitted, `run -0` is shorter
5. "pipe using `run`": made an extra section instead of bullet point. However, the recommendation is about to change with #663, so we might want to reword that later on

kolyshkin:
@martin-schulze-vireso thank you, I was just checking bats-core docs and was glad to see these examples. Sorry for not following up on this myself.

kolyshkin:
@martin-schulze-vireso I would also appreciate adding "Co-authored-by" footer next time 😉

martin-schulze-vireso:
> @martin-schulze-vireso I would also appreciate adding "Co-authored-by" footer next time 😉 

You mean in the commit message? I am sorry that this got swept away in the many revisions this has undergone. Usually I can rely on rebase to preserve this.

PR #367, opened 2020-10-13
run: add optional exit-status test

edsantiago:
If `run` is invoked with '=N' or '!' as its
first argument, it will check the exit code
of the invoked command and fail if it does
not match the requirement (exact N for =N,
or any nonzero for '!').

Additionally, `run` now logs each command invoked
as well as its entire output. This will be invisible
when tests pass, but will be shown on test failure
and may offer crucial context for someone analyzing
a failed run.

Signed-off-by: Ed Santiago <santiago@redhat.com>


edsantiago:
Sincere apologies for the churn; I was obviously not as well prepared as I thought I was.

CI is now green. I would appreciate any feedback, whether big-picture or details.

FWIW we've been using a (narrower-focused) variant of this code on our containers projects, and the enhanced logging makes it soooo much easier to track down bugs, often with just a glance at the BATS results.

kolyshkin:
This looks very interesting!

I have yet to review it in more details, for now what I see is

1. Instead of introducing an env var changing the behavior of `run`, we can add `brun`, `erun`, or `run2`, or `chk`, or something, keeping the original intact. This will actually help to smoothen the transition to the new functionality -- one can leave old test cases with run, and write the new ones using the enhanced `*run*` (whatever we call it). After some time, we can make legacy `run` emit a warning that it is obsoleted and users should switch to `erun`.

2. `expect` is not a good name as it collides with a well-known `expect` tool. This def needs to be renamed.

3. Does `expect` only supports the whole string match? I have many cases with substring match (as in `[[ "$output" == *"substring"* ]]` and a few with a regex match (as in `[[ "$output" ~= [Rr]egexp*$ ]]`. Would be nice to cover those in some way, too. I guess numeric comparisons (`[ "$output" -lt 5 ]`, `[[ "$output" <= 99 ]]`) are also not uncommon.

edsantiago:
re: nonconflicting `run` name - wow, flashback. In my original attempt, over a year ago, I considered that -- but can no longer remember why I dropped it and went with the envariable override. It might've simply been that I couldn't think of a suitable name. I do think it would make for much more readable code and fewer unexpected snags, so yes, I'll give that a lot of thought. Thank you.

re: expect - does anyone still use it? I used it heavily in the 80s and 90s, but haven't heard it mentioned even once in any of my programming communities in over ten years. Even if it's still in use (and I hope it is, because it's such a neat tool), is there likely to be any overlap of tcl-expect being used inside a BATS script? 

re: whole-string matches - my bad, I should've spent more time documenting it. Yes, it should be possible to use any operator, including the nonexistent `!~`. See [one of the newly-added tests](https://github.com/bats-core/bats-core/pull/367/files#diff-460ea3a764fb6aa17acec25ea69be9ccbf0dc8645507c506e3639c437e88aa84) for examples.

Thank you for your feedback! Coming up with good names is difficult for me, but I promise to give it a lot of thought over the next day or two.

martin-schulze-vireso:
I think most of the supported use cases are in this PR are already implemented. As far as I see, there are three main features:

1. fail when the command fails
2. print the output (only) when the command fails
3. check if the output matches

We already know [when not to use run](https://github.com/bats-core/bats-core#when-not-to-use-run) and have [bats-assert](../../bats-assert) which gives us:

1. every command that is not started via `run` will fail the test if it returns non zero
2. every command that is not started via `run` will show its outputs to stdout/stderr in the log when the test fails
3. bats-assert already offers many output checks in concert with `run` and only prints the output on failure

The advantages with the approach in this PR seem to be:

1. no additional library for the assertion (-> we could do that with bats-assert too)
2. less noise in the log compared to execution without `run` (-> bats-assert+run deals with that)
3. checking the status implicitly while printing the output

Did I miss something?

edsantiago:
@martin-schulze-vireso thank you, I was not acquainted with bats-assert. I'll be taking a close look at it soon.

kolyshkin:
Some more thoughts on `run` in its current form, and the proposed improvements to those who'll be reviewing this.

After reading and fixing lots of bats code using `run`, I have submitted https://github.com/bats-core/bats-core/pull/343 to better document it (and its various pitfalls), and I think I nailed it in https://github.com/bats-core/bats-core/pull/366. The gist of the latter it is

* do __not__ use run unless you want to check __both__ non-zero exit status and the output

as in all the other cases not using run will result in a simpler/better code. I won't duplicate the examples here, please see https://github.com/kolyshkin/bats-core/blob/patch-2/README.md#run-test-other-commands (make sure to also scroll down to "when not to use run").

Now, if users follow this advice, the result is OK, but it's not uniform. In some cases `run` is used, in some it's not. In some cases error checking is implicit (through `set -e`), in some cases it's explicit (`[ "$status" -eq 42 ]`).

Here is a not-so-good example when run is used everywhere (I've seen lots of such code recently):

```bats
@test foo {
   run cmd start something ...
   [ "$status" -eq 0 ]
   id="$output"

   run cmd exec "$id" something ...
   echo "cmd exec something: $output"
   [ "$status" -eq 0 ]
   [[ "$output" == *"success"* ]]

   run cmd exec "$id"_bad something ...
   [ "$status" -ne 0 ]
   [[ "$output" == *"invalid id"* ]]

   run cmd exec "$id" false
   [ "$status" -ne 0 ]
```

The problem is, it is too verbose. To fix this, it can be rewritten to

```bash
@test foo {
   id=$(cmd start something ...)

   output=$(cmd exec "$id" something ...)
   echo "cmd exec something: $output"
   [[ "$output" == *"success"* ]]

   run cmd exec "$id"_bad something ...
   [ "$status" -ne 0 ]
   [[ "$output" == *"invalid id"* ]]

   ! cmd exec "$id" false
```

This is better, but have the following issues:

* `run` is not used uniformly, and it is not too obvious why (unless you know the "when to use run" rule above);
* we still need echo to show the output in case we need to collect it;
* this echo has to duplicate the command and its arguments;
* when the `run` is used, it's still too verbose.

The new proposed `run` (let's call it `erun` for now) apparently solves all this:

```bash
@test foo {
   erun cmd start something ... # implicit zero status check
   id="$output"

   erun cmd exec "$id" something ... # implicit logging, too!
   [[ "$output" == *"success"* ]]

   erun '!' cmd exec "$id"_bad something ...
   [[ "$output" == *"invalid id"* ]]

   erun '!' cmd exec "$id" false    
```

Now this is both uniform and compact (note I'm using '!' to signify I want to check that cmd exited with non-zero), plus errors are (presumably) reported better.

Finally, if we add the proposed expect to the list, we can improve error reporting for mismatched output.

Hope that all make sense.

martin-schulze-vireso:
> Now, if users follow this advice, the result is OK, but it's not uniform.

Okay, this is what I did not notice above and it all makes sense now.

I think we should separate the discussion into two parts: The output checking and the return code checks. bats-assert offers solutions for both, which rely on `run`.

I think the output checking will always require an additional command, or the *run statement would become very convoluted. Therefore, the current bats-assert solution is close to optimal in this regard.

However, the return code checks are more interesting for this PR. I think the proposed solution of an exit code checked run command that offers the same output parsing interface to be compatible with existing assert_output solutions would enhance readability a lot. It would suffice to allow for exit code =0, !=0 (and maybe =X) to solve 99% of cases and the rest should be dealt with via explicit checking of `$status`.

To stay with the naming above, I would propose the following terse syntax:

```
erun exit 0 # succeeds (expects return code 0, got 0)
erun exit 1 # fails (expected return code 0, got 1)
erun ! exit 1 # succeeds (expected non zero exit code)
erun !2 exit 1 # fails (expects exit code 2)
```

edsantiago:
Thank you both for the suggestions and encouragement. I haven't played with bats-assert yet because of installation and come-up-to-speed-itis, but I did take a few moments to adopt your ideas. In particular:
* environment variable dependency is gone. New function is `runv` - is that ok? It's slightly easier to type, the 'v' can stand for 'verbose' and/or 'validation' which (maybe?) makes it more memorable.
* bare number as first argument is gone; expectation is now '!', or '!N', or '?' (or, if none, zero). I'm not sure how I feel about this: `runv 1 false` looks so much cleaner than `runv '!1' false`. Would anyone object to my adding bare-number usage back?
* `expect` is now `xpect`, and I added more usage examples to `README-runv`. One reason I've appreciated my expect helper, which I'd like to see if bats-assert can do, is that mine prints actual/expect on separate lines, making it so easy to catch single-character errors. But I can't really comment because I haven't yet dived into bats-assert. That may be a weekend task.

martin-schulze-vireso:
While working on #368 I noticed that the current debug trap tracing can be problematic with the run function. This might hit you as well, so quick heads up: please add a test that verifies the correct error location being reported for your new run function.

edsantiago:
>quick heads up: please add a test that verifies the correct error location being reported

Yikes. Thank you; seems like my best course is to wait for #368 to merge, then rebase on it.

edsantiago:
I've pushed one more commit; this time converging on `xrun` and `xpect`. Not quite as easy on the eyes as `runv`, IMO, but the alliteration makes them memorable and (I hope) easier to use & read.

And, I had a chance to use `bats-assert`, and see much to like in it, but there are many cases where I think `xpect` produces output that can save someone a lot of time. In particular, the (admittedly optional) test description can give someone an immediate insight into what went wrong:
```
 xrun seq 10001 10005
 xpect "${#lines[*]}" = 6 "Number of output lines"   | assert [ "${#lines[*]}" = "6" ]
 #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv     |  -- assertion failed --
 #|     FAIL: Number of output lines                 |  expression : [ 5 = 6 ]
 #| expected: '6'                                    |  --
 #|   actual: '5'                                    |
 #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^     |
```
I also like the vertical layout:
```
 expect="The quick brown fox jumped over the lazy dog"
 xrun echo "The quick brown fox jumper over the lazy dog"
 xpect "$expect" "Comparing fail layout on long string"
 #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
 #|     FAIL: Comparing fail layout on long string
 #| expected: 'The quick brown fox jumped over the lazy dog'
 #|   actual: 'The quick brown fox jumper over the lazy dog'
 #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

vs

 assert [ "$output" = "$expect" ]
 -- assertion failed --
 expression : [ The quick brown fox jumper over the lazy dog = The quick brown fox jumped over the lazy dog ]
 --
```
In the world I code in, too often I have to hand-edit an error log to put expected vs actual on top of each other, to see the actual differences.

martin-schulze-vireso:
> > quick heads up: please add a test that verifies the correct error location being reported
> 
> Yikes. Thank you; seems like my best course is to wait for #368 to merge, then rebase on it.

I did not look into it that deep but I see two ways this can play out: Either your code already works as intended, or rebasing won't be enough. I had to do special case handling with aborting during `run`. Now that I come do think of it, I probably would have to handle each internal function as a special case, when it is aborted via CTRL+C, it's just that the run functions are more likely to hang/be aborted.

>  `assert [ "$output" = "$expect" ]`

I think you should look into assert_equal for a fair comparison.

edsantiago:
I've addressed the wrong-line-number issue, although I'm not sure I've done so in the best possible way. I've left it as a separate commit for ease of review.

>we could simply upgrade the `run` function

That would prevent the usage I most like: `xrun command` with implicit zero-exit check. I think requiring test writers to remember to write `run '!0' command` on each invocation is a heavy burden, which means it won't be done. Remembering to type `xrun` isn't much better, but I think it's a slightly easier habit to form. I really don't know what the big world out there prefers, though.

martin-schulze-vireso:
> That would prevent the usage I most like: xrun command with implicit zero-exit check [...] I really don't know what the big world out there prefers, though.

Well, I don't know either but I added a rationale to the other issue (https://github.com/bats-core/bats-core/issues/377#issuecomment-723662815).

However, I agree that return code checks should be default. In a sense we already have this situation by having "normal" commands (without `run`) be return code checked, so using `run` explicitly disables this check. However, the way many people seem to think of bats tests is: you have to `run` every command you want to check.

> I think requiring test writers to remember to write `run '!0'` command on each invocation is a heavy burden

I was a bit unclear. I think we should have something like:

- `run <command>` -> current solution, ignores return code
- `run ? <command>` -> expect success, this is equivalent to `<command>` (except unconditionally printing the output)
- `run ! <command>` -> expect failure
- `run =<return code> <command>` -> expect <return code>
- `run !=<return code> <command> -> expect not <return code>

The advantage over a separate function is the unified control flow, which reduces special case handling.

edsantiago:
Nice. I like the `=N` form, that's short and memorable and friendly. I like `!` also, but am not sure I understand the reasoning behind `?` or `!=N`. Could you elaborate on how/why you see these being used?

This has grown somewhat beyond the scope of my original submission, so I've prepared a separate local branch with just the changes to `run` as described in your last comment (except for `?` and `!=`, pending further exploration), and with no `xpect` because that clearly merits a separate discussion. I'm leaning toward submitting it as a separate PR and closing this one, because it's so different, but that loses this conversation. What do you recommend?

martin-schulze-vireso:
>  I understand the reasoning behind ? or !=N.

I wanted to make the return code check explicit with a short notation so `?` basically says: Check for 0, but this is somewhat redundant to `=0`, so we could omit that. `!=N` is intended for cases where we don't want a specific non-null return code but that might be so rare that we could omit that too. One could always store the `$status` and do more complex checks later on. The beauty of letting `run` do the checks is that the failing command is visible in the error message.

> I'm leaning toward submitting it as a separate PR and closing this one, because it's so different, but that loses this conversation. What do you recommend?

Well you could force push over this branch or just reference this PR in the other and copy over a summary of the discussion here.

edsantiago:
I've pushed the new branch here, updated PR title and description. What do you think of this approach?

README.md
```diff
@@ -340,14 +340,28 @@ The `$status` variable contains the status code of the command, and the
 `$output` variable contains the combined contents of the command's standard
 output and standard error streams.
 
+If invoked with one of the following as the first argument, `run`
+will perform an implicit check on the exit status of the invoked command:
+
+    =N  expect exit status N (0-255), fail if otherwise
+    '!' expect nonzero exit status (1-255), fail if command succeeds
```
martin-schulze-vireso, line 347: I think `run !` without the `''` is perfectly fine. My bash documentation says this:

> Reserved words are words that have a special meaning to the shell.  The following words are recognized as **reserved when unquoted** and either the **first word of a simple command** (see SHELL GRAMMAR below) or the **third word of  a case or for** command: 
> **!** case  coproc  do done elif else esac fi for function if in select then until while { } time [[ ]]

And something like the following just works:
```
$ echo ! word
! word
```

README.md
```diff
@@ -377,29 +391,11 @@ command args ...
...
-is equivalent to
-
-```bash
-command ... >/dev/null
-```
-
-Note that the output is only shown if the test case fails.
-
-3. In case you need to assign command output to a variable (and maybe check
+2. In case you need to assign command output to a variable (and maybe check
```
martin-schulze-vireso, line 394: I think this whole "when not to use run" section needs to be renamed rewritten, now that we actually want to support these cases directly with run.

lib/bats-core/test_functions.bash
```diff
@@ -29,7 +29,37 @@ load() {
   source "${file}"
 }
 
+#########
+#  die  #  Abort with helpful and visible message
+#########
+die() {
```
martin-schulze-vireso, line 35: I'd be cautious with adding such short function names without prefix in this context. That might clash with user's code.

lib/bats-core/test_functions.bash
```diff
@@ -29,7 +29,37 @@ load() {
...
+}
+
 run() {
+  local expected_rc=
+
+  # Possible first arguments: '!' = any nonzero, '=n' = exactly n
+  if [[ "$1" = '!' ]]; then
+    expected_rc=-1
+    shift
+  elif [[ "$1" =~ ^=([0-9]|[1-9][0-9]|[12][0-9][0-9])$ ]]; then
```
martin-schulze-vireso, line 55: Wouldn't `^=[0-9]+$` suffice? It seems the goal is to restrict the value to 0-255 here, which it already fails at because 299 is matches too. If we really want to check, I would propose to check if `$expected_rc` is out of bounds and print a descriptive error message instead of just silently ignoring the check.

I'd go as far as saying: check the first character, if its `!` ensure that nothing else is there. If its `=` check that only a number follows or give an error. If we need to run a command without checks that starts with `!` or `=` we can add a `--` marker to separate the command from the options.

lib/bats-core/test_functions.bash
```diff
@@ -29,7 +29,37 @@ load() {
...
+    expected_rc=${1#=}
+    shift
+  fi
+
+  # stdout is emitted only on error; this echo is to help debug test failures
+  printf "${BATS_PS1}%s\n" "$*"
```
martin-schulze-vireso, line 61: I understand that this should print the command under `run` even when there is no failure. In the failure case I find this to duplication to be counter-productive. In the case of a later failure (not in this run) I would reserve this behavior for an eventual trace functionality.

lib/bats-core/test_functions.bash
```diff
@@ -42,6 +72,37 @@ run() {
   IFS=$'\n' lines=($output)
   IFS="$origIFS"
   set "-$origFlags"
+
+  # Show results. Without quotes, multiple lines are glommed together into one
+  if [ -n "$output" ]; then
+    printf "%s\n" "$output"
```
martin-schulze-vireso, line 78: Again, this should be part of an eventual trace feature, not of everyday run. I am surprised that this change of behavior does not fail some of our tests already.

lib/bats-core/test_functions.bash
```diff
@@ -42,6 +72,37 @@ run() {
...
+  fi
+
+  # Check exit status if requested
+  if [[ "$status" -ne 0 ]]; then
```
martin-schulze-vireso, line 82: Again, this would be great for an eventual trace feature but not unconditional prints.

lib/bats-core/test_functions.bash
```diff
@@ -42,6 +72,37 @@ run() {
...
+        printf "(** EXPECTED %d **) " $expected_rc;
+      fi
+    fi
+    printf "]\n"
+  fi
+
+  if [[ -n "$expected_rc" ]]; then
+    if [[ "$expected_rc" = "-1" ]]; then
+      if [[ "$status" -eq 0 ]]; then
+        die "exit code is $status; expected nonzero"
```
martin-schulze-vireso, line 100: I find the messages from `die` to be a stark difference from the current standard outputs. I think the current error handling already gives enough visibility to problems.

lib/bats-core/tracing.bash
```diff
@@ -119,15 +120,25 @@ bats_trim_filename() {
...
-		# Therefore we capture the stacktrace but use it only after the next debug
-		# trap fired.
-		# Expansion is required for empty arrays which otherwise error
-		BATS_CURRENT_STACK_TRACE=("${BATS_STACK_TRACE[@]+"${BATS_STACK_TRACE[@]}"}")
-		bats_capture_stack_trace
-	fi
+	local stack_trace_offset=
+	if [[ "$1" == $BATS_ROOT/lib/* || "$1" == $BATS_ROOT/libexec/* ]]; then
+        	if [[ "${FUNCNAME[1]}" != "run" ]]; then
+			return
```
martin-schulze-vireso, line 126: The formatting seems to be a bit off here.

lib/bats-core/tracing.bash
```diff
@@ -119,15 +120,25 @@ bats_trim_filename() {
...
+                fi
+
+                # We are inside 'run', which can throw a test failure.
+                # Make sure we log the calling function (presumably in user's
+                # .bats file) so we can emit proper line numbers in messages.
+                stack_trace_offset=1
```
martin-schulze-vireso, line 132: I am not sure if this is the best way to solve it. I need to think about this issue a bit more.

libexec/bats-core/bats
```diff
@@ -279,6 +279,15 @@ done
 # shellcheck source=lib/bats-core/validator.bash
 source "$BATS_ROOT/lib/bats-core/validator.bash"
 
+# Prompt to show in verbose results: distinguish between root and non
+if [ -z "$BATS_PS1" ]; then
```
martin-schulze-vireso, line 283: What kind of problem are you trying to solve here? Shouldn't users be aware whether they ran their tests under root or not?

test/run-check-status.bats
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+# Positive assertions: each of these should succeed
+@test "basic return-code checking" {
+  run      true
```
martin-schulze-vireso, line 7: I think `run false` would be in order too, just to show that "normal" `run` suppresses the return code.

test/run-check-status.bats
```diff
@@ -0,0 +1,111 @@
...
+        lines[$lineno]="${lines[$lineno]}   [MATCHED]"
+        break
+      fi
+    done
+  done
+}
+
+@test "run: unexpected fail" {
+  local -a commands=("run =0 false")
+  _run_test "\(in test file .*, line 2\)" \
```
martin-schulze-vireso, line 66: I find `_run_test` quite involved. The usual way of doing this would be extra files for each test that can then be called via `bats testfile`. I am not sure if this solution will be harder to maintain. It definitely requires more understanding than simply running a file with bats. @sublimino What do you think about this approach?

martin-schulze-vireso:
One thing that is still nagging me: `run`'s current behavior is to ignore the return code by default to allow for checking later on. I guess this makes sense from that standpoint where you just use the command without run if you want the implicit check. But then there is the use case of wanting to check the output but not ignoring the return code. 

I'd say ignoring the return code is far less common than checking it, so checking should be the default (as @edsantiago pushed for the whole time). Maybe we should offer the following solution:

1. add `=ignore` or `=any` to explicitly ignore the return code
2. add a variable `BATS_MINIMUM_VERSION` which will toggle the default behavior of `run` when it is set to a version equal to or above of the version where this change is introduced (this is similar to [CMake's policies](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#manual:cmake-policies(7)))

This should allow for an explicit migration to the new behavior while allowing existing code to continue working. The only drawback that I see is forgetting to set this variable, but we might nag users with a warning if it is not set.

edsantiago:
This is wonderful feedback and much appreciated. I'm in long meetings today and will have much catching-up to do the rest of the week, so rest assured I will get to this as soon as I can. Thank you.

README.md
edsantiago: Oh, nice, thank you. It looks a little uncomfortable to my eye, due to my fear that bash will interpret it as a history event... but I see there's no conflict, and I can learn.

README.md
edsantiago: Hmmm, agreed. I thought I had done a better job of editing that. Deferring for now, until we converge on something we're all happy with. (Note to self, I'd also like to add a warning about using redirection in `run`)

lib/bats-core/test_functions.bash
edsantiago: Point taken. Renamed to `_bats_die()` for now. I'm uncertain about the underscore, because in our project use the last few years we've found it very convenient to have a test-accessible `die MESSAGE`.

lib/bats-core/test_functions.bash
edsantiago: Thanks. I reworked that code to make it more robust, and added tests.

lib/bats-core/test_functions.bash
edsantiago: May I push back on this? I can't count the number of times this has saved us in analyzing test failures: being able to see earlier commands run, and their output, has been invaluable when trying to understand a sequence of events leading to failures. Put another way, the root cause of a failure is sometimes buried in a command run some time before.

lib/bats-core/tracing.bash
edsantiago: It does not conform to the [formatting guidelines](https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md#formatting), but neither does the rest of that file. I chose to remain internally consistent within that context.

lib/bats-core/tracing.bash
edsantiago: Suggestions welcome!

libexec/bats-core/bats
edsantiago: Actually, that too is invaluable in our project CI: the [podman](https://github.com/containers/podman) tool is internally very different between root and rootless contexts, and often the most important first question when looking at a failure is: root, or rootless? (We expose that in other places in CI logs, but the prompt makes it even easier). I'm not strongly committed to this, though.

test/run-check-status.bats
edsantiago: Good point. Added, but elsewhere.

test/run-check-status.bats
edsantiago: One of my earlier iterations had a set of `fixtures` tests (for `xpect`). What I was trying to solve here was being able to have the commands and expected output very close together, for maintainability. I could not find a good way to do so using individual fixture test files; but it's very likely I may have missed something.

edsantiago:
New push, incorporating your feedback. Pushed as a separate commit, for ease of review.

As for changing the behavior of `run` via envariable, I'm open to that but I think it merits careful discussion.

Thank you again for all your time.

lib/bats-core/test_functions.bash
martin-schulze-vireso: I'd like to maintain a minimal interface on bats, because each "public" function needs to maintain backwards compatibility. The `die` function can easily be added, if needed.

lib/bats-core/test_functions.bash
martin-schulze-vireso: Yes, I experienced something like this myself. However, one of the core tenets of `run` is being silent to the log by default. As indicated above, I have absolutely nothing against allowing a switch like `--trace` that would make run more verbose if the user chooses to.

lib/bats-core/tracing.bash
martin-schulze-vireso: Yes, the formatting is not consistent but in the preview the return is less indented than the if. Maybe this is a tabs vs spaces issue?

test/run-check-status.bats
martin-schulze-vireso: Well, the code in question is in a separate file but the bats output still shows what went wrong, so I feel there should be enough context to see what is going on.

lib/bats-core/test_functions.bash
edsantiago: >one of the core tenets of run is being silent to the log by default

Is that a hard requirement, or is it even documented? I've been struggling with this comment, because without verbose `run` I don't see any point to this PR. But I think maybe I haven't described how it works well enough. This usage is absolutely silent on everything except failure; and on failure, it only shows `run`s for the failed test - nothing else. Given a test setup with lots of tiny tests, this does not add clutter. (If, OTOH, you have one giant `@test` that runs hundreds of commands, it could be harder to read). 

To put it another way, here's a [recent example](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5159711932678144/html/sys-podman-fedora-33-rootless-host.log.html) of a failure we had today. In this case the failure message shows container `0e7a2d3`, but that would be utterly useless without the earlier context in which we see `0e7a2d3` as the output of the second `podman run`. We see failures like these often, typically race conditions, where every little bit of context is critical for identifying the cause. As you can see, only the failing test shows `run` output; the other 100+ passing/skipped tests are silent. Does that help alleviate your concerns?

martin-schulze-vireso: > Is that a hard requirement, or is it even documented?

I don't think it is documented and I'd say it is not a hard requirement per se but you have to remember that there are loads and loads of tests out there which rely on the current default behavior. 

> This usage is absolutely silent on everything except failure

Yes, that is the default behavoir of bats. However, there are cases like this:

```
run succeeding_command_with_enormous_output
assert_output ...

run failing_command_with_output
assert_output ...
```

Where you'd have to wade through lots and lots of uninteresting output. The current implementation makes all users pay for this. All I am asking for is a switch that allows users to decide for themselves if they want that much output or not. This is what I was alluding to with the (not yet implemented) `--trace` switch. You could start the whole bats run with that flag or we should allow for more finegrained control, e.g. via 

```
BATS_TRACE=only-commands # print only commands in run here
run command 1
...
run command n
BATS_TRACE=output # print commands and stdout/stderr here
run command n+1
...
unset BATS_TRACE # print neither command nor output in the following
run command 
```

The unset `BATS_TRACE` case resembles the current default and should remain the default.

I have seen several tests where the amount of output of a single command is too much and there are several in the test.

martin-schulze-vireso:
I'd like to get this one going again and need some voting (from @sublimino @cyphar @denhamparry) here. My main issues right now:

1. `_bats_die` is very loud and somewhat duplicates bats-assert's functionality. Do we really need/want it in core? (Then again, moving the exit code check into bats-core also duplicates assert_success/failure of bats-assert but is definitely intended here.)
2. I have an uneasy feeling about `_run_test`. IMHO, it looks too complicated for the benefit it achieves. I also have the feeling that it might rely too much on internals that might change in the future.
3. Do we want to print `$output` (and other diagnostics) unconditionally and let the upper levels filter away the output of successful tests? It feels wasteful and like a breaking change to the current behavior.

Also review the open discussion items to inform your decision, please.

edsantiago:
Thank you for your followup. I too am interested in pursuing this, but prefer to wait for a more defined idea of what will be acceptable to the community. One comment, in case it helps:

>Do we want to print $output (and other diagnostics) unconditionally 

I find BATS useless without this: just two weeks ago I had a CI job fail; the failure was a flake in a `run something` which did not use a logging `run`, and without `$output` it was impossible to track down the failure. (I managed to get very, very lucky and reproduce the failure on a test VM). However: for purposes of granularity, I'm willing to move that functionality to a separate PR, and concentrate here on the status-checking.

NorseGaud:
@edsantiago check out https://github.com/bats-core/bats-core/pull/467. I think that will be a useful feature for you.

martin-schulze-vireso:
I rebased onto current master and resolved the conflicts. I also restructured the test code and removed the tracing parts that might go into #467 and `_bats_die` together with it's verbose output. I hope I did not break anything.

martin-schulze-vireso:
One thing that is still missing is the printing of $output. I removed that because it is intended for the trace feature. However, I find the usecase of printing the output on error quite compelling. So when we don't check the exit code, run should behave as before and remain silent. When we do exit code checking and the check fails, I'd propose to print output.

martin-schulze-vireso:
@edsantiago Do you mind reviewing if I broke some of your intentions?

@bats-core/bats-core I'd also like the other maintainers' opinions on this.

edsantiago (commented):
@martin-schulze-vireso thank you for taking this on again. Looks good at first pass, but by now this is different enough from my original PR that I'm OK with you closing and resubmitting as yours.

martin-schulze-vireso:
@edsantiago I am okay with sharing the fame for your ground work if you are okay with sharing the `git blame`. ;)

martin-schulze-vireso (approved):

PR #369, opened 2020-10-17
Add --shellcheck switch for checking *.bats files

martin-schulze-vireso:
Bats files are not valid bash, so ~shellcheck~ some linters will choke on the special syntax, thus preventing us from checking our *.bats files.
This PR adds the `--shellcheck` flag which feeds the preprocessed bats files, that are valid bash, to shellcheck. Thus, bats acts as a wrapper around shellcheck to allow for checking test files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM bash:${bashver}
 
 # Install parallel and accept the citation notice (we aren't using this in a
 # context where it make sense to cite GNU Parallel).
-RUN apk add --no-cache parallel ncurses && \
+RUN apk add --no-cache parallel ncurses shellcheck && \
     mkdir -p ~/.parallel && touch ~/.parallel/will-cite
 
 RUN ln -s /opt/bats/bin/bats /usr/local/bin/bats
--- a/README.md
+++ b/README.md
@@ -225,6 +225,8 @@ Usage: bats [OPTIONS] <tests>
   -o, --output <dir>        Directory to write report files
   -p, --pretty              Shorthand for "--formatter pretty"
   -r, --recursive           Include tests in subdirectories
+  --shellcheck              Run shellcheck on the *.bats files instead of 
+                            running the tests
   -t, --tap                 Shorthand for "--formatter tap"
   -T, --timing              Add timing information to tests
   -v, --version             Display the version number
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -301,11 +301,11 @@ if [[ -n $run_shellcheck ]]; then
       done
     }
 
-    BATS_SHELLCHECK_OPTIONS+=(--shell=bash) # override shellcheck's shell detection
+    BATS_SHELLCHECK_OPTIONS+=("--shell=bash") # override shellcheck's shell detection
     # since we pipe shellcheck's output through the function above, shellcheck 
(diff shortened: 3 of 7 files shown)


jansorg:
ShellCheck 0.7+ supports .bats file syntax, https://github.com/koalaman/shellcheck/blob/master/CHANGELOG.md#added-2, afaik.

martin-schulze-vireso:
Shame on me for not checking again. I just remembered https://github.com/bats-core/bats-core/issues/192#issue-415262835 and thought there was an easy fix that I needed right now.

I am not opposed to closing this PR. However, it still offers some benefits:
1. the shellcheck executable path can be overridden, so it might be possible to use this with other checkers/linters as well
2. if we were to add more custom syntax, we don't have to wait for shellcheck and others to catch up

jansorg:
I don't have an opinion here, either way is fine, I guess.

martin-schulze-vireso:
I was thinking about this again with regard to `shfmt`. However, this would require transforming the formatted file back, which might not be trivial since linenumbers could have changed. Anyways, I think the correct way forward is a bash compatible test format to begin with, so we don't have compatibility issues with linters/formatters that can deal with correct bash.

sublimino:
Agreed it would be better to be actual Bash here. 

Until then, as an FYI, I have a (nasty!) `shfmt-bats` function. The important bit is the perl line, the rest is just a messy wrapper:

```bash
cat "${FILE}" | perl -pe 's/^(\@test ['\''"].*['\''"])[[:space:]]*\{[[:space:]]*/\1\n{\n/g' | shfmt /dev/stdin | perl -pe 's/^(\@test ['\''"].*['\''"])[[:space:]]*/\1 /g') | sponge "${FILE}"
```

In a function:

```bash
shfmt-bats () 
{ 
    ( _bash_safe_mode;
    local FILES;
    local IS_WRITE=0;
    if [[ "${1:-}" != "" ]]; then
        FILES=("${@}");
        for FILE in "${FILES[@]}";
        do
            if [[ "${FILE}" == "-w" || "${FILE}" == "--write" ]]; then
                IS_WRITE=1;
            else
                if [[ -d "${FILE}" ]]; then
                    shopt -s nullglob;
                    local DIR_FILES=(${FILE}/*);
                    shopt -u nullglob;
                    FILES+=("${DIR_FILES[@]}");
                fi;
            fi;
        done;
    else
        FILES=("/dev/stdin");
    fi;
    local FILE;
    for FILE in "${FILES[@]}";
    do
        if [[ ! -f "${FILE}" || "${FILE}" == "-w" || "${FILE}" == "--write" ]]; then
            continue;
        fi;
        echo "${FILE}";
        if OUTPUT=$(cat "${FILE}" | perl -pe 's/^(\@test ['\''"].*['\''"])[[:space:]]*\{[[:space:]]*/\1\n{\n/g' | shfmt /dev/stdin | perl -pe 's/^(\@test ['\''"].*['\''"])[[:space:]]*/\1 /g');
        [[ "${OUTPUT}" != "" ]]; then
            if [[ "${IS_WRITE}" == 1 ]]; then
                echo "${OUTPUT}" | sponge "${FILE}";
            else
                echo "# OUTPUT" 1>&2;
                echo "${OUTPUT}";
            fi;
        fi;
    done;
    return 0 )
}
```

PR #370, opened 2020-10-17
Include symlinks during --recursive operation

jo-bitsch:
Previously, to implement recursive operation, `bats` used `find` without the `-L` switch. As a consequence, when the test cases were structured in subdirectories, where some of the folders were symlinks or the testcase itself was a symlink to a testcase, it would be ignored. The `-L` option of `find` follows symlinks and checks the test predicates on the target of the link.

Changes:
- added the `-L` option to the `find` call in the main program.
- added a new fixture in `suites/recursive_with_symlinks` that refers to the recursive test case via symlink.
- added a new test case in `suite.bats` that is skipped on OSes where symlinks are not functional.

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -264,7 +264,7 @@ for filename in "${arguments[@]}"; do
     if [[ "$recursive" -eq 1 ]]; then
       while IFS= read -r -d $'\0' file; do
         filenames+=("$file")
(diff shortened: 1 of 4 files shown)


martin-schulze-vireso (approved):

martin-schulze-vireso:
LGTM, thanks for also providing a test.

Issue #371, opened 2020-10-25
Debian 10 installation not working OOTB


ghost:
I posted on Gitter, but there doesn't look like there is much activity over there.  After installing with Linuxbrew or installing it from the source code, I get the same errors, albeit the line number is different for Linuxbrew.  The below error is repeated for each test in the *pyenv* project, which is the reason why I installed bats.
___
Command I am using is: `bats $HOME/pyenv/test`

Error message:

```bash
/usr/local/libexec/bats-core/bats-exec-file: line 242: bats-exec-test: command not found
/usr/local/libexec/bats-core/bats-exec-file: line 242: bats-exec-test: command not found
/usr/local/libexec/bats-core/bats-exec-file: line 242: bats-exec-test: command not found
/usr/local/libexec/bats-core/bats-exec-file: line 242: bats-exec-test: command not found
/usr/local/libexec/bats-core/bats-exec-file: line 242: bats-exec-test: command not found
   bats warning: Executed 0 instead of expected 174 tests

174 tests, 0 failures, 174 not run
```
___
**Additional info**

bats-core location:
```zsh
➜  pyenv (master) ✗) which bats
/usr/local/bin/bats
```

bats executable search:
```zsh
(master)⚡)% ls /usr/local/bin | grep bats
bats
```

bromps:
I'm getting the same issue with Homebrew on macOS Catalina 10.15.7 as well.

bromps:
Copying the rest of the executables to the `/usr/local/bin` directory does the trick.  However, I would think that the main `bats` executable would be smart enough to find the libexec files.

```
sudo cp /usr/local/Cellar/bats-core/1.2.1/libexec/bats-core/bats-* /usr/local/bin
```

martin-schulze-vireso:
Thanks for trying gitter first.

Could you give the exact steps you took for installing on each of the platforms?

I just tried `brew install bats-core` and got 1.2.1 where I cannot reproduce this issue with a simple test file.

martin-schulze-vireso:
I can reproduce with pyenv. On closer inspection their `test/` contains a `libexec/`. Maybe that messes up something within bats? 1.2.1 is the first version with exec-file which sources the scripts before running them, so they might mess up internal state.

bromps:
Ah, good catch.  Since you were able to reproduce with pyenv, but not in other projects, I'm going to open an issue in that repo.  Thanks for looking into this.

martin-schulze-vireso:
[...]

PR #387, opened 2020-12-17
Use absolute executable path to fix command-not-found error when PATH updated in tests.

uzxmx:
Dear maintainers,

First, please look at below snippet:

https://github.com/bats-core/bats-core/blob/e81e2ef3a2b4d6a45d5f50deeb3d4f0e0aeca8b7/libexec/bats-core/bats-exec-file#L85

Because of the `source` command shown above, if `PATH` environment variable is reset or updated in user tests, which causes `$BATS_ROOT/libexec/bats-core` removed from `PATH`,  it will fail with below error:
```
/usr/local/libexec/bats-core/bats-exec-file: line 242: bats-exec-test: command not found
```

I met the above problem when I ran tests of [ruby-build](https://github.com/rbenv/ruby-build). In all its tests (e.g. [this test](https://github.com/rbenv/ruby-build/blob/763b14af70523064c9a81469c471b21423a47f8b/test/arguments.bats#L3)), they load a [test_helper](https://github.com/rbenv/ruby-build/blob/master/test/test_helper.bash#L8) script which resets `PATH` before running tests, which causes `command not found` error.

IMHO, no matter if the user is updating `PATH` purposely or unpurposely, we should make this library more robust to satisfy such need or tolerate such error. So I made this PR.

Please have a review. Thanks for your time!

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -68,7 +68,8 @@ expand_path() {
   printf -v "$result" '%s/%s' "$dirname" "${path##*/}"
 }
 
-export BATS_LIBEXEC="$(dirname "$(bats_readlinkf "${BASH_SOURCE[0]}")")"
+BATS_LIBEXEC="$(dirname "$(bats_readlinkf "${BASH_SOURCE[0]}")")"
+export BATS_LIBEXEC
 export BATS_CWD="$PWD"
 export BATS_TEST_PATTERN="^[[:blank:]]*@test[[:blank:]]+(.*[^[:blank:]])[[:blank:]]+\{(.*)\$"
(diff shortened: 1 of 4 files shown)


martin-schulze-vireso (changes requested):
Please provide a test case that triggers the problem as well.

martin-schulze-vireso:
Note this is the same error as reported in #371

uzxmx:
@martin-schulze-vireso  

Test added. BTW, I found that when I added another test after the test case in my commit to watch `PATH`, it restored to its original value. So I didn't  change `PATH` back in `test/fixtures/bats/update_path_env.bats`. And this simple test can verify if the absolute path is not used, it won't pass.

martin-schulze-vireso (approved):

PR #386, opened 2020-12-17
preprocessing: clean up incorrect unset variable handling

cyphar:
Without this patch, a script that has "set -u" set will end up with one
of the following errors:

  preprocessing.bash: line 3: TMPDIR: unbound variable
  preprocessing.bash: line 28: BATS_TEST_SOURCE: unbound variable

Fixes: #385
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

- [x] I have reviewed the [Contributor Guidelines][contributor].
- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md
[coc]:         https://github.com/bats-core/bats-core/blob/master/docs/CODE_OF_CONDUCT.md

PR #392, opened 2020-12-21
Fix unbound variables inside bats

### LINKED ISSUE #385: unbound variables: TMPDIR, BATS_TEST_SOURCE with nounset
preprocessing.bash currently assumes TMPDIR and BATS_TEST_SOURCE are declared (probably more correct to say it assumes nounset is not set :-) ):
https://github.com/bats-core/bats-core/blob/master/lib/bats-core/preprocessing.bash#L3
https://github.com/bats-core/bats-core/blob/master/lib/bats-core/preprocessing.bash#L28

This will cause bats to fail when TMPDIR is not set (e.g. Alpine, as the Docker image currently uses), and `set -o nounset` is set and exported to subshells (e.g. via `export SHELLOPTS`). The resulting error message is
`preprocessing.bash: line 3: TMPDIR: unbound variable`

The test should probably be ?
`if [[ -z "${TMPDIR:-}" ]]; then`
and
`if [[ -z "${BATS_TEST_SOURCE:-}" ]]; then`

martin-schulze-vireso:
This builds on top of #386, fixes #385

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -76,7 +76,7 @@ export BATS_TEST_PATTERN_COMMENT="[[:blank:]]*([^[:blank:]()]+)[[:blank:]]*\(?\)
 export BATS_TEST_FILTER=
 export PATH="$BATS_LIBEXEC:$PATH"
 export BATS_ROOT_PID=$$
-if [[ -z "$TMPDIR" ]]; then
+if [[ -z "${TMPDIR-}" ]]; then
   export BATS_TMPDIR="/tmp"
 else
   export BATS_TMPDIR="${TMPDIR%/}"
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -105,8 +105,8 @@ set -- "${arguments[@]}"
 arguments=()
 
 unset flags recursive formatter_flags
-flags=()
-formatter_flags=()
+flags=('--dummy-flag') # add a dummy flag to prevent unset varialeb errors on empty array expansion in old bash versions
+formatter_flags=('--dummy-flag') # add a dummy flag to prevent unset varialeb errors on empty array expansion in old bash versions
 formatter='tap'
 report_formatter=''
 recursive=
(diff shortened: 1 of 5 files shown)


libexec/bats-core/bats
```diff
@@ -76,7 +76,7 @@ export BATS_TEST_PATTERN_COMMENT="[[:blank:]]*([^[:blank:]()]+)[[:blank:]]*\(?\)
 export BATS_TEST_FILTER=
 export PATH="$BATS_LIBEXEC:$PATH"
 export BATS_ROOT_PID=$$
-if [[ -z "$TMPDIR" ]]; then
+if [[ -z "${TMPDIR-}" ]]; then
```
sublimino, line 79: Should this be `:-` as per https://github.com/bats-core/bats-core/pull/392/files#diff-585433432247165e74657c49237de62aa771151126d24c8ee4aa6f47befc7578R28 (and my preference, I think it's a little safer by default)

libexec/bats-core/bats
```diff
@@ -105,8 +105,8 @@ set -- "${arguments[@]}"
 arguments=()
 
 unset flags recursive formatter_flags
-flags=()
-formatter_flags=()
+flags=('--dummy-flag') # add a dummy flag to prevent unset varialeb errors on empty array expansion in old bash versions
+formatter_flags=('--dummy-flag') # add a dummy flag to prevent unset varialeb errors on empty array expansion in old bash versions
```
sublimino, line 109: Typo on "varialeb"

libexec/bats-core/bats-exec-suite
```diff
@@ -8,7 +8,7 @@ have_gnu_parallel=
 bats_parallel_args=()
 bats_no_parallelize_across_files=${BATS_NO_PARALLELIZE_ACROSS_FILES-}
 bats_no_parallelize_within_files=
-flags=()
+flags=('--dummy-flag') # add a dummy flag to prevent unset varialeb errors on empty array expansion in old bash versions
```
sublimino, line 11: Typo on "varialeb"

sublimino (commented):
Nits and typos

PR #405, opened 2021-02-06
Add Github Actions for tests

### LINKED ISSUE #404: Fix CI Docker hub rate limit
The CI builds [started failing](https://travis-ci.org/github/bats-core/bats-core/jobs/756881727) due to [Docker hub's rate limit](https://www.docker.com/increase-rate-limits). This leaves us without bash version testing.

I already tried to add `docker login` to the build but the associated credential variables won't be available in external PRs.

How can we work around this? Even if we had a paid plan for Docker hub, we would need to expose the login credentials to external users for their PRs to work.

The bash images don't seem to be available on alternative registries, yet. I opened an [issue](https://github.com/tianon/docker-bash/issues/23) on their repo.

martin-schulze-vireso:
Fixes #404 

I had to work around some problems:

1. Github Actions does [not provide a  TTY](https://github.com/actions/runner/issues/241) which trips up the pretty formatter/tput detection -> I added workarounds that should be removed, once the issue is closed
2. the parallel tests seem to break due to slower CI systems -> I split them off into their own runner and turned off parallelization on the test file itself


.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+  workflow_dispatch:
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run shellcheck
+        run: |
+          sudo apt-get install shellcheck
```
sublimino, line 21: This may need an `sudo apt update &&` to mitigate against repo updates

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
...
+        os: ['ubuntu-20.04', 'ubuntu-18.04', 'ubuntu-16.04']
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run test on OS ${{ matrix.os }}
+        shell: 'script -q -e -c "bash {0}"' # work around tty issues
+        run: |
+          bash --version
+          bash -c "time bin/bats --formatter tap test"
+          # allow for some parallelity without GNU parallel, e.g. on MacOS
```
sublimino, line 37: This comment seems not to be relevant to non-MacOS, but I defer to you @martin-schulze-vireso :pray:

sublimino (commented):
Just a package update nit and question :pray:

.github/workflows/tests.yml
martin-schulze-vireso: You're right, I just copied this over from the travis code without thinking about it.

.github/workflows/tests.yml
martin-schulze-vireso: A adapted the text a little

.github/workflows/tests.yml
jasonkarns: I think there's a good chance that shellcheck is included natively on the Ubuntu image

martin-schulze-vireso: @jasonkarns It might be, but using update and install should ensure we get the latest available version. When I started with `ubuntu-latest` we got 16.04 which gave us shellcheck 0.4...

.github/workflows/tests.yml
```diff
@@ -0,0 +1,100 @@
+name: Run CI tests
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
jasonkarns, line 9: I'm not clear why we're limiting these here. I'd say we want the tests to run on all branches, no? A PR to a PR for instance, should still run the tests.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,100 @@
+name: Run CI tests
```
jasonkarns, line 1:
```suggestion
name: Tests
```

Just a nit here. I think the "run/ci" bits are redundant?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,100 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run shellcheck
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install shellcheck
+          ./shellcheck.sh
+  
+  # This workflow contains a single job called "build"
```
jasonkarns, line 25:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,100 @@
...
+  linux:
+    strategy:
+      matrix:
+        os: ['ubuntu-20.04', 'ubuntu-18.04', 'ubuntu-16.04']
+        options:
+          - ''
+          # allow for some parallelity without GNU parallel, since it is not installed by default
+          - 'BATS_NO_PARALLELIZE_ACROSS_FILES=1 BATS_NUMBER_OF_PARALLEL_JOBS=2'
```
jasonkarns, line 33: Are these options being used somewhere that I'm not seeing?

jasonkarns (approved):
After this lands, I'd like to move towards having a single job for all the environments. Any differences in running per OS should be pain points I think we'd want to resolve/eliminate as much possible within bats.

To that end, I'd like to dig in later as to the root cause of the TTY issue. What are we doing that makes it sensitive to the bug on Actions?

.github/workflows/tests.yml
martin-schulze-vireso: I started with the example template, so these are the default values. I assume we should just use *?

.github/workflows/tests.yml
martin-schulze-vireso: That should read env_vars instead of options. This is very likely a renaming error.

martin-schulze-vireso:
> After this lands, I'd like to move towards having a single job for all the environments. Any differences in running per OS should be pain points I think we'd want to resolve/eliminate as much possible within bats.

I was starting out with a single job but that did not work. The mean reason for differences is the workaround handling for the TTY problem. Additionally, we have differences in availability of GNU parallel.

The TTY issue has already a merged PR but it seems not fixed, so I assume the release is pending.
Fixing the availability of GNU parallel might not be worth the effort: I am planning on replacing it completely with the bash semaphore implementation that is already in use for the parallelization within files.

> To that end, I'd like to dig in later as to the root cause of the TTY issue. What are we doing that makes it sensitive to the bug on Actions?

There are two pain points:

1. without docker: tput does not get the terminal it wants
2. with `docker run -it`: `the input device is not a TTY`

.github/workflows/tests.yml
jasonkarns: The easiest thing is to eliminate the filter entirely. I'll add another comment suggestion that's committable.

jasonkarns: Oh, you already got it! 😸

.github/workflows/tests.yml
```diff
@@ -0,0 +1,94 @@
+name: Tests
+
+# Controls when the action will run. 
+on: [push, pull_request]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
jasonkarns, line 7:
```suggestion
on: [push, pull_request, workflow_dispatch]
```

This should fix the syntax error.

martin-schulze-vireso: Wow, this is bad: I got no error message, that the test could not be executed. I thought everything worked.

jasonkarns: double-high-five on including workflow_dispatch in the first place. I never fail to regret omitting it from a workflow.

martin-schulze-vireso:
Oh boy, those pesky parallel tests with their timing problems. I need to think a bit longer about getting them more robust.

martin-schulze-vireso:
I ran the last commit thrice to see if it remains green.

sublimino:
Getting a failure running in docker locally. I'm able to reproduce this fairly often, although sometimes it passes. 

I'm not much help I'm afraid, I can't get it to fail outside the container, and I can't figure out the race condition from this PR :/

```bash
not ok 131 extended syntax in suite
# (in test file /opt/bats/test/suite.bats, line 63)
#   `[ "${lines[4]}" = "suite $FIXTURE_ROOT/multiple/b.bats" ]' failed
# output: 1..3
# suite /opt/bats/test/fixtures/suite/multiple/a.bats
# begin 1 truth
# ok 1 truth
# begin 2 truth
# ok 2 truth
# suite /opt/bats/test/fixtures/suite/multiple/b.bats
# begin 2 more truth
# ok 2 more truth
# begin 3 quasi-truth
# not ok 3 quasi-truth
# # (in test file /opt/bats/test/fixtures/suite/multiple/b.bats, line 6)
# #   `[ -z "$FLUNK" ]' failed
# begin 4 more truth
# ok 4 more truth
# begin 5 quasi-truth
# not ok 5 quasi-truth
# # (in test file /opt/bats/test/fixtures/suite/multiple/b.bats, line 6)
# #   `[ -z "$FLUNK" ]' failed
```

martin-schulze-vireso:
Which command exactly are you running?

martin-schulze-vireso:
Okay, so it looks like the test "truth" gets executed twice, which seems very unusual.

Unfortunately, I cannot reproduce the issue you described above. I tried the steps from the bash-version tests for versions 3,4 and 5 with and without `-e BATS_NUMBER_OF_PARALLEL_JOBS=2` but could not reproduce the issue. Can you give more details like:

- used container
- bash version
- exact run command
- checked out git commit

@sublimino ?

sublimino:
Using the Dockerfile here, the arg `bash:latest` which is v5.

- Cloned this PR and build the latest image `docker build --tag bats/bats:latest .`
- `docker run -t -v /tmp/:/tmp/ bats/bats:latest /opt/bats/test --tap -f 'extended syntax in suite'`

About ~70% of the time this results in an ordering error, and the ordering generally differs per-run.

I've noticed that mounting the host `/tmp` directory to the container causes the error, could it be a race condition introduced by filesystem latency? I've been running containerised bats in this way for a while and it's previously been ok.

Do the above commands reproduce it for you @martin-schulze-vireso?

```bash
1..1                                                                                           
not ok 1 extended syntax in suite                                                              
# (in test file /opt/bats/test/suite.bats, line 63)                                            
#   `[ "${lines[4]}" = "suite $FIXTURE_ROOT/multiple/b.bats" ]' failed                         
# output: 1..3                                                                                 
# suite /opt/bats/test/fixtures/suite/multiple/a.bats                                          
# begin 1 truth                                                                                
# ok 1 truth                                                                                   
# begin 2 truth                                                                                
# ok 2 truth                                                                                   
# begin 3 truth                                                                                
# ok 3 truth                                                                                   
# begin 4 truth                                                                                
# ok 4 truth                                                                                   
# suite /opt/bats/test/fixtures/suite/multiple/b.bats                                          
# begin 2 more truth                                                                           
# ok 2 more truth                                                                              
# begin 3 quasi-truth                                                                          
# not ok 3 quasi-truth                                                                         
# # (in test file /opt/bats/test/fixtures/suite/multiple/b.bats, line 6)                       
# #   `[ -z "$FLUNK" ]' failed                                                                 
# begin 4 more truth                                                                           
# ok 4 more truth                                                                              
# begin 5 quasi-truth                                                                          
# not ok 5 quasi-truth                                                                         
# # (in test file /opt/bats/test/fixtures/suite/multiple/b.bats, line 6)                       
# #   `[ -z "$FLUNK" ]' failed                                                                 
# begin 6 more truth                                                                           
# ok 6 more truth                                                                              
# begin 7 quasi-truth                                                                          
# not ok 7 quasi-truth                                                                         
# # (in test file /opt/bats/test/fixtures/suite/multiple/b.bats, line 6)                       
# #   `[ -z "$FLUNK" ]' failed                                                                 
# begin 8 more truth                                                                           
# ok 8 more truth                                                                              
# begin 9 quasi-truth                                                                          
# not ok 9 quasi-truth                                                                         
# # (in test file /opt/bats/test/fixtures/suite/multiple/b.bats, line 6)                       
# #   `[ -z "$FLUNK" ]' failed
```

martin-schulze-vireso:
Now I can reproduce the issue. However, it is not related to this PR, its already on master!
I already got a hunch this might be related but have a look at #409. The mounting of /tmp ensures persistence (which I did not have in my tests), while running under docker ensures equal PIDs each time the test run is started.

Now, `BATS_RUN_TMPDIR=$BATS_TMPDIR/bats-run-$BATS_ROOT_PID` can lead to collisions.


Under normal circumstances this directory is deleted and we would get an error for a preexisting `BATS_RUN_TMPDIR` and the directory is deleted at the end of each run, both inside `libexec/bats-core/bats`. However, in the failing test we use `bats-exec-suite` directly and the function `emulate_bats_env` sets up the directory, so neither the preexisting directory is detected nor deleted.

@sublimino Since this issue predates this PR and is fixed by #409, I would propose to ignore it during reviews for this PR.

sublimino:
Ah excellent, thanks for the investigation @martin-schulze-vireso! The bug is indeed fixed in #409 so I'm happy to take your direction and exclude that from this test. 

On that basis this is good to go from my perspective, everything else passes deterministically, and I'll merge now :pray:

PR #406, opened 2021-02-07
Publish to npm/GHPR

jasonkarns:
Adds workflow to publish as an npm package to the npmjs registry and github package registry.
The workflow has been tested and works (pre-release alphas have been published to both registries).
This also incorporates a host of administrative changes regarding orgs and permissions on npm.

Remaining todo: determine what the release workflow will actually be and change the trigger on this workflow accordingly. My assumption is that we'll want the creation of a Release on github to be manual (while we still manually create release notes, etc). That is to say, we probably _don't_ want the release process to be wholly automated off of, say, a git tag event. (yet)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,15 @@
+name: Release
+
+on: [push]
+
+jobs:
+  github-npm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          registry-url: "https://npm.pkg.github.com"
+      - run: npm publish --ignore-scripts
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,8 @@
 name: Release
 
-on: [push]
+on:
+  release: { types: [published] }
+  workflow_dispatch:
 
 jobs:
   npmjs:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+
+  github-npm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          registry-url: "https://npm.pkg.github.com"
+      - name: scope package name as required by GHPR
+        run: npm init -y --scope ${{ github.repository_owner }}
```
martin-schulze-vireso, line 27: I have no npm experience, why is this necessary here, but not in the other job?

martin-schulze-vireso (approved):
I am not familiar with npm but the GHA workflow looks good to me.

.github/workflows/release.yml
jasonkarns: On npm, the package is published unscoped (as `bats`). The GHPR, on the other hand, doesn't support or allow unscoped packages. The package name, therefore, must be scoped to the organization: `@bats-core/bats`.

This was the simplest way I could think of to change the package.json such that the package name is scoped prior to publishing to GHPR.

An alternative would be to deprecate the unscoped `bats` package and publish it scoped on both registries. That might be something we should consider.

martin-schulze-vireso: I'd say we can defer this discussion to get v1.3.0 out the door. Maybe you want to open an issue for that?

PR #409, opened 2021-02-12
Use mktemp

rockandska:

Using mktemp is safer than using $PID and avoid race condition, especially when running inside docker because the PID is rarely != 1.
That PR allows to running multiples times the same tests safely with `--no-tempdir-cleanup` and avoid race condition too.
Without this PR we hit on the second run and is quite annoying :

```
Error: BATS_RUN_TMPDIR (/tmp/bats-run-1) already exists
Reusing old run directories can lead to unexpected results ... aborting!
```

As an example for debugging :
```bash
$ cat test.sh
mkdir -p "$PWD/tmp"
for BASHVER in 3.2 4.1 4.4;do
	docker build --build-arg bashver="${BASHVER}" --tag "bats/bats:bash-${BASHVER}" . || exit $?
	docker run -it "bash:${BASHVER}" --version || exit $?
	time docker run \
          -u "$(id -u):0" \
          -it \
          -v "/etc/passwd:/etc/passwd:ro" \
          -v "${PWD}/tmp:/tmp" \
          "bats/bats:bash-${BASHVER}" \
          --tap /opt/bats/test \
          --no-tempdir-cleanup \
          ${@:+--filter "${@}"} \
          || exit $?
done
$ ./test.sh '.*tmpdir.*'
```
Result after some run of bats core self tests :
```
$ ls -1 tmp/
bats-run-1-Jamgap
bats-run-1-KGLIOm
bats-run-1-kJLIJD
bats-run-1-NdbhKe
bats-run-1-npAgfd
bats-run-1-oklMGC
bats-run-test-tmpdir-9354-DPMCIb
bats-run-test-tmpdir-9354-kgppIg
bats-run-test-tmpdir-9354-ooIgGd
bats-run-test-tmpdir-9355-EkNegL
bats-run-test-tmpdir-9410-fbLkJa
bats-run-test-tmpdir-9410-gBFNEK
bats-run-test-tmpdir-9410-PBifMi
bats-run-test-tmpdir-9411-kclEOi
bats-run-test-tmpdir-9481-dPLPNB
bats-run-test-tmpdir-9481-elDdeH
bats-run-test-tmpdir-9481-nLAEbE
bats-run-test-tmpdir-9482-DEFDdi
bats-run-test-tmpdir-9540-cbCCdl
bats-run-test-tmpdir-9542-ngkhMF
bats-run-test-tmpdir-9597-mFaEac
bats-run-test-tmpdir-9599-lEaffJ
bats-run-test-tmpdir-9669-cpmdHK
bats-run-test-tmpdir-9671-CaipNk
```

### DIFF
--- a/docs/source/writing-tests.md
+++ b/docs/source/writing-tests.md
@@ -305,6 +305,9 @@ There are several global variables you can use to introspect on Bats tests:
 - `$BATS_TEST_NUMBER` is the (1-based) index of the current test case in the test file.
 - `$BATS_SUITE_TEST_NUMBER` is the (1-based) index of the current test case in the test suite (over all files).
 - `$BATS_TMPDIR` is the location to a directory that may be used to store temporary files.
+- `$BATS_RUN_TMPDIR` is the location to the temporary directory used by bats to
+   store all its internal temporary files during the tests.
+   (default: `$BATS_TMPDIR/bats-run-$BATS_ROOT_PID-XXXXXX`)
 - `$BATS_FILE_EXTENSION` (default: `bats`) specifies the extension of test files that should be found when running a suite (via `bats [-r] suite_folder/`)
 
 ### Libraries and Add-ons
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -81,7 +81,7 @@ if [[ -z "${TMPDIR-}" ]]; then
 else
   export BATS_TMPDIR="${TMPDIR%/}"
 fi
-export BATS_RUN_TMPDIR="$BATS_TMPDIR/bats-run-$BATS_ROOT_PID"
+export BATS_RUN_TMPDIR=
 
 arguments=()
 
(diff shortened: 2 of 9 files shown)


libexec/bats-core/bats
```diff
@@ -200,12 +200,19 @@ while [[ "$#" -ne 0 ]]; do
   shift
 done
 
-if [[ -d "$BATS_RUN_TMPDIR" ]]; then
-  printf "Error: BATS_RUN_TMPDIR (%s) already exists\n" "$BATS_RUN_TMPDIR" >&2
-  printf "Reusing old run directories can lead to unexpected results ... aborting!\n" >&2
-  exit 1
+if [[ -n "${BATS_RUN_TMPDIR:-}" ]];then
```
martin-schulze-vireso, line 203: So far, we don't allow for setting `BATS_RUN_TMPDIR` from outside. It is already customizable via `BATS_TMPDIR`, so I think we don't need this additional complexity here.

btamayo: Do you think having this check (for `-n "${BATS_RUN_TMPDIR:-}") here anyway could potentially expose and add a guardrail against race conditions?

docs/source/writing-tests.md
```diff
@@ -305,6 +305,9 @@ There are several global variables you can use to introspect on Bats tests:
 - `$BATS_TEST_NUMBER` is the (1-based) index of the current test case in the test file.
 - `$BATS_SUITE_TEST_NUMBER` is the (1-based) index of the current test case in the test suite (over all files).
 - `$BATS_TMPDIR` is the location to a directory that may be used to store temporary files.
+- `$BATS_RUN_TMPDIR` is the location to the temporary directory used by bats to
```
btamayo, line 308: I propose ([here](https://github.com/bats-core/bats-core/pull/410/files#r575897751) as well) we move internal variables to a separate section to minimize folks using them without breaking existing logic.

libexec/bats-core/bats
martin-schulze-vireso: It would allow an opportunity for better reporting in cases where it is detectable but I don't think it can eliminate race conditions entirely, since the check and `mktemp` are not one atomic operation.

btamayo:
I think we can add this to 1.3.1. @rockandska Thank you for this. Can you resolve the conflicts and address the comment(s). Otherwise, LGTM.

rockandska:
@martin-schulze-vireso / @btamayo :

- add an error and exit if mktemp failed
- moved BATS_RUN_TMPDIR to an `internal variables section`
- rebased with master

martin-schulze-vireso (approved):

PR #410, opened 2021-02-12
BATS_TMPDIR set order

rockandska:

- Follow the ideas from #365 regarding the `$BATS_TMPDIR` init.
  - Use `BATS_TMPDIR` if set
  - Use `$TMPDIR` if `$BATS_TMPDIR` is not set
  - Use /tmp if `$TMPDIR` and `$BATS_TMPDIR` are not set

- Changed the doc according to the init process
- Add tests for those changes

### DIFF
--- a/docs/source/writing-tests.md
+++ b/docs/source/writing-tests.md
@@ -304,7 +304,9 @@ There are several global variables you can use to introspect on Bats tests:
 - `$BATS_TEST_DESCRIPTION` is the description of the current test case.
 - `$BATS_TEST_NUMBER` is the (1-based) index of the current test case in the test file.
 - `$BATS_SUITE_TEST_NUMBER` is the (1-based) index of the current test case in the test suite (over all files).
-- `$BATS_TMPDIR` is the location to a directory that may be used to store temporary files.
+- `$BATS_TMPDIR` is the base temporary directory used by bats to create its
+   temporary files / directories.
+   (default: `$TMPDIR`. If `$TMPDIR` is not set, `/tmp` is used.)
 - `$BATS_RUN_TMPDIR` is the location to the temporary directory used by bats to
    store all its internal temporary files during the tests.
    (default: `$BATS_TMPDIR/bats-run-$BATS_ROOT_PID-XXXXXX`)
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -76,12 +76,16 @@ export BATS_TEST_PATTERN_COMMENT="[[:blank:]]*([^[:blank:]()]+)[[:blank:]]*\(?\)
 export BATS_TEST_FILTER=
 export PATH="$BATS_LIBEXEC:$PATH"
 export BATS_ROOT_PID=$$
-if [[ -z "${TMPDIR-}" ]]; then
-  export BATS_TMPDIR="/tmp"
-else
-  export BATS_TMPDIR="${TMPDIR%/}"
-fi
+export BATS_TMPDIR="${TMPDIR:-/tmp}"
 export BATS_RUN_TMPDIR=
+
+if [[ ! -d "${BATS_TMPDIR}" ]];then
+  printf "Error: BATS_TMPDIR (%s) do not exist or is not a directory" "${BATS_TMPDIR}" >&2
+  exit 1
+elif [[ ! -w "${BATS_TMPDIR}" ]];then
+  printf "Error: BATS_TMPDIR (%s) is not writable" "${BATS_TMPDIR}" >&2
+  exit 1
+fi
 
 arguments=()
 
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -80,7 +80,7 @@ export BATS_TMPDIR="${TMPDIR:-/tmp}"
 export BATS_RUN_TMPDIR=
 
 if [[ ! -d "${BATS_TMPDIR}" ]];then
-  printf "Error: BATS_TMPDIR (%s) do not exist or is not a directory" "${BATS_TMPDIR}" >&2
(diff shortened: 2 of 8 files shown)


docs/source/writing-tests.md
```diff
@@ -304,7 +304,9 @@ There are several global variables you can use to introspect on Bats tests:
 - `$BATS_TEST_DESCRIPTION` is the description of the current test case.
 - `$BATS_TEST_NUMBER` is the (1-based) index of the current test case in the test file.
 - `$BATS_SUITE_TEST_NUMBER` is the (1-based) index of the current test case in the test suite (over all files).
-- `$BATS_TMPDIR` is the location to a directory that may be used to store temporary files.
+- `$BATS_TMPDIR` is the base temporary directory used by bats to create its
```
martin-schulze-vireso, line 307: As expounded in https://github.com/bats-core/bats-core/pull/365#issuecomment-778841245, I would propose to just mark `BATS_TMPDIR` as deprecated. Users should either go for `TMPDIR` or `BATS_RUN_TMPDIR`.

rockandska: > As expounded in #365 (comment), I would propose to just mark BATS_TMPDIR as deprecated. 

Indeed, but keep the logic as it is in the time being and mark it as **deprecated**

> Users should either go for TMPDIR or BATS_RUN_TMPDIR.

Only `TMPDIR`.
For me, even `BATS_RUN_TMPDIR` should not be documented since it is an `internal` variable automatically generated from `BATS_TMPDIR/TMPDIR` as prefix. (could be discussed on the other PR)

> I propose to enforce this decision with failing if BATS_TMPDIR is set when starting bats. Since BATS_TMPDIR was set to TMPDIR//tmp unconditionally, we should only "break" users who erroneously believed they were overriding it.

Starting with a warning could be a first step before throwing an error. WDTY ?

libexec/bats-core/bats
```diff
@@ -76,13 +76,17 @@ export BATS_TEST_PATTERN_COMMENT="[[:blank:]]*([^[:blank:]()]+)[[:blank:]]*\(?\)
...
-if [[ -z "${TMPDIR-}" ]]; then
-  export BATS_TMPDIR="/tmp"
-else
-  export BATS_TMPDIR="${TMPDIR%/}"
-fi
+export BATS_TMPDIR="${BATS_TMPDIR:-${TMPDIR:-/tmp}}"
 export BATS_RUN_TMPDIR="$BATS_TMPDIR/bats-run-$BATS_ROOT_PID"
 
+if [[ ! -d "${BATS_TMPDIR}" ]];then
+  printf "Error: BATS_TMPDIR (%s) do not exist or is not a directory" "${BATS_TMPDIR}" >&2
```
btamayo, line 83: Tiny grammar nit: 

```suggestion
  printf "Error: BATS_TMPDIR (%s) does not exist or is not a directory" "${BATS_TMPDIR}" >&2
```

test/bats.bats
```diff
@@ -733,4 +733,21 @@ END_OF_ERR_MSG
...
-}
\ No newline at end of file
+}
+
+@test "Fail if BATS_TMPDIR do not exist or is not writable" {
+  local BATS_TMPDIR
+  BATS_TMPDIR=$(mktemp -u "${BATS_RUN_TMPDIR}/donotexist.XXXXXX")
+  run bats "$FIXTURE_ROOT/BATS_TMPDIR.bats"
+  [ "$status" -eq 1 ]
+  [ "${lines[0]}" = "Error: BATS_TMPDIR (${BATS_TMPDIR}) do not exist or is not a directory" ]
```
btamayo, line 743: Same grammar nit as above:
```suggestion
  [ "${lines[0]}" = "Error: BATS_TMPDIR (${BATS_TMPDIR}) does not exist or is not a directory" ]
```

docs/source/writing-tests.md
btamayo: I agree with keeping the logic and marking it as deprecated before removing it. Although it really does seem like `BATS_TMPDIR` is unconditionally set to `TMPDIR`/`/tmp`, there's a part of me that's leaning towards safe than sorry in this case.

Additionally, maybe we should move the internal variables outside of this list and perhaps under a new "Internal Variables" subsection? Or maybe we can at least mark which vars are internal and not meant for user config here in this list. (In the future, we really should have better conventions around internal variables - at least a naming/casing convention to differentiate between globals/externally-configured.)

test/bats.bats
```diff
@@ -733,4 +733,21 @@ END_OF_ERR_MSG
 
   # should also find preprocessed files!
   [ $(find "$TEST_TMPDIR" -name '*.src' | wc -l) -eq 1 ]
-}
\ No newline at end of file
+}
+
+@test "Fail if BATS_TMPDIR do not exist or is not writable" {
```
btamayo, line 738:
```suggestion
@test "Fail if BATS_TMPDIR does not exist or is not writable" {
```

martin-schulze-vireso:
@rockandska I have reverted the `BATS_TMPDIR` handling to unconditional initialization from TMPDIR or /tmp.

martin-schulze-vireso:
Obviously, the tests are  broken now. I think that I found the place where you wanted to put the line you referenced in gitter @rockandska. The tests fail mostly because the semantics have changed. Are you sure that this line was there before? I did not find it in the history.

rockandska:
Fixed.

The [part](https://github.com/bats-core/bats-core/pull/410/files#diff-cd529682f31fb50b4e5e6ac2eb7f031126ad56def53f42c44b74eb6601ad5fe3L208-R220) for the error handling on creation of `BATS_RUN_TMPDIR` could be moved to a new PR if necessary.

martin-schulze-vireso (approved):

martin-schulze-vireso:
@rockandska I needed to rebase your commits to sign them.

PR #413, opened 2021-02-20
Add BATS_SUITE_TMPDIR / BATS_FILE_TMPDIR / BATS_TEST_TMPDIR

rockandska:

BATS_SUITE_TMPDIR : tmp directory specific to the suite
BATS_FILE_TMPDIR : tmp directory specific to the test file
BATS_TEST_TMPDIR : tmp directory specific to the test

Allow the users to use those directories to store files necessary in tests depending on if they want to share them across suite, test file, tests or not.

Solve #283 

To summarize :
```
BATS_TMPDIR
  |_ BATS_RUN_TMPDIR
    |_ BATS_SUITE_TMPDIR
      |_ BATS_FILE_TMPDIR
        |_ BATS_TEST_TMPDIR
```

### DIFF
--- a/libexec/bats-core/bats-exec-file
+++ b/libexec/bats-core/bats-exec-file
@@ -49,6 +49,12 @@ fi
 
 BATS_TEST_FILENAME="$filename"
 
+if ! BATS_FILE_TMPDIR=$(mktemp -d "${BATS_SUITE_TMPDIR}/file-tmpdir-$(basename "${BATS_TEST_FILENAME}")-XXXXXX");then
+  printf '%s\n' "Failed to create BATS_FILE_TMPDIR with mktemp" >&2
+  exit 1
+fi
+export BATS_FILE_TMPDIR
+
 # shellcheck source=lib/bats-core/preprocessing.bash
 # shellcheck disable=SC2153
 source "$BATS_ROOT/lib/bats-core/preprocessing.bash"
--- a/libexec/bats-core/bats-exec-suite
+++ b/libexec/bats-core/bats-exec-suite
@@ -120,6 +120,12 @@ if [[ "${test_count}" == 0 ]]; then
   exit
 fi
 
+if ! BATS_SUITE_TMPDIR=$(mktemp -d "${BATS_RUN_TMPDIR}/suite-tmpdir-XXXXXX");then
+  printf '%s\n' "Failed to create BATS_SUITE_TMPDIR with mktemp" >&2
+  exit 1
+fi
+export BATS_SUITE_TMPDIR
+
 # Deduplicate filenames (without reordering) to avoid running duplicate tests n by n times.
 # (see https://github.com/bats-core/bats-core/issues/329)
 # If a file was specified multiple times, we already got it repeatedly in our TESTS_LIST_FILE.
--- a/libexec/bats-core/bats-exec-test
+++ b/libexec/bats-core/bats-exec-test
@@ -43,6 +43,12 @@ elif [[ ! -f "$BATS_TEST_FILENAME" ]]; then
   exit 1
 fi
 
+if ! BATS_TEST_TMPDIR=$(mktemp -d "${BATS_FILE_TMPDIR}/test-tmpdir-${BATS_TEST_NAME}-XXXXXX");then
+  printf '%s\n' "Failed to create BATS_TEST_TMPDIR with mktemp" >&2
+  exit 1
+fi
+export BATS_TEST_TMPDIR
+
 # load the test helper functions like `load` or `run` that are needed to run a (preprocessed) .bats file without bash errors
 # shellcheck source=lib/bats-core/test_functions.bash disable=SC2153
 source "$BATS_ROOT/lib/bats-core/test_functions.bash"
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -3,11 +3,6 @@
 load test_helper
 fixtures bats
 
-teardown() {
-  # cleanup the test local tmpdir to avoid cleaning up all tests' at once
-  test_helper::cleanup_tmpdir "$BATS_TEST_NAME"
-}
-
 @test "no arguments prints message and usage instructions" {
   run bats
   [ $status -eq 1 ]
@@ -143,7 +138,7 @@ teardown() {
 }
 
 @test "setup is run once before each test" {
-  make_bats_test_suite_tmpdir "$BATS_TEST_NAME"
+  export BATS_TEST_SUITE_TMPDIR="${BATS_TEST_TMPDIR}"
   run bats "$FIXTURE_ROOT/setup.bats"
   [ $status -eq 0 ]
   run cat "$BATS_TEST_SUITE_TMPDIR/setup.log"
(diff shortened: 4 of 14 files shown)


martin-schulze-vireso:
Currently, there is only one suite per run. This makes the suite and run tmpdir somewhat redundant, so I'd say we should avoid one of them. Unfortunately, I'd find suite, file, test more uniform than run, file, test. However, the `BATS_RUN_TMPDIR` might already have users because it has been released, albeit undocumented.

rockandska:
> Currently, there is only one suite per run.

Indeed.

> This makes the suite and run tmpdir somewhat redundant, so I'd say we should avoid one of them.

BATS_RUN_TMPDIR ans BATS_SUITE_TMPDIR have different purpose. BATS_RUN_TMPDIR is intend to be used internally by bats (store SRC, logs, etc..) when BATS_SUITE_TMPDIR is intend to be used by users in tests (will be used in a future `before_all` functionality) to store temp files common to all tests.

> Unfortunately, I'd find suite, file, test more uniform than run, file, test. However, the `BATS_RUN_TMPDIR` might already have users because it has been released, albeit undocumented.

The actual design in this PR will not break tests if users actually using it.

martin-schulze-vireso:
> BATS_RUN_TMPDIR ans BATS_SUITE_TMPDIR have different purpose. BATS_RUN_TMPDIR is intend to be used internally by bats (store SRC, logs, etc..) when BATS_SUITE_TMPDIR is intend to be used by users in tests (will be used in a future `before_all` functionality) to store temp files common to all tests.

Ahh, that makes sense. I thought they all were for internal usage only. Are you aware of https://github.com/bats-core/bats-file#working-with-temporary-directories? There seems to be some overlap.

rockandska:
> Are you aware of https://github.com/bats-core/bats-file#working-with-temporary-directories? There seems to be some overlap.

Was not aware of `temp_make` but not a big fan of this for those reasons :
- users are the ones who need to create them (more code to write)
- users are the ones who need to manage the deletion (more code to write)
- users need to use additional env var to preserve temporary files (this PR implementation use the command-line parameter already documented and available in `--help` of bats and keep a logic directory structure as describe in the PR description)
- users need to add additional library totally uncorrelated from the core

libexec/bats-core/bats-exec-file
```diff
@@ -49,6 +49,12 @@ fi
 
 BATS_TEST_FILENAME="$filename"
 
+if ! BATS_FILE_TMPDIR=$(mktemp -d "${BATS_SUITE_TMPDIR}/bats-file-tmpdir-XXXXXX");then
```
martin-schulze-vireso, line 52: I think it would be easier for debugging/navigation to have the filename as part of the tmpdir. Maybe `${BATS_SUITE_TMPDIR}/file-tmpdir-$(basename "${BATS_FILENAME}")-XXXX`.

libexec/bats-core/bats-exec-test
```diff
@@ -43,6 +43,12 @@ elif [[ ! -f "$BATS_TEST_FILENAME" ]]; then
   exit 1
 fi
 
+if ! BATS_TEST_TMPDIR=$(mktemp -d "${BATS_FILE_TMPDIR}/bats-test-tmpdir-XXXXXX");then
```
martin-schulze-vireso, line 46: Likewise to the comment on BATS_FILE_TMPDIR it is more readable to have `${BATS_FILE_TMPDIR}/test-tmpdir-${BATS_TESTNAME}-XXXX`.

libexec/bats-core/bats-exec-suite
```diff
@@ -120,6 +120,12 @@ if [[ "${test_count}" == 0 ]]; then
   exit
 fi
 
+if ! BATS_SUITE_TMPDIR=$(mktemp -d "${BATS_RUN_TMPDIR}/bats-suite-tmpdir-XXXXXX");then
```
martin-schulze-vireso, line 123: I think the `$BATS_RUN_TMPDIR` prefix gives us enough scope to omit the bats- prefix on lower levels, so `${BATS_RUN_TMPDIR}/suite-tmpdir-XXXXX` should be enough.

test/junit-formatter.bats
```diff
@@ -103,8 +102,7 @@ TESTSUITES_REGEX="<testsuites time=\"$FLOAT_REGEX\">"
 }
 
 @test "junit formatter as report formatter creates report.xml" {
-  make_bats_test_suite_tmpdir
-  cd "$BATS_TEST_SUITE_TMPDIR" # don't litter sources with output files
+  cd "$BATS_TEST_TMPDIR" # don't litter sources with output files
```
martin-schulze-vireso, line 105: I know it is not really part of this PR but could you also take care of the `tab in filename` test? It litters source with the transformed file and should be trivially fixable with the new tmpdirs.

martin-schulze-vireso (commented):
Only some minor issues.

test/junit-formatter.bats
rockandska: I don't get what you would like to see to be done.
If it is not related directly to this PR, why not address this in another one ?
Always better to split PR for different purpose

test/bats.bats
```diff
@@ -656,8 +650,7 @@ END_OF_ERR_MSG
 }
```
martin-schulze-vireso, line 650:
```suggestion
  cp "$BATS_TEST_TMPDIR/tab in filename.bats" "$FIXTURE_ROOT/tab"$'\t'"in filename.bats"
  bats "$BATS_TEST_TMPDIR/tab"$'\t'"in filename.bats"
}
```

I meant to change this, which is very similar to your change below and profits from your newly added variable.

martin-schulze-vireso, line 650: I meant you to change this, which is very similar to your change below and profits from your newly added variable:

```
  cp "$BATS_TEST_TMPDIR/tab in filename.bats" "$FIXTURE_ROOT/tab"$'\t'"in filename.bats"
  bats "$BATS_TEST_TMPDIR/tab"$'\t'"in filename.bats"
}
```

test/junit-formatter.bats
martin-schulze-vireso: See my new comment. I think it is very fitting for this PR.

test/bats.bats
rockandska: Should be fixed now

rockandska:
My last concern, and seems directly related to the tests themself, is some tmp directory still here at the of tests
```
ls -1 tmp/
bats-run-test-tmpdir-2081-bOOCIO
bats-run-test-tmpdir-2196-LlHPJF
bats-run-test-tmpdir-2254-mKmbdo
bats-run-test-tmpdir-9566-OcLJnk
bats-run-test-tmpdir-9633-NNJAin
bats-run-test-tmpdir-9715-MLHeea
tmp.gfkgam
tmp.mgfpCl
tmp.nDgkNa
```

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for the cleanup. I'll need to sign these commits to merge them. If you can, please do so yourself to reduce barriers.

rockandska:
> Thanks for the cleanup. I'll need to sign these commits to merge them. If you can, please do so yourself to reduce barriers.

done

PR #430, opened 2021-04-16
Install bats binary with npm install

felipecrs:
So npm install --global bats && bats --version is supposed to work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "author": "Sam Stephenson <sstephenson@gmail.com> (http://sstephenson.us/)",
   "repository": "github:bats-core/bats-core",
   "bugs": "https://github.com/bats-core/bats-core/issues",
+  "bin": "bin/bats",
   "files": [
     "bin",
     "libexec",


martin-schulze-vireso (approved):

jasonkarns:
What changed with this? Per the docs, `"directories": { "bin": "bin",` is supposed to create the bin symlink already.

jasonkarns:
Actually, this needs reverted. Not only is `bin` not supposed to be necessary, specifying _both_ is actually an error:

> If you specify a bin directory in directories.bin, all the files in that folder will be added.

> Because of the way the bin directive works, specifying both a bin path and setting directories.bin is an error. If you want to specify individual files, use bin, and for all the files in an existing bin directory, use directories.bin.

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#directoriesbin

felipecrs:
I'm really sorry about it. So it turns out that it's an issue in Volta.

https://github.com/volta-cli/volta/issues/978

martin-schulze-vireso:
@jasonkarns I'll take that as a lesson to let my fingers off node related PRs. I read the documentation about bin and it sounded fitting. No mention about directories.bin there.

felipecrs:
@martin-schulze-vireso @jasonkarns https://github.com/bats-core/bats-core/pull/435

PR #452, opened 2021-06-10
install.sh: consider multilib

kkang-wr:
It may install files to /usr/lib64 when multilib is enabled. Add an
optional parameter for install.sh to support it.

Signed-off-by: Kai Kang <kai.kang@windriver.com>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,6 +4,7 @@ set -e
 
 BATS_ROOT="${0%/*}"
 PREFIX="$1"
+LIBDIR="${2:-lib}"
 
 if [[ -z "$PREFIX" ]]; then
   printf '%s\n' \
@@ -12,10 +13,10 @@ if [[ -z "$PREFIX" ]]; then
   exit 1
 fi
 
-install -d -m 755 "$PREFIX"/{bin,libexec/bats-core,lib/bats-core,share/man/man{1,7}}
+install -d -m 755 "$PREFIX"/{bin,libexec/bats-core,${LIBDIR}/bats-core,share/man/man{1,7}}
 install -m 755 "$BATS_ROOT/bin"/* "$PREFIX/bin"
 install -m 755 "$BATS_ROOT/libexec/bats-core"/* "$PREFIX/libexec/bats-core"
-install -m 755 "$BATS_ROOT/lib/bats-core"/* "$PREFIX/lib/bats-core"
+install -m 755 "$BATS_ROOT/lib/bats-core"/* "$PREFIX/${LIBDIR}/bats-core"
 install -m 644 "$BATS_ROOT/man/bats.1" "$PREFIX/share/man/man1"
 install -m 644 "$BATS_ROOT/man/bats.7" "$PREFIX/share/man/man7"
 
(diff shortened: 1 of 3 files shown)


martin-schulze-vireso:
Please add a test for this feature.

install.sh
```diff
@@ -4,6 +4,7 @@ set -e
 
 BATS_ROOT="${0%/*}"
 PREFIX="$1"
+BASELIB="${2:-lib}"
```
martin-schulze-vireso, line 7: If we only want to support lib and lib64, shouldn't we just check the existence of lib vs lib64 and choose automatically?

martin-schulze-vireso (changes requested):
Please add a test for this feature and document its usage on line 11-12

install.sh
kkang-wr: For cross compile, it can not check the directory on build machine and requires to pass the libdir to the script. BTW, this patch breaks some existing script and I'll send v2. Thanks.

martin-schulze-vireso: Do you have an ETA for that?

kkang-wr: It turns out the failure I mean is not bats related, so this version of the patch is enough. But I am not familiar with writing bats test cases. Is there any doc about it, please? Thanks.

martin-schulze-vireso: I think you can get inspiration from the other tests in test/install.bats. Aditionally, you can view the normal documentation on readthedocs.

kkang-wr: @martin-schulze-vireso I copied and updated a test case in install.bats. Would you like to have a look at it, please? Thanks.

martin-schulze-vireso (changes requested):
The actual lib folder is not checked. This is already a shortcoming of the test you used as a source for copying. Could you fix both please?

kkang-wr:
> The actual lib folder is not checked. This is already a shortcoming of the test you used as a source for copying. Could you fix both please?

Done.

martin-schulze-vireso (approved):
I added the missing tests for successful removal of the lib folder which uncovered that uninstall.sh did not allow for configuring the lib folder. Fixed that too.

kkang-wr:
> I added the missing tests for successful removal of the lib folder which uncovered that uninstall.sh did not allow for configuring the lib folder. Fixed that too.

Thanks.

martin-schulze-vireso:
Did you actually try to use this on an existing system @kkang-wr ? The internal code relies on a hardcoded path from `libexec/bats-core/bin` to `lib/bats-core/*`, so this failed for me in a bash docker container:

```
bash-4.4# ./install.sh /usr lib64
Installed Bats to /usr/bin/bats
bash-4.4# bats /bats-core/test
/usr/libexec/bats-core/bats: line 300: /usr/lib/bats-core/validator.bash: No such file or directory
```

The install script would actually have tell the `bats` binary about the changed lib path for this to work. This is basically the same issue as with `uninstall.sh` not knowing about the correct lib folder.

kkang-wr:
It is my fault. I'll fix it ASAP.

kkang-wr:
@martin-schulze-vireso I submiited a pr #487 to fix the issue. Would you like to review it, pleas?

Issue #462, opened 2021-07-02
Failure of negated statement not caught (`! command` never fails)


ad8-bdl:
# Summary
Commands within a test that use bash's negation operator `!` and that should fail (e.g. `! true`) don't cause the test to fail.

If such a test just happens to be the last operation in the test, that operation's exit status will propagate through to the test's return code, and _that_ will trigger the test to fail.

A workaround is to suffix affected commands with `|| false`, e.g. `! true || false`.

# Repro

bash-5.1$ bats --version
Bats 1.3.0
bash-5.1$ bash --version
GNU bash, version 5.1.8(1)-release (x86_64-apple-darwin19.6.0)
bash-5.1$ sw_vers 
ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H1217

- bash installed via brew:
bash-5.1$ brew list --versions bash
bash 5.1.8

Bats script:
```
# not working as expected:
@test "tf1" {
  true  # test failure is incorrectly attributed to this line
  ! true
} # test fails, but only due to the failing command being the final statement

@test "tf2" {
  ! true  # test should fail here, but doesn't
  true
} # test passes (wrong)

# working as expected:
@test "tf3" {
  true
  ! true || false    # test fails at this line
} 

@test "tf4" {
  ! true || false    # test fails at this line
  true
}
```

Actual Results:
```
bash-5.1$ bats /tmp/t.bats 
 ✗ tf1
   (in test file /tmp/t.bats, line 3)
     `true  # test failure is incorrectly attributed to this line' failed
 ✓ tf2
 ✗ tf3
   (in test file /tmp/t.bats, line 15)
     `! true || false    # test fails at this line' failed
 ✗ tf4
   (in test file /tmp/t.bats, line 19)
     `! true || false    # test fails at this line' failed

4 tests, 3 failures
```

Expected results:
- all tests tf[1-4] should fail
- the failure should accurately report the failing command

martin-schulze-vireso:
Unfortunately, that seems to be the intended behavior by bash:

```
 -e      Exit immediately if a pipeline (which may consist of a single simple command), a list, or a compound command (see SHELL GRAMMAR above), exits with a non-zero sta‐
                      tus.  The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test following the
                      if or elif reserved words, part of any command executed in a && or || list except the command following the final && or ||, any command  in  a  pipeline  but  the
[...]

PR #463, opened 2021-07-05
FAQ: add entry for negation `!`

ad8-bdl:
#462

- [ x] I have reviewed the [Contributor Guidelines][contributor].
- [ x] I have reviewed the [Code of Conduct][coc] and agree to abide by it

### DIFF
--- a/docs/source/gotchas.rst
+++ b/docs/source/gotchas.rst
@@ -9,6 +9,14 @@ convention of using return code 0 to signify success and everything non-zero to 
 
 Please adhere to this idiom while using bats, or you will constantly work against your environment.
 
+My negated statement (e.g. ! true) does not fail the test, even when it should.
+-------------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


martin-schulze-vireso (changes requested):
Thanks for writing this up. I have following suggestions:

1. Make it part of the Gotchas instead
2. recommend to use `run` and check for `[ $status != 0 ]`

In one of the next releases we will get a run function that does the checking itself, so the second line will become obsolete then. Our goal is to unify execution under the run umbrella.

docs/source/gotchas.rst
```diff
@@ -9,6 +9,15 @@ convention of using return code 0 to signify success and everything non-zero to
 
 Please adhere to this idiom while using bats, or you will constantly work against your environment.
 
+Can I use bash's negation operator `!`?
```
martin-schulze-vireso, line 12: Please formulate this as a problem like: 

> My negated statement (e.g. `! x`) does not fail the test, even when it should.

martin-schulze-vireso (approved):

martin-schulze-vireso:
I rebased to sign and squashed with a slightly rephrased commit message. Thanks again for your input.

PR #444, opened 2021-05-11
Add a $command variable, and colorize summary

msterin:
Fixes #163 

Motivation:
* See https://github.com/bats-core/bats-core/issues/163 for $command motivation. I second it and also it's helpful when the command passed to `run` contains multiple shell vars, so $command shows the final.
* Our test pass is fairly long and it's super convenient to have colorized summary (pass/fail) at the end of manual runs

We use bats with these fixes for a long time now, time for a PR.

- [x] I have reviewed the [Contributor Guidelines][contributor].
- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md
[coc]:         https://github.com/bats-core/bats-core/blob/master/docs/CODE_OF_CONDUCT.md

PR #467, opened 2021-07-19
Verbose output for faster troubleshooting of test problems

NorseGaud:
https://github.com/bats-core/bats-core/pull/137 seems to be dead.

Goal: Provide a better way to see exactly what was run and any STDERR/OUT for the commands

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,14 @@ name: Tests
 on: [push, pull_request, workflow_dispatch]
 
 jobs:
+  changelog:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check that PR is mentioned in Changelog
+        run:
+          grep "#${{github.event.pull_request.number}}" docs/CHANGELOG.md
+    if: ${{github.event.pull_request}}
   shellcheck:
     runs-on: ubuntu-20.04
     steps:
--- a/libexec/bats-core/bats-exec-file
+++ b/libexec/bats-core/bats-exec-file
@@ -291,8 +291,14 @@ if [[ -n "$extended_syntax" ]]; then
   printf "suite %s\n" "$filename"
 fi
 
-bats_preprocess_source "$filename"
 bats_read_tests_list_file
+# don't run potentially expensive setup/teardown_file
+# when there are no tests to run
+if [[ ${#tests_to_run[@]} -eq 0 ]]; then
+  exit 0
+fi
+
+bats_preprocess_source "$filename"
 bats_run_setup_file
 bats_run_tests
 bats_run_teardown_file
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,11 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+* don't run setup_file/teardown_file on files without tests, e.g. due to
+  filtering (#484)
+
 ## [1.4.1] - 2021-07-24
 
 ### Added
--- a/test/file_setup_teardown.bats
+++ b/test/file_setup_teardown.bats
@@ -125,7 +125,7 @@ not ok 1 failing test
 }
 
 @test "teardown_file runs even if all tests in the file are skipped" {
-  # shellcheck disable=SC2031
+  # shellcheck disable=SC2031,SC2030
   export LOG="$BATS_TEST_TMPDIR/teardown_file_skipped.log" 
   run bats "$FIXTURE_ROOT/teardown_file_even_if_all_tests_are_skipped.bats"
   [[ $status -eq 0 ]]
@@ -175,4 +175,14 @@ ok 2 must not see variable from first run" ]]
 
 @test "variables exported in setup_file are visible in tests" {
   [[ $SETUP_FILE_EXPORT_TEST == "true" ]]
+}
+
+@test "Don't run setup_file for files without tests" {
+  # shellcheck disable=SC2031
+  export LOG="$BATS_TEST_TMPDIR/setup_file.log"
+  # only select the test from no_setup_file
+  run bats -f test "$FIXTURE_ROOT/setup_file.bats" "$FIXTURE_ROOT/no_setup_file.bats"
+
+  [ ! -f "$LOG" ] # setup_file must not have been executed!
+  [ "${lines[0]}" == '1..1' ] # but at least one test should have been run
 }


.github/workflows/tests.yml
```diff
@@ -31,7 +32,10 @@ jobs:
           TERM: linux # fix tput for tty issue work around
         run: |
           bash --version
-          bash -c "time ${{ matrix.env_vars }} bin/bats --formatter tap test"
+          bash -c "time ${{ matrix.env_vars }} bin/bats --log --formatter tap test" || EC=$?
+          [[ -f ./bats.log ]] && cat ./bats.log
```
martin-schulze-vireso: I'd rather have this as a test Case in test/bats.bats

lib/bats-core/tracing.bash
```diff
@@ -72,18 +78,24 @@ bats_print_failed_command() {
...
 
 	if [[ "$BATS_ERROR_STATUS" -eq 1 ]]; then
-		printf 'failed\n'
+    output+="failed"
 	else
-		printf 'failed with status %d\n' "$BATS_ERROR_STATUS"
+    output+="failed with status $BATS_ERROR_STATUS"
 	fi
+  
+  echo "$output"
```
martin-schulze-vireso, line 95: Please use printf for all outputs in bats' bash files.

libexec/bats-core/bats
```diff
@@ -46,6 +46,7 @@ HELP_TEXT_HEADER
   -p, --pretty              Shorthand for "--formatter pretty"
   -r, --recursive           Include tests in subdirectories
   -t, --tap                 Shorthand for "--formatter tap"
+  -l, --log                 Output a log file containing STDOUT/ERR from each test and commands
```
martin-schulze-vireso, line 49: I find this moniker to generic for the usecase. How about --trace-log?

martin-schulze-vireso (commented):
This is somewhat similar to the trace PR. I like the idea of having a separate log file but there may be situations where we warnt the output in the test result.

My suggestion would to have two options --trace and --trace-log to decide where the output goes. 

Alternatively we could add FD 4 that is meant for trace output an can be redirected/merged at will. I'd have to think a bit about this.

libexec/bats-core/bats
NorseGaud: Hey! Thanks for the replies! Isn't `--trace` a different type of output? I tried the code in that PR thinking that it was what I wanted but it seems to output what bats is doing, not command STDOUT/ERR.

What about `--std-log`?

.github/workflows/tests.yml
NorseGaud: I can add tests once we can agree on what this should look like. I'd rather not spend the time just to have to change them :)

lib/bats-core/tracing.bash
NorseGaud: > This is somewhat similar to the trace PR. I like the idea of having a separate log file but there may be situations where we warnt the output in the test result.
> 
> My suggestion would to have two options --trace and --trace-log to decide where the output goes.
> 
> Alternatively we could add FD 4 that is meant for trace output an can be redirected/merged at will. I'd have to think a bit about this.

Having this output to STDOUT of bats would be awesome, but I can't seem to get it working. This is why I just resorted to using a log file.

libexec/bats-core/bats
NorseGaud: Or maybe `--command-std-log`

martin-schulze-vireso: The current trace PR is pretty old and I don't have looked too far into it. My current vision for trace feature would be to have several verbosity levels which add:

1. only commands
2. the outputs

Furthermore, I am hoping for a finegrained control via variables that allows to turn tracing on/off for specific section to avoid overloading the user with output.

The output from commands under `run` is not visible by default but stored in `$output`. Nevertheless, you should be able to print from there if you want it. Then the next hurdle is that output from passing tests is also not printed. You could use FD3 for that, but that has its problems too. I think if we want to have a trace/logging feature, we should also allow for verbosity that enables output even for passing tests.

martin-schulze-vireso:
I am wondering if it is worthwhile to separate the output for each test into its own file. This would make it easier to find the interesting parts.

libexec/bats-core/bats
NorseGaud: That's awesome, cause I found this too by working on docker-mailserver ;P

I'll take a look at what would be necessary to accomplish what you're thinking. I like the idea. 

What do you want to do with the older trace PR? It looks dead IMO.

NorseGaud:
> I am wondering if it is worthwhile to separate the output for each test into its own file. This would make it easier to find the interesting parts.

Absolutely! And, it would allow us to cat just the failing test file?

> I am wondering if it is worthwhile to separate the output for each test into its own file. This would make it easier to find the interesting parts.

It would allow for targetting the specific file to cat should it fail or the config set success STDOUT to be shown. I like that idea.

libexec/bats-core/bats
martin-schulze-vireso: I meant to revive it but it will definitely need some rework to fit into the current framework. I am afraid that this will still take some time. I am currently working on the last blocker for the 1.4.0 release and won't take more code PRs into that. But I am open to putting this on the track for the 1.5.0 release, which should be due in September.

NorseGaud: No worries. Would you mind if I work on the trace revival? I'd like to add the verbosity levels you spoke about.

martin-schulze-vireso: Go for it, I am glad for as many helping hands as I can get :smile: , but be warned that it can take some time (just look at the backlog of PRs).

NorseGaud: @martin-schulze-vireso , any major changes in 1.4.0 that I need to consider before moving forward? Or is it fairly the same structure?

martin-schulze-vireso: I think looking at [this](https://github.com/bats-core/bats-core/blob/master/docs/CHANGELOG.md#unreleased) should give you an idea.

martin-schulze-vireso: There is another PR lined up that changes `run`'s internals which I'd also like to get into the next release and which I'd give higher priority.

NorseGaud:
@martin-schulze-vireso , is there a way to allow contributors to cancel actions jobs that they started? I generated a bunch and want to kill them all.

martin-schulze-vireso:
I don't think so. If you want full control, you should work with a private branch in your fork.

peternewman:
> Since #137 seems to be dead and honestly isn't really a critically/urgently needed feature for users of bats (many of which I know personally)

I'm unclear of the exact difference between these PRs, but I've been stuck a few times trying to add or extend tests using BATS and really struggled due to the lack of easy way to debug what's going on, am I matching slightly the wrong output, is me calling it via BATS generating slightly different output, is my BATS test broken, or something else? Having to run the command standalone and in BATS and update both when debugging stuff is a real faff. A number of these PRs have stalled at this point, as the pain is too much and I've just gone onto other things.

> I am wondering if it is worthwhile to separate the output for each test into its own file. This would make it easier to find the interesting parts.

> It would allow for targetting the specific file to cat should it fail or the config set success STDOUT to be shown. I like that idea.

FWIW from my experience with other CI, what I actually want is to see the debug info if the test run fails. E.g. with Travis CI with both configure and make check, I have on fail bits of the config that cat config.log test-suite.log respectively if the build fails, ideally I, or particularly a contributor, don't want to have to tweak the CI each time there's a failure, so some easy way to cat just the failing files would be great, either only writing them to disk if a test failed, or something built in which outputs all of a failed test just before BATS dies.

Thanks for this though, I'm really looking forward to it!

NorseGaud:
> > Since #137 seems to be dead and honestly isn't really a critically/urgently needed feature for users of bats (many of which I know personally)
> 
> 
> 
> I'm unclear of the exact difference between these PRs, but I've been stuck a few times trying to add or extend tests using BATS and really struggled due to the lack of easy way to debug what's going on, am I matching slightly the wrong output, is me calling it via BATS generating slightly different output, is my BATS test broken, or something else? Having to run the command standalone and in BATS and update both when debugging stuff is a real faff. A number of these PRs have stalled at this point, as the pain is too much and I've just gone onto other things.
> 
> 
> 
> > I am wondering if it is worthwhile to separate the output for each test into its own file. This would make it easier to find the interesting parts.
> 
> 
> 
> > It would allow for targetting the specific file to cat should it fail or the config set success STDOUT to be shown. I like that idea.
> 
> 
> 
> FWIW from my experience with other CI, what I actually want is to see the debug info if the test run fails. E.g. with Travis CI with both configure and make check, I have on fail bits of the config that cat config.log test-suite.log respectively if the build fails, ideally I, or particularly a contributor, don't want to have to tweak the CI each time there's a failure, so some easy way to cat just the failing files would be great, either only writing them to disk if a test failed, or something built in which outputs all of a failed test just before BATS dies.
> 
> 
> 
> Thanks for this though, I'm really looking forward to it!

I can check again and make sure the level 2 output is included on failure, by default. Thanks for the feedback!

NorseGaud:
Hey @martin-schulze-vireso , I think I'm ready for a review. No tests yet, but again I'm going to wait until we're solid on the logic and how this should look before spending the time on that.

This is really cool so far 🥳

lib/bats-core/test_functions.bash
```diff
@@ -144,6 +148,9 @@ run() { # [--keep-empty-lines] [--output merged|separate|stderr|stdout] [--] <co
     ;;
   esac
 
+  # shellcheck disable=SC2034
+  BATS_TRACE_COMMAND_OUTPUT=("${lines[@]:-}")
```
martin-schulze-vireso, line 152: Do we really want to use $lines here ore rather $output?

lib/bats-core/tracing.bash
```diff
@@ -50,7 +50,7 @@ bats_print_stack_trace() {
 			# don't print "from function `source'"",
 			# when failing in free code during `source $test_file` from bats-exec-file
 			! [[ "$fn" == 'source' &&  $index -eq $count ]]; then 
-			printf "from function \`%s' " "$fn"
+			printf "from function '%s' " "$fn"
```
martin-schulze-vireso, line 53: I am not sure why ` was used in the first place but changing that is going to require some rework at least in our test suite. Can you motivate why this change is necessary?

lib/bats-core/tracing.bash
```diff
@@ -77,7 +77,7 @@ bats_print_failed_command() {
 	bats_frame_lineno "$frame" 'lineno'
 	bats_extract_line "$filename" "$lineno" 'failed_line'
 	bats_strip_string "$failed_line" 'failed_command'
-	printf '%s' "#   \`${failed_command}' "
+	printf '%s' "#   '${failed_command}' "
```
martin-schulze-vireso, line 80: Again, I have no problem with \`. However, my OCD kicks in when I see that this is not `printf '#  "'s'" "$failed_command"`.

libexec/bats-core/bats
```diff
@@ -116,6 +121,10 @@ report_formatter=''
 recursive=
 export BATS_TEMPDIR_CLEANUP=1
 output=
+export TRACE_LEVEL="${TRACE_LEVEL:-}"
```
martin-schulze-vireso, line 124: Why not BATS_TRACE_LEVEL? Also, I'd rather "firewall" the value to not transfer via the ENV.

libexec/bats-core/bats-exec-file
```diff
@@ -146,11 +146,13 @@ bats_file_exit_trap() {
     printf "not ok %d %s\n" "$((test_number_in_suite + 1))" "$FAILURE_REASON failed" >&3
     bats_print_stack_trace "${BATS_STACK_TRACE[@]}" >&3
     bats_print_failed_command >&3
-    while IFS= read -r line; do
-      printf "# %s\n" "$line"
-    done <"$BATS_OUT" >&3
-    if [[ -n "$line" ]]; then
-      printf '# %s\n' "$line"
+    if [[ "${TRACE_LEVEL}" -le 1 ]]; then
```
martin-schulze-vireso, line 149: Why is output disabled for LEVEL 2?

libexec/bats-core/bats-exec-test
```diff
@@ -106,6 +106,15 @@ bats_exit_trap() {
       BATS_ERROR_STATUS=1
     fi
     printf 'not ok %d %s\n' "$BATS_SUITE_TEST_NUMBER" "${BATS_TEST_DESCRIPTION}${BATS_TEST_TIME}" >&3
+    if [[ -z "${TRACE_LEVEL}" || "${TRACE_LEVEL}" -gt 0 ]]; then
+      if [[ -n "$BATS_TRACE_COMMAND" ]]; then
```
martin-schulze-vireso, line 110: I believe printing here will omit the trace for multicommand tests like:

```
@test {
    run commandA
    run commandB
}
```

will only print for `run commandB`.

libexec/bats-core/bats-exec-test
```diff
@@ -115,8 +124,16 @@ bats_exit_trap() {
     if [[ -n "$line" ]]; then
       printf '# %s\n' "$line"
     fi
+
     status=1
   else
+    if [[ "${TRACE_LEVEL}" -gt 0 && -n "$BATS_TRACE_COMMAND" ]]; then
```
martin-schulze-vireso, line 130: Printing before the test result when the test did not fail?

martin-schulze-vireso (changes requested):
It might be intentional to only print the failing command but most people will understand `--trace` to mean: print everything. What is done here is more like: print on failure with two modes command (arguably no big improvement over the status quo) and print output (this seems to be what you want to get at.)

What I envisioned was to switch inside `run` based on the TRACE_LEVEL, whether to print the executed command and output unconditionally. I think your current implementation has another usecase and both might be needed.

lib/bats-core/test_functions.bash
```diff
@@ -108,6 +109,9 @@ run() { # [--keep-empty-lines] [--output merged|separate|stderr|stdout] [--] <co
     ;;
   esac
 
+  # shellcheck disable=SC2034
+  BATS_TRACE_COMMAND="${*}"
```
martin-schulze-vireso, line 113: Capturing this as array might enable reuse like for #444

lib/bats-core/test_functions.bash
NorseGaud: Yep, $output doesn't keep the indentation:

Using $output:

```
❯ TRACE_LEVEL=2 bin/bats --filter "run should exit if tmpdir exist" test/bats.bats
 ✓ run should exit if tmpdir exist
 + bats --tempdir /var/folders/vb/pjyhgj4s491_k0t025_j56y00000gn/T//bats-run-36208-ZnuhSP/BATS_RUN_TMPDIR_TEST.pO5GPI /Users/norsegaud/bats-core/test/fixtures/bats/passing.bats
   Error: BATS_RUN_TMPDIR (/var/folders/vb/pjyhgj4s491_k0t025_j56y00000gn/T//bats-run-36208-ZnuhSP/BATS_RUN_TMPDIR_TEST.pO5GPI) already exists
Reusing old run directories can lead to unexpected results ... aborting!

1 test, 0 failures
```

Using $lines[@]

```
❯ TRACE_LEVEL=2 bin/bats --filter "run should exit if tmpdir exist" test/bats.bats
 ✓ run should exit if tmpdir exist
 + bats --tempdir /var/folders/vb/pjyhgj4s491_k0t025_j56y00000gn/T//bats-run-36294-rsFdVS/BATS_RUN_TMPDIR_TEST.AeY0tI /Users/norsegaud/bats-core/test/fixtures/bats/passing.bats
   Error: BATS_RUN_TMPDIR (/var/folders/vb/pjyhgj4s491_k0t025_j56y00000gn/T//bats-run-36294-rsFdVS/BATS_RUN_TMPDIR_TEST.AeY0tI) already exists
   Reusing old run directories can lead to unexpected results ... aborting!

1 test, 0 failures
```

lib/bats-core/tracing.bash
NorseGaud: Yep, I reworked the test suite to support the ' instead of the `. It's actually been there since day one and I'm almost certain it was a mistake: https://github.com/sstephenson/bats/commit/c850527cce7134f4adf4fe6dac07214678deb72b#diff-2285302ec7310b89fd9edfef3694598bebfa056fdbfd7e5d607026f64a68d734R80

libexec/bats-core/bats
NorseGaud: > Also, I'd rather "firewall" the value to not transfer via the ENV.

Sorry, can you clarify what you mean?

I can change the ENV names 👍🏼

libexec/bats-core/bats-exec-file
NorseGaud: Good catch, forgot to remove that!

libexec/bats-core/bats-exec-test
NorseGaud: Wow! What an oversight on my part 😅

I made the changes necessary to support this and updated the examples in the description :)

lib/bats-core/tracing.bash
NorseGaud: Reverted -- I'll do a separate PR for this

NorseGaud:
Ok, added tests and fixed an ordering issue. I believe it's ready for a full review at this point.

NorseGaud:
Also, I noticed that the ` causes code blocks (and probably CI) to freak out on styling:

Screen Shot 2021-07-25 at 3 32 53 PM

libexec/bats-core/bats
martin-schulze-vireso: When running the selftest suite setting BATS_TRACE_LEVEL during the outer bats run would result in enabling it for the inner run too. I'd like to avoid that.

NorseGaud: What do you recommend? I don't think I quite understand to be able to change what you want me to :( Sorry

libexec/bats-core/bats-exec-test
```diff
@@ -115,8 +115,16 @@ bats_exit_trap() {
...
     fi
+    
+    if [[ ( -z "${BATS_TRACE_LEVEL}" || "${BATS_TRACE_LEVEL}" -gt 0 ) && -n "${BATS_TRACE_OUTPUT_COMBINED[*]:-}" ]]; then
+      printf "# %s\n" "${BATS_TRACE_OUTPUT_COMBINED[@]}" >&3
+    fi
+
     status=1
   else
+    if [[ "${BATS_TRACE_LEVEL}" -gt 0 ]]; then
+      printf "   %s\n" "${BATS_TRACE_OUTPUT_COMBINED[@]}" >&3
```
martin-schulze-vireso, line 126: Why omit the leading `#`? This will break tap output.

Also, why print that before the `ok` but the other one after the `not ok`?
I think that should be unified and probably moved behind the if. However, I'd vote for only printing this output in failure cases.

test/bats.bats
```diff
@@ -872,12 +871,12 @@ EOF
...
 
   run cat "$TEMPFILE"
-  echo "$output"
+  
 
-  [[ "${lines[1]}" == "not ok 1 test" ]]
-  [[ "${lines[2]}" == "# (in test file ${RELATIVE_FIXTURE_ROOT}/hang_in_run.bats, line 7)" ]]
-  [[ "${lines[3]}" == "#   \`run sleep 10' failed with status 130" ]]
-  [[ "${lines[4]}" == "# Received SIGINT, aborting ..." ]]
+  [[ "${lines[*]}" =~ "not ok 1 test" ]]
```
martin-schulze-vireso, line 876: I think the test was more stringent before. This might allow for reordered output now.

test/bats.bats
```diff
@@ -1001,3 +1000,87 @@ EOF
   run stars
   [ "${lines[0]}" = '**' ]
 }
+
+
+@test "--trace: 0 + BATS_TRACE_LEVEL=0 doesn't show anything on failure" {
+  OUTPUT="$(bats --trace 0 --filter "test 1" "$FIXTURE_ROOT/read_from_stdin.bats")"
+  [[ "${OUTPUT[@]}" == "1..1
+ok 1 test 1" ]]
+  BATS_TRACE_LEVEL=0
```
martin-schulze-vireso, line 1009: Putting the variable inside the `$()` will allow you to omit the unset.

test/bats.bats
```diff
@@ -1001,3 +1000,87 @@ EOF
...
+  OUTPUT="$(bats --filter "test 1" "$FIXTURE_ROOT/read_from_stdin.bats")"
+  [[ "${OUTPUT[@]}" == "1..1
+ok 1 test 1" ]]
+  unset BATS_TRACE_LEVEL
```
martin-schulze-vireso, line 1013: I believe that is another problem of having the internal var being the same name as the external one.

docs/source/writing-tests.md
```diff
@@ -270,6 +275,28 @@ your custom text. Here are some detailed guidelines to refer to:
...
+    output both commands and their stdout/err.
+
+    ```bash
+    ❯ bin/bats --trace 2 --filter "CTRL-C aborts and fails the current teardown_file" test
+    ✗ CTRL-C aborts and fails the current teardown_file
+      (in test file test/bats.bats, line 981)
+        \`[[ "${lines[1]}" == "ok 1 empty" ]]' failed
+      + cat /var/folders/vb/pjyhgj4s491_k0t025_j56y00000gn/T//bats-run-83801-9CU13w/suite-tmpdir-gJzQY7/file-tmpdir-bats.bats-FXmB66/test-tmpdir--IyhaL1/test_CTRL-2dC_aborts_and_fails_the_current_teardown-2d5ffile.log
+      1..1
+
```
martin-schulze-vireso, line 289: I am seeing spurious empty lines in my trace output too. Is that intended?

test/bats.bats
```diff
@@ -1001,3 +1000,87 @@ EOF
...
+
+@test "--trace: 1 + BATS_TRACE_LEVEL=1" {
+  OUTPUT="$(bats --trace 1 --filter "test 1" "$FIXTURE_ROOT/read_from_stdin.bats")"
+  [[ "${OUTPUT}" =~ "1..1" ]]
+  echo "---------------"
+  echo "${OUTPUT}"
+  echo "$FIXTURE_ROOT"
+  [[ "${OUTPUT}" =~ "   + bash -c $FIXTURE_ROOT/cmd_using_stdin.bash" ]]
+  [[ "${OUTPUT}" =~ "ok 1 test 1" ]]
+  run bats --trace 1 --filter "test 1" "$FIXTURE_ROOT/read_from_stdin.bats"
```
martin-schulze-vireso, line 1024: Why is this using `run` and the first one `OUTPUT=$()`?

martin-schulze-vireso (changes requested):
I find the output a bit confusing. I understand that you might want to have `--trace` also output on successful test but I think it might be too verbose. I'd rather have another `--verbose` switch that turns on output for successful tests as well. Furthermore, I'd like to inline the trace output with the test output, so you get to print your own context information right next to it.

My suggestion would be to do the printing inside the `run` function. That should simplify the code quite a bit and achieve all the points I mentioned above.

NorseGaud:
@martin-schulze-vireso 

> Furthermore, I'd like to inline the trace output with the test output, so you get to print your own context information right next to it.

Mind providing an example?

>  I understand that you might want to have --trace also output on successful test but I think it might be too verbose. I'd rather have another --verbose switch that turns on output for successful tests as well.

Ok, I can change the scope of this to ONLY output on failure, but I also want to confirm that this is indeed what you're saying to do. It seems like you reviewed a lot of code, so I need a bit of clarity on what changes you want and how you see this exactly so that we can get this PR approved and moving forward.

Thanks for your time!

NorseGaud:
OK, changes were made using Martin's example. I think the last thing is to add some docs for how to use these features. Pending a review first before I do that.

martin-schulze-vireso:
We might also want to offer tracing non `run` commands via the debug trap.

martin-schulze-vireso:
More tasks:
- ensure the linked, closed issues/PRs are still solved
- add --trace to the Tutorial
- add an example for toggling trace/verbose from inside tests via the variable

NorseGaud:
Ok, changes made. Ready for review

martin-schulze-vireso (commented):
I think we still have a communication problem what trace/verbose should actually achieve, so I'd like to take a step back and look at the situation. We have following planned features:

1. display the commands that are executed (as in `set -x`)
2. display the output of all `run` statements
3. display output of succeeding tests
4. (optional) provide 3rd's output as separate log files instead of stdout

I realized now, that the wording `trace` should only be applicable to one, everything else would be confusing. `verbose` could match both 2 and 3. IMHO it would be acceptable to have `run --verbose` print run's output. However, `bats --verbose` printing every runs output **and** succeeding tests' too is to much for my taste.

As to 4, I just realized that the current output is already stored in the temp folder per test. @dimo414 Would this be sufficient for your usecase?

My suggestion would be to split the verbose mode up. Have `bats --verbose` for printing succeeding tests' output and `bats --verbose-run`/`run --verbose`/`BATS_VERBOSE_RUN` for getting `run`'s output printed by default.

NorseGaud:
Please keep in mind that this has now been rewritten almost entirely, multiple times now. I do admit that original code may not have been the best, but I have spent hours at this point and want to avoid spending any more unnecessary effort on rewrites. Let's get absolutely aligned before I do any more coding.

I think splitting `--verbose` up is overkill. How about just ditching `--trace` and allow verbose to have different levels? Then, we'll just output commands and STDOUT/ERR by default on failure? That's by far the easiest and most intuitive version of this. I don't think anyone would disagree with this approach. When wouldn't you want to see everything that happened in a failing test?

martin-schulze-vireso:
> Please keep in mind that this has now been rewritten almost entirely, multiple times now. I do admit that original code may not have been the best, but I have spent hours at this point and want to avoid spending any more unnecessary effort on rewrites. Let's get absolutely aligned before I do any more coding.

Yeah, I am very sorry about that. It was quite a journey and tacked on some additional features along the way. I appreciate your work and don't want it to go to waste. However, I also have to align it with the status quo and other feature requests.

> I think splitting `--verbose` up is overkill. How about just ditching `--trace` and allow verbose to have different levels? Then, we'll just output commands and STDOUT/ERR by default on failure? That's by far the easiest and most intuitive version of this. I don't think anyone would disagree with this approach. When wouldn't you want to see everything that happened in a failing test?

I just today did a survey of about 300 projects which are using bats to find out whether I would break them by changing some mechanics of the run function. There is a wide spectrum of usages and it is hard to anticipate them all. Many discussions in the issues revolve around people who don't understand each others usecases. I can see two good reasons why one would not always want get `$output` printed at the end of a failed test:

1. it is too long/irrelevant (we are only interested in the return code is one reason for using `run`, so far)
2. it already was printed (take for example `assert_success` from bats-assert that will still be used by many projects out there)

I guess the point I am trying to make above is this: There is already a huge amount of variability in the "published" usage of bats. However, what is not published is the interaction of people while working on bats tests. I for one, would like to be able to see the `run --verbose` log but am not interested in the output of succeeding tests. That is not my usecase but I still intend to support it for those who requested it. However, coupling both usecases together under `--verbose` forces the unwanted half unto the other "unsuspecting" users. I really dislike the verbosity (no pun intended) of the split up interface. The only thing I can do to decide on an acceptable solution for both sides is to throw my ideas out there and listen for feedback.

Now I'll have to get some sleep (its well past midnight) and maybe the picture will clear tomorrow.

NorseGaud:
Is it fair then to say we could, as a first step, only output verbose information from run on failure then?

peternewman:
>     1. it is too long/irrelevant (we are only interested in the return code is one reason for using `run`, so far)

Given pretty much everyone is likely to match on the return code at some point during their tests and from a look at the example output from this PR in:
https://github.com/bats-core/bats-core/pull/467#issuecomment-886247460

It's still not showing that return code, it would be awesome if that could be output in one of the verbose modes to save everyone having to manually add it to their test run (although maybe I should open a separate issue/wait for another PR for that once this one is in)?

>     2. it already was printed (take for example `assert_success` from bats-assert that will still be used by many projects out there)

You seem to be suggesting bats-assert is less used, so apologies if I should be still achieving the above via that, but I figure with the verbose stuff we've got a good chance of understanding why line matching tests are failing, which would just leave the exit code as the remaining issue.

martin-schulze-vireso:
@peternewman have a look at #367 which is about to be merged before this PR. It is intended as a better alternative to `assert_success` and its ilk. Hence my hint that its use might decline.

@NorseGaud you seem to be pushing for a fast introduction of this. Is there any pressing need or so you just want to be done with it?

NorseGaud:
> @NorseGaud you seem to be pushing for a fast introduction of this. Is there any pressing need or so you just want to be done with it? I would not want to change the default behavior. However, I am fine with an additional switch that turns this on.
> 
> I am somewhat sceptical of this recent option explosion but I think it is better than having a one size fits all approach and is indicative of the progress in feature requests that has long been stalled.

Not really blocked. It's more of a convenience thing for my own usage of bats. I also in general have a bit of a rushed nature to the way I work, so apologies if it's stressing anyone out 😅  I have to checkout this ref/branch in all of the projects I'm working on so I can get verbose output and speed up my creation and troubleshooting of tests. I also had a discussion recently with a coworker and a friend about how they both dislike using bats because there isn't a feature like this. I have a bit more of a create-if-missing/fix-if-broken mentality, but I share slightly in the sentiment.

Here are the two things that I feel are essential to whatever we decide on as a final version:

1. Seeing the commands and stdout/err for both failing and successful tests. I often do sanity checks on my tests to be sure the interpolation and other things are correct and exactly what I want them to be. It really speeds me up :)
2. A way to show commands and stdout/err only on failure. I think that having a way to do this with a flag is fine; it doesn't need to be default.

I'm cool with reworking this as long as is necessary, so definitely not a "be done with it" attitude currently :)

martin-schulze-vireso:
> > @NorseGaud you seem to be pushing for a fast introduction of this. Is there any pressing need or so you just want to be done with it? I would not want to change the default behavior. However, I am fine with an additional switch that turns this on.
> > 
> > I am somewhat sceptical of this recent option explosion but I think it is better than having a one size fits all approach and is indicative of the progress in feature requests that has long been stalled.
> 
> Not really blocked. It's more of a convenience thing for my own usage of bats. I also in general have a bit of a rushed nature to the way I work, so apologies if it's stressing anyone out 😅  I have to checkout this ref/branch in all of the projects I'm working on so I can get verbose output and speed up my creation and troubleshooting of tests. I also had a discussion recently with a coworker and a friend about how they both dislike using bats because there isn't a feature like this. I have a bit more of a create-if-missing/fix-if-broken mentality, but I share slightly in the sentiment.

Well that same attitude got me here. Ijust wanted to put one little PR into docker-mailserver...

I'd like to be faster but I don't want to create more work down the line and am juggling multiple PRs at the same time.

> 
> Here are the two things that I _feel_ are essential to whatever we decide on as a final version:
> 
> 1. Seeing the commands and stdout/err for both failing and successful tests. I often do sanity checks on my tests to be sure the interpolation and other things are correct and exactly what I want them to be. I'm ok with a lot of output. It really speeds me up :)
> 2. A way to show commands and stdout/err only on failure. I think that having a way to do this with a flag is fine; it doesn't need to be default.
> 
> I'm cool with reworking this as long as is necessary, so definitely not a "be done with it" attitude currently :)

My goal is to get all of this in. I just want to do it right: without breaking existing users, which includes bad performance especially in large test suites, and without locking ourselves into an untenable interface.

NorseGaud:
> > > @NorseGaud you seem to be pushing for a fast introduction of this. Is there any pressing need or so you just want to be done with it? I would not want to change the default behavior. However, I am fine with an additional switch that turns this on.
> > > 
> > > I am somewhat sceptical of this recent option explosion but I think it is better than having a one size fits all approach and is indicative of the progress in feature requests that has long been stalled.
> > 
> > Not really blocked. It's more of a convenience thing for my own usage of bats. I also in general have a bit of a rushed nature to the way I work, so apologies if it's stressing anyone out 😅  I have to checkout this ref/branch in all of the projects I'm working on so I can get verbose output and speed up my creation and troubleshooting of tests. I also had a discussion recently with a coworker and a friend about how they both dislike using bats because there isn't a feature like this. I have a bit more of a create-if-missing/fix-if-broken mentality, but I share slightly in the sentiment.
> 
> Well that same attitude got me here. Ijust wanted to put one little PR into docker-mailserver...
> 
> I'd like to be faster but I don't want to create more work down the line and am juggling multiple PRs at the same time.
> 
> > 
> > Here are the two things that I _feel_ are essential to whatever we decide on as a final version:
> > 
> > 1. Seeing the commands and stdout/err for both failing and successful tests. I often do sanity checks on my tests to be sure the interpolation and other things are correct and exactly what I want them to be. I'm ok with a lot of output. It really speeds me up :)
> > 2. A way to show commands and stdout/err only on failure. I think that having a way to do this with a flag is fine; it doesn't need to be default.
> > 
> > I'm cool with reworking this as long as is necessary, so definitely not a "be done with it" attitude currently :)
> 
> My goal is to get all of this in. I just want to do it right: without breaking existing users, which includes bad performance especially in large test suites, and without locking ourselves into an untenable interface.
> 
> 

Understood, no worries. Just let me know what we need to turn this PR into to fulfill on the goal. I get the feeling doing the smallest and most impactful change as a first step is likely the best course of action. For example, verbose output on failure would be my recommendation. That's when I need to see exactly what's going on 100% of the time.

NorseGaud:
Wanted to update here that I've been using this code in my day-to-day adding new tests or troubleshooting failures and it's awesome! The speed at which I can write bats tests now is really pleasant.

martin-schulze-vireso:
I am working on getting a rebased version pushed with additional flags. There is more to come...

martin-schulze-vireso:
@NorseGaud I see that this was the master branch on your repository. I hope I did not break your existing scripts with force pusing over that. I also hope that I resolved all merge conflicts successfully without breaking either side.

NorseGaud:
> @NorseGaud I see that this was the master branch on your repository. I hope I did not break your existing scripts with force pusing over that. I also hope that I resolved all merge conflicts successfully without breaking either side.

Nah, it's all good. I mistakenly did that and just was too lazy to fix it 😅

NorseGaud:
How do we want to handle shellcheck stuff?

martin-schulze-vireso:
@NorseGaud maybe we should hop over to gitter to resolve this more quickly.

martin-schulze-vireso:
Just to give a short summary of what is in this right now:

- `bats --trace` (or via `$BATS_TRACE_LEVEL=1` env var, this name might be subject to change): print executed commands
- `bats --print-output-on-failure` (or via `$BATS_PRINT_OUTPUT_ON_FAILURE` env var): print the value of "$output" when a test fails
- `bats --show-output-of-succeeding-tests` (or via `$BATS_SHOW_OUTPUT_OF_SUCCEEDING_TESTS` env var)
- `bats --verbose-run` (or via `$BATS_VERBOSE_RUN` env var): make `run` print `$output` by default
- `bats --gather-test-outputs-in <dir>` (or via `$BATS_GATHER_TEST_OUTPUTS_IN=D<dir>`, but then you have to ensure the directory exists): put output of each test (even succeeding!) into the given folder as files of the format `<test number in suite>-<test description>.log`

NorseGaud:
@martin-schulze-vireso , is there `--trace` but without all of the extra stuff? It's painful to use and scroll through to see what's running and output. Here is the output from my usage:

```
 ✗ testtest checking setup.sh: setup.sh relay add-domain
   (from function `assert_success' in file test/test_helper/bats-assert/src/assert.bash, line 114,
    in test file test/tests.bats, line 1152)
     `assert_success' failed
   $$$ [tests.bats:8]
   $$$ setup
   $$$ [tests.bats:9]
   $$$ run_setup_file_if_necessary
   $$$$ [common.bash:123]
   $$$$ native_setup_teardown_file_support
   $$$$$ [common.bash:107]
   $$$$$ local VERSION_REGEX='([0-9]+)\.([0-9]+)\.([0-9]+)'
   $$$$$ [common.bash:109]
   $$$$$ [[ "${BATS_VERSION}" =~ ${VERSION_REGEX} ]]
   $$$$$ [common.bash:110]
   $$$$$ numeric_version=$(( (BASH_REMATCH[1] * 100 + BASH_REMATCH[2]) * 100 + BASH_REMATCH[3] ))
   $$$$$ [common.bash:111]
   $$$$$ [[ ${numeric_version} -ge 10201 ]]
   $$$$$ [common.bash:112]
   $$$$$ [ "${BATS_TEST_NAME}" == 'test_first' ]
   $$$$$ [common.bash:115]
   $$$$$ return 0
   $$$$ [common.bash:123]
   $$$$ return 0
   $ [tests.bats:1145]
   $ ./setup.sh relay add-domain example1.org smtp.relay1.com 2525
   $ [tests.bats:1146]
   $ ./setup.sh relay add-domain example2.org smtp.relay2.com
   $ [tests.bats:1147]
   $ ./setup.sh relay add-domain example3.org smtp.relay3.com 2525
   $ [tests.bats:1148]
   $ ./setup.sh relay add-domain example3.org smtp.relay.com 587
   $ [tests.bats:1151]
   $ run /bin/sh -c "cat $(private_config_path mail)/postfix-relaymap.cf | grep -e \"^@example1.org\s\+\[smtp.relay1.com\]:2525\" | wc -l | grep 1"
   $ [tests.bats:1152]
   $ assert_success
   $$ [assert.bash:109]
   $$ (( status != 0 ))
   $$ [assert.bash:113]
   $$ batslib_decorate 'command failed'
   $$ [assert.bash:114]
   $$ fail
   $$$ [error.bash:39]
   $$$ (( $# == 0 ))
   $$$ [error.bash:39]
   $$$ batslib_err
   $$$$ [output.bash:38]
   $$$$ (( $# > 0 ))
   $$$$ [output.bash:41]
   $$$$ cat -
   $$$ [output.bash:274]
   $$$ echo
   
   $$$ [output.bash:275]
   $$$ echo "-- $1 --"
   -- command failed --
   $$$ [output.bash:276]
   $$$ cat -
   $$ [assert.bash:110]
   $$ local -ir width=6
   $$ [assert.bash:111]
   $$ batslib_print_kv_single "$width" 'status' "$status"
   $$$ [output.bash:130]
   $$$ local -ir col_width="$1"
   $$$ [output.bash:130]
   $$$ shift
   $$$ [output.bash:131]
   $$$ (( $# != 0 ))
   $$$ [output.bash:132]
   $$$ printf '%-*s : %s\n' "$col_width" "$1" "$2"
   status : 1
   $$$ [output.bash:133]
   $$$ shift 2
   $$$ [output.bash:131]
   $$$ (( $# != 0 ))
   $$ [assert.bash:112]
   $$ batslib_print_kv_single_or_multi "$width" 'output' "$output"
   $$$ [output.bash:183]
   $$$ local -ir width="$1"
   $$$ [output.bash:183]
   $$$ shift
   $$$ [output.bash:184]
   $$$ local -a pairs=("$@")
   $$$ [output.bash:186]
   $$$ local -a values=()
   $$$ [output.bash:187]
   $$$ local -i i
   $$$ [output.bash:188]
   $$$ ((i=1))
   $$$ [output.bash:188]
   $$$ ((i < 2))
   $$$ [output.bash:189]
   $$$ values+=("${pairs[$i]}")
   $$$ [output.bash:188]
   $$$ ((i+=2 ))
   $$$ [output.bash:188]
   $$$ ((i < 2))
   $$$ [output.bash:192]
   $$$ batslib_is_single_line "${values[@]}"
   $$$$ [output.bash:81]
   $$$$ for string in "$@"
   $$$$ [output.bash:82]
   $$$$ (( $(batslib_count_lines "$string") > 1 ))
   /Users/norsegaud/docker-mailserver/test/test_helper/bats-support/src/output.bash: line 82: $$$$ [output.bash:82]
   $$$$ batslib_count_lines "$string"
   $$$$$ [output.bash:63]
   $$$$$ local -i n_lines=0
   $$$$$ [output.bash:64]
   $$$$$ local line
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:65]
   $$$$$ [[ -n $line ]]
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:65]
   $$$$$ [[ -n $line ]]
   $$$$$ [output.bash:68]
   $$$$$ echo "$n_lines"
   8 > 1 : syntax error: operand expected (error token is "$$$$ [output.bash:82]
   $$$$ batslib_count_lines "$string"
   $$$$$ [output.bash:63]
   $$$$$ local -i n_lines=0
   $$$$$ [output.bash:64]
   $$$$$ local line
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:65]
   $$$$$ [[ -n $line ]]
   $$$$$ [output.bash:66]
   $$$$$ (( ++n_lines ))
   $$$$$ [output.bash:65]
   $$$$$ IFS='' read -r line
   $$$$$ [output.bash:65]
   $$$$$ [[ -n $line ]]
   $$$$$ [output.bash:68]
   $$$$$ echo "$n_lines"
   8 > 1 ")
   $$$$ [output.bash:84]
   $$$$ return 0
   $$$ [output.bash:193]
   $$$ batslib_print_kv_single "$width" "${pairs[@]}"
   $$$$ [output.bash:130]
   $$$$ local -ir col_width="$1"
   $$$$ [output.bash:130]
   $$$$ shift
   $$$$ [output.bash:131]
   $$$$ (( $# != 0 ))
   $$$$ [output.bash:132]
   $$$$ printf '%-*s : %s\n' "$col_width" "$1" "$2"
   output : cat: $: No such file or directory
   cat: [tests.bats:1151]: No such file or directory
   /bin/sh: line 1: $: command not found
   /bin/sh: line 2: 29678: command not found
   /bin/sh: line 3: 29678: command not found
   /bin/sh: line 4: /Users/norsegaud/docker-mailserver/test/duplicate_configs/mail/postfix-relaymap.cf: Permission denied
   $$$$ [output.bash:133]
   $$$$ shift 2
   $$$$ [output.bash:131]
   $$$$ (( $# != 0 ))
   $$$ [output.bash:277]
   $$$ echo '--'
   --
   $$$ [output.bash:278]
   $$$ echo
   
   $$$ [error.bash:40]
   $$$ return 1
```
Definitely way too much output

martin-schulze-vireso:
Obviously there is a lot going on. The problems that I can identify:

1. the trace also contains internals of the bats-assert library. This is definitely not what we want. However, from core is hard to anticipate where it is installed. I have to think about a general solution how to disable traces in specific areas.
2. The file labels are too much on many locations. I think they would suffice at each stack push/pop event.
3. the trace output is not printed into its own FD, this likely breaks some internals.

I'll see to it ASAP.

martin-schulze-vireso:
Okay, I pushed the changes discussed above apart from the FD change. I am not sure if it is necessary and I'd really like the interleaving of the output, which will be broken when using another FD that is not buffered by bats.

martin-schulze-vireso:
Okay, the separate FD is necessary but I still got the interleaving. I opted for using static FD4. Hopefully that is enough for most cases.

NorseGaud:
Way more managable now and doesn't fail the tests anymore! Nice! 

Some examples after a changed some tests to fail:

```
❯ test/bats/bin/bats --trace --filter "checking setup.sh: setup.sh delquota" test/tests.bats 
 ✗ checking setup.sh: setup.sh delquota
   (in test file test/tests.bats, line 1103)
     `assert_failure' failed
   $ [tests.bats:1094]
   $ run ./setup.sh email add quota_user@example.com test_password
   $ run ./setup.sh email add quota_user2@example.com test_password
   $ run ./setup.sh quota set quota_user@example.com 12M
   $ assert_success
   $ run /bin/sh -c 'cat ./test/duplicate_configs/mail/dovecot-quotas.cf | grep -E "^quota_user@example.com\:12M\$" | wc -l | grep 1'
   $ assert_success
   $ run ./setup.sh quota del unknown@domain.com
   $ assert_failure
   $ run /bin/sh -c 'cat ./test/duplicate_configs/mail/dovecot-quotas.cf | grep -E "^quota_user@example.com\:12M\$" | wc -l | grep 1'
   $ assert_failure
   
   -- command succeeded, but it was expected to fail --
   output :        1
   --
   

1 test, 1 failure
```

```
❯ test/bats/bin/bats --trace --verbose-run --filter "checking setup.sh: setup.sh delquota" test/tests.bats
 ✗ checking setup.sh: setup.sh delquota
   (in test file test/tests.bats, line 1103)
     `assert_failure' failed
   $ [tests.bats:1094]
   $ run ./setup.sh email add quota_user@example.com test_password
   
   $ run ./setup.sh email add quota_user2@example.com test_password
   
   $ run ./setup.sh quota set quota_user@example.com 12M
   
   $ assert_success
   $ run /bin/sh -c 'cat ./test/duplicate_configs/mail/dovecot-quotas.cf | grep -E "^quota_user@example.com\:12M\$" | wc -l | grep 1'
          1
   $ assert_success
   $ run ./setup.sh quota del unknown@domain.com
   Usage: delquota <username@domain>
   Error :: user unknown@domain.com does not exist
   Aborting.
   
   --- UNCHECKED ERROR
     - script    = setup.sh
     - function  = _docker_container / ${CRI} exec "${USE_TTY}" "${CONTAINER_NAME}" "${@:+$@}"
     - line      = 245
     - exit code = 1
   
   Make sure you use a version of this script that matches
   the version / tag of docker-mailserver. Please read the
   'Get the tools' section in the README on GitHub careful-
   ly and use ./setup.sh help and read the VERSION section.
   $ assert_failure
   $ run /bin/sh -c 'cat ./test/duplicate_configs/mail/dovecot-quotas.cf | grep -E "^quota_user@example.com\:12M\$" | wc -l | grep 1'
          1
   $ assert_failure
   
   -- command succeeded, but it was expected to fail --
   output :        1
   --
   

1 test, 1 failure
```

NorseGaud:
I have what I need with the following:

`--trace --verbose-run` seems to work great for failures.
`--trace --verbose-run --show-output-of-succeeding-tests` works great for success (though output is in red).

martin-schulze-vireso:
@NorseGaud I am sorry, I had a typo in the (force) push command and probably erased the master branch on your remote. It looks like I can't reinstate it from my end and can't even reopen this PR.

NorseGaud:
All good @martin-schulze-vireso , I can restore the branch

PR #483, opened 2021-08-18
Add --filter-status failed for running only failed tests from the last completed run

martin-schulze-vireso:
Often only a little subset of tests is failing. While working on fixing them, running other tests just takes up time. While the filtering function can be used to filter down to a small number of tests, it can become cumbersome to select all tests that failed.  This flag to Bats introduces that feature. What this PR does:

- check for .bats/run-logs/ in the test directory (suite dir or containing dir of first test file)
  - if this directory exists, a log of all passed/failed tests will be written
  - if a test run is aborted via SIGINT/CTRL+C no log will be stored
- when `--filter-status` is activated it picks the most recent log in .bats/run-logs
  - if no log does exist, we assume a first run and will run all tests (subject to filtering/file selection)
  - if the log exist, run all tests of the specified status in the last run (+missed tests: tests that were not encountered in the last run e.g., due to renaming/adding them)

TODO:
- documentation
  - [x] man page
  - [x] --help text

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /bats-*.tgz
 # we don't have any deps; un-ignore if that changes
 /package-lock.json
+test/.bats/run-logs/
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 * using external formatters via `--formatter <absolute path>` (also works for
   `--report-formatter`) (#602)
+* running only tests that failed in the last run via `--filter-status failed` (#483)
 
 #### Documentation
 
--- a/lib/bats-core/common.bash
+++ b/lib/bats-core/common.bash
@@ -63,3 +63,36 @@ bats_require_minimum_version() { # <required version>
     BATS_GUARANTEED_MINIMUM_VERSION="$required_minimum_version"
   fi
 }
+
+bats_binary_search() { # <search-value> <array-name>
+  local -r search_value=$1 array_name=$2
+  
+  if [[ $# -ne 2 ]]; then
+    printf "ERROR: bats_binary_search requires exactly 2 arguments: <search value> <array name>\n" >&2
+    return 2
+  fi
+
+  # we'd like to test if array is set but we cannot distinguish unset from empty arrays, so we need to skip that
+
+  local start=0 mid end mid_value 
+  # start is inclusive, end is exclusive ...
+  eval "end=\${#${array_name}[@]}"
+
+  # so start == end means empty search space
+  while (( start < end )); do
+    mid=$(( (start + end) / 2 ))
+    eval "mid_value=\${${array_name}[$mid]}"
+    if [[ "$mid_value" == "$search_value" ]]; then
+      return 0
+    elif [[ "$mid_value" < "$search_value" ]]; then
+      # This branch excludes equality -> +1 to skip the mid element.
+      # This +1 also avoids endless recursion on odd sized search ranges.
+      start=$((mid + 1))
+    else
+      end=$mid
+    fi
+  done
+
+  # did not find it -> its not there
+  return 1
+}
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -41,6 +41,10 @@ HELP_TEXT_HEADER
                             or 'custom' which requires setting $BATS_BEGIN_CODE_QUOTE and 
                             $BATS_END_CODE_QUOTE. Can also be set via $BATS_CODE_QUOTE_STYLE
   -f, --filter <regex>      Only run tests that match the regular expression
+  --filter-status <status>  Only run tests with the given status in the last completed (no CTRL+C/SIGINT) run.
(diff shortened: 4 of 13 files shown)


martin-schulze-vireso:
I dislike two things about the current solution:
1. if we abort the recording mid run, we won't catch all failing tests, which will omit them from reruns.
2. we might want to lock in a set of failing tests and only run this subset without narrowing down further. This can happen when we change something that affects multiple tests.

I've got the feeling that there is a more minimal/modular solution lurking under the surface: Splitting the recording and "replay" functionality into separate flags would allow for 2.

Similarly, it might be a good idea to record all tests that ran, and save their state as failing/passing, then a replay run can skip all passed tests and run those that failed or were not run.

cyphar:
I think it might be a good idea to implement it as a list of test names with the pass and fail state. That would be much simpler than splitting the feature into two options and would solve the problem of the test run being cancelled (or the problem of new tests being added between `--rerun-failed` runs).

However, the file creation logic and lifetime is a little bit worrisome. You can't use `--rerun-failed` without having to specify it the first time (which is not a great UX), but at the same time this means if you wanted to start from scratch you need to know to delete the `.bats-*` file because in order to record the file you need to pass the option but if you pass the option and the file exists it won't write a new file from scratch.

Ideally we would always record the test results and then when `--rerun-failed` is used, we read that file. It might also be nice for `--rerun-failed` to update the test results file but I'm not sure if that's always something that a user would want to do (maybe they want to re-run a set of flaky tests multiple times). I'm also a little worried about creating such a file in the test directory -- if it accidentally gets checked into version control then that's just asking for trouble. But I guess we can't put it in `~/.cache/bats` because then we'd have to figure out a way to represent the path in a nice way.

Also IMHO the option name is somewhat confusing (are we retrying a test immediately after it failed?). `--run-last-failed` or something might be better, but I'm just spitballing.

martin-schulze-vireso:
> I think it might be a good idea to implement it as a list of test names with the pass and fail state. That would be much simpler than splitting the feature into two options and would solve the problem of the test run being cancelled (or the problem of new tests being added between `--rerun-failed` runs).

Yes, I was thinking about this too. The current state just was an implementation shortcut because the format already works with the internal test list. We also would have to make sure that the test names are parsable from on line.

> However, the file creation logic and lifetime is a little bit worrisome. You can't use `--rerun-failed` without having to specify it the first time (which is not a great UX), but at the same time this means if you wanted to start from scratch you need to know to delete the `.bats-*` file because in order to record the file you need to pass the option but if you pass the option and the file exists it won't write a new file from scratch. The fact it's a hidden file (and one that most users would add to `.gitignore`) makes this even more of an issue.

No, I'd actually like to update the file.

> Ideally we would always record the test results and then when `--rerun-failed` is used, we read that file. It might also be nice for `--rerun-failed` to update the test results file but I'm not sure if that's always something that a user would want to do (maybe they want to re-run a set of flaky tests multiple times). I'm also a little worried about creating such a file in the test directory -- if it accidentally gets checked into version control then that's just asking for trouble. But I guess we can't put it in `~/.cache/bats` because then we'd have to figure out a way to represent the path in a nice way.

I was torn on where to put it too. I am not sure if `~/.cache/` is available on all supported systems and as you said, once you centralize you need to pick it apart again. I looked at how other test frameworks deal with this and found the containing folder to be a common solution. I chose to use  `.bats/` to allow for reusing it later on for other files. The hidden nature should actually encourage people to ignore it. We might be even clearer with something like `.bats-vcs-ignore`?

> Also IMHO the option name is somewhat confusing (are we retrying a test immediately after it failed?). `--run-last-failed` or something might be better, but I'm just spitballing.

I tried to avoid creating the failed test log unconditionally to not force it unto unsuspecting users with the new version. A middle ground might be to require explicit approval on first run. I am not sure if this is a feature of interest for CI systems. I could envision exporting the failed test log as an artifact for download, so you can streamline your local error search or later runs.

Anyways, my idea of --rerun-failed semantics would be to narrow down the list of failing tests incrementally, so each new complete run would capture more detail. I think solution that only overwrites with a new capture if the test suite completed would be good enough for now. This can be improved later on. However, the interface should not change.

martin-schulze-vireso:
@cyphar Some more ideas:

We could require the user to create the `.bats/` to avoid confusion about whether to put it into .gitignore. If the directory does not exist, the files will be gathered in a central place, e.g. `/tmp` and can be moved over, once the user decides to rerun failed tests. That way, we don't lose the last run.

martin-schulze-vireso:
@cyphar After some experimenting and reading over your comment again I have the following suggestion:

1. rename to `--filter-status failed`, this leaves open the possibility to filter for other results like passed or skipped or whatever
2. only capture when there is a `.bats/run-logs` directory in the tests folder, which the user has to create actively
  - this means if we use `--filter-status` without `.bats/run-logs` existing, the user is instructed to create the folder and .gitignore it. They may lose the previous run but I think the complexity of catching that corner case is not worth the effort and problems with automatically generating the folder
4. when this directory exists, each non-aborted run will capture the status of executed tests

martin-schulze-vireso:
@cyphar Please have a look again. I rewrote it with the following changes:

PR #491, opened 2021-08-24
Add timeout support

### LINKED ISSUE #396: Add optional timeout for setup functions, files, and tests
Sometimes we want timing restrictions on `setup_file`, `setup`, single tests, or a whole file. Currently, we can only timeout a whole test run from the outside.

This issue should gather ideas and provide space for a discussion on where to find low hanging fruits.

TApplencourt:

Hi,

The PR tries to fix #396.  It adds a global timeout options (`-E <second>`) to `./bin/bats`. 
This option adds a timeout to each test.  If a test takes longer than the timeout, it will be killed and will exit with an error code set at `141`. In the case of the `--tap` formater, a `# timetout` will be printed in the terminal (similar to the `#skip` behavior)

### Technical detail

The implementation is inspired by the great work of https://github.com/MountainField/bash-timeout.
 I did tweak it a little bit so it works on my SUSE system.  The main idea is to spawn a subprocess who will sleep and kill himself and his parents when waking up. 
 
 ### Summary 
 
 I'm not a bash expert, nor a `bats` one. But I implemented a version of timeout that worked for my use case, and maybe also useful for other people.
 
 Please don't hesitate if you have any feedback,  I will be happy to do any modification to the PR to make it more "professional",
  
 Hope this help,
 Regards@


martin-schulze-vireso:
Using SIGPIPE might get caught in some user scripts. I would propose to use SIGALRM or SIGUSR1/2.

TApplencourt:
Thanks a lot for the review.

>Using SIGPIPE might interfere with some user scripts. I would propose to use SIGALRM or SIGUSR1/2

I spent a few hours today trying to do so but wasn't success full. My trap was either never called or without a trap, I got some error message corresponding to the trap call. Will potentially invest more time later :) 

>Furthermore, I think we should name the variable BATS_TEST_TIMEOUT to allow for other kinds of timeouts later on (like setup/setup_file,....)

Good idea will do!

martin-schulze-vireso:
I already have some local changes for that. I am not sure how to orocees with the traps but will try my own solution

martin-schulze-vireso:
Sigh, Windows does not support SIGUSR1/2. SIGALRM interferes with `sleep`. I chose `SIGABRT` now in the hope of not needing it in the test itself. It may still be interfered with from user code sending it...

martin-schulze-vireso:
The cygwin Bash on the CI Windows also does not sport pkill, so I had to work around that too.

lib/bats-core/formatter.bash
```diff
@@ -70,8 +71,13 @@ function bats_parse_internal_extended_tap() {
             if [[ "$line" =~ $not_ok_line_regexpr ]]; then
                 not_ok_index="${BASH_REMATCH[1]}"
                 test_name="${BASH_REMATCH[2]}"
-                if [[ "$line" =~ $timing_expr ]]; then
-                    bats_tap_stream_not_ok --duration "${BASH_REMATCH[1]}" "$not_ok_index" "$test_name"
+                if [[ "$line" =~ $timeout_line_regexpr ]]; then
+                    not_ok_index="${BASH_REMATCH[1]}"
+                    test_name="${BASH_REMATCH[2]}"
+                    #timeout="${BASH_REMATCH[3]}"
```
martin-schulze-vireso, line 77: Should we pass on information about timed out tests?

martin-schulze-vireso:
@TApplencourt Do you mind giving this a spin to see if it fits your expectations. I'd be especially keen on feedback about the timeout reporting in the pretty formatter (with and without timing information).

martin-schulze-vireso:
I decide that this closes #396, have a look over there for the rationale.

martin-schulze-vireso (approved):

PR #545, opened 2022-02-04
bats: add bazel support

jonathan-enf:
Motivation: my real goal was to fix the path-to-bats issue created when bats is invoked in certain bazel environments (a naive sh_binary rule will place bats in `${BATS_ROOT}/bats_core/bats` instead of `${BATS_ROOT}/bats_core/bin/bats`, but while I was at it, I figured I would contribute my BUILD.bazel file as well.

To demonstrate this bug fix, I'm also contributing all of the additional bazel-related files necessary to integrate bats_core into a bazel build flow.  This will save your users from having to roll-their-own -- they will be able to simply add a dependency on bats_core to their WORKSPACES file and go.  The files I added to make this happen are:

* `WORKSPACES` defines the bazel workspace (what dependencies bats needs to run. In this case, we're just importing a standard bazel library that lets us do shell-escaping of command line arguments correctly.
* `BUILD.bazel` defines the sh_binary rule to compile the "bats" tool in bazel's hermetic build environment, for use by other bazel rules.
* `bats.bzl` is a bazel library that defines a `bats_test` rule that users can add to their own BUILD files to quickly run bats tests.  Under the hood, this library depends on the `bats` tool compiled in the previous file.
* `test/BUILD.bazel` is an example of importing the bats_test rule and using it to run the entire bats regression suite.

TESTED:
`bazel test //test:bats_test_suite` executes correctly and passes.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -4,7 +4,10 @@
 
 sh_library(
     name = "bats_lib",
-    srcs = glob(["libexec/**"]),
+    srcs = glob([
+        "libexec/**",
+        "lib/**",
+    ]),
 )
 
 sh_binary(
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,17 @@
+# BATS: Bash Automated Testing System
+# This is the BUILD.bazel file for compiling the bats binary.
+# Additional rules might need to be added to enable the other bats libraries.
+
+sh_library(
+    name = "bats_lib",
+    srcs = glob(["libexec/**"]),
+)
+
+sh_binary(
+    name = "bats",
+    srcs = ["bin/bats"],
+    deps = [
+        ":bats_lib",
+    ],
+    visibility = ["//visibility:public"],
+)
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -10,8 +10,8 @@ sh_library(
 sh_binary(
     name = "bats",
     srcs = ["bin/bats"],
+    visibility = ["//visibility:public"],
     deps = [
         ":bats_lib",
     ],
-    visibility = ["//visibility:public"],
 )
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -4,21 +4,12 @@ bats_test(
     name = "bats_test_suite",
     srcs = ["bats.bats"],
     deps = [
-        "concurrent-coordination.bash",
-        "file_setup_teardown.bats",
-        "install.bats",
-        "junit-formatter.bats",
-        "parallel.bats",
-        "pretty-formatter.bats",
-        "root.bats",
-        "run.bats",
-        "suite.bats",
-        "test_helper.bash",
-        "trace.bats",
         # Because of a space in a filename, this directory must be handled
         # specially (like this, instead of a glob):
         "fixtures/bats/.",
     ] + glob([
+        "*.bats",
+        "*.bash",
         "fixtures/*",
         "fixtures/bats/evaluation_count/**",
         "fixtures/bats/issue-433/**",
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -7,17 +7,11 @@ bats_test(
         # Because of a space in a filename, this directory must be handled
         # specially (like this, instead of a glob):
         "fixtures/bats/.",
-    ] + glob([
-        "*.bats",
-        "*.bash",
-        "fixtures/*",
-        "fixtures/bats/evaluation_count/**",
-        "fixtures/bats/issue-433/**",
-        "fixtures/file_setup_teardown/**",
-        "fixtures/junit-formatter/**",
-        "fixtures/parallel/**",
-        "fixtures/run/**",
-        "fixtures/suite/**",
-        "fixtures/trace/**",
-    ]),
+    ] + glob(
+        ["**"],
+        exclude = [
+            "BUILD.bazel",
+            "fixtures/bats/*",
+        ],
+    ),
 )
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -1,0 +1,32 @@
+load("//:bats.bzl", "bats_test")
+
+bats_test(
+    name = "bats_test_suite",
+    srcs = ["bats.bats"],
+    deps = [
+        "concurrent-coordination.bash",
+        "file_setup_teardown.bats",
+        "install.bats",
+        "junit-formatter.bats",
+        "parallel.bats",
+        "pretty-formatter.bats",
+        "root.bats",
+        "run.bats",
+        "suite.bats",
+        "test_helper.bash",
+        "trace.bats",
+        # Because of a space in a filename, this directory must be handled
+        # specially (like this, instead of a glob):
+        "fixtures/bats/.",
+    ] + glob([
+        "fixtures/*",
+        "fixtures/bats/evaluation_count/**",
+        "fixtures/bats/issue-433/**",
+        "fixtures/file_setup_teardown/**",
+        "fixtures/junit-formatter/**",
+        "fixtures/parallel/**",
+        "fixtures/run/**",
+        "fixtures/suite/**",
+        "fixtures/trace/**",
+    ]),
+)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,18 @@
+workspace(
+    name = "bats_core",
+)
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "bazel_skylib",
+    sha256 = "c6966ec828da198c5d9adbaa94c05e3a1c7f21bd012a0b29ba8ddbccb2c93b0d",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
+        "https://github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
+    ],
+)
+
+load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
+
+bazel_skylib_workspace()
--- a/bats.bzl
+++ b/bats.bzl
@@ -1,4 +1,16 @@
 # Bazel rules for running bats tests.
+#
+# Example of use, assuming bats_core is loaded in the @bats_core workspace:
+#
+# load("@bats_core//:bats.bzl", "bats_test")
+#
+# bats_test(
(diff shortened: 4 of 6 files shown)


martin-schulze-vireso:
Please add a test that shows how this should be used and would fail without your new code.

jonathan-enf:
> Please add a test that shows how this should be used and would fail without your new code.

I've added a "bats_test" rule in the enclosed bats.bzl file.
I've added a test/BUILD.bazel file that runs the entire bats regression suite in bazel, using a "bats_test" rule.

Without my bin/bats change:
  - these rules would pass in a normal bazel execution environment, but would fail in certain remote execution bazel configurations.

With my bin/bats change:
  - this rule passes when executed both locally and remotely.

Configuring bazel for remote execution is beyond the scope of this PR.

test/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
+load("//:bats.bzl", "bats_test")
+
+bats_test(
+    name = "bats_test_suite",
+    srcs = ["bats.bats"],
+    deps = [
```
martin-schulze-vireso, line 6: That looks quote involved and I wonder why the internal .bash files need to be added. I'd eyxpect this to be part of the "install" script?

Anyways, I think we should have a dedicated test file to avoid failing tests due to missing dependencies, e.g. when new fixtures are added.

martin-schulze-vireso (changes requested):
Thanks for the tests. I am no bazel user so please bear with me If there are some obvious mistakes. I am a still unsure how to handle this, as I'd Like to avoid pulling everything into  the core repo but also don't want untested code.

Do you have a publicly visible example of the described error?

BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+# BATS: Bash Automated Testing System
+# This is the BUILD.bazel file for compiling the bats binary.
+# Additional rules might need to be added to enable the other bats libraries.
+
+sh_library(
+    name = "bats_lib",
+    srcs = glob(["libexec/**"]),
```
martin-schulze-vireso, line 7: What about `lib/bats-core/*`?

jonathan-enf:
> > Without my bin/bats change:
> > 
> > * these rules would pass in a normal bazel execution environment, but would fail in certain remote execution bazel configurations.
> 
> It is a bit unfortunate, that we don't have a reproducer. This defeats the purpose of having a regression test. Since I also don't have seen the problem, I have to take your word for its existence and the fix.

I agree, this isn't an ideal situation.  Setting up a public remote-execution bazel server would be a huge task, so unfortunately I'm hoping that you'll trust me on this one.  For what it's worth, the change that you're trusting me for is a very small one: the change that adds flexibility in stripping /bin/bats from the BATS_ROOT path.

> > Motivation: my real goal was to fix the path-to-bats issue created when bats is invoked in certain bazel environments (a naive sh_binary rule will place bats in ${BATS_ROOT}/bats_core/bats instead of ${BATS_ROOT}/bats_core/bin/bats, but while I was at it, I figured I would contribute my BUILD.bazel file as well.
> 
> I think there is a misunderstanding here. There are two `bats` binaries. One in `[/usr/]bin/bats` and one in `${BATS_ROOT}/libexec/bats-core/`. The former should use its own path to extract `${BATS_ROOT}`, which is what you changed.

It's not a misunderstand, it's a weird bazel-ism.  Bazel has a whole system for importing third-party packages into a BUILD environment, that basically involves: configuring bazel to fetch a package, checking the SHA, and then providing a BUILD file to compile any libraries or tools inside that package.  This enables us to build and run the bats tool inside of bazel's hermetic build environment, and is instead of installing the bats binary in a container or image.

There's just one catch: bazel uses a "sh_binary" rule to "compile" bash scripts into executable tools.  Internally, this is just about getting the dependencies right, but the upshot as far as this PR is concerned is that sometimes the "compiled" bats binary lands in the root directory of the workspace, rather than the bin directory.

Building the bats "binary" looks like this with bazel:

```
$ bazel build @bats_core//:bats
Running: /home/jonathan/.cache/bazelisk/downloads/bazelbuild/bazel-4.2.0-linux-x86_64/bin/bazel --bazelrc=/tmp/bazelrc.kCZKL3 build @bats_core//:bats
INFO: Invocation ID: eb3a4172-cd4a-4f52-8c02-7417fa5674f6
INFO: Analyzed target @bats_core//:bats (51 packages loaded, 229 targets configured).
INFO: Found 1 target...
Target @bats_core//:bats up-to-date:
  bazel-bin/external/bats_core/bats
INFO: Elapsed time: 9.109s, Critical Path: 0.07s
INFO: 4 processes: 4 internal.
INFO: Build completed successfully, 4 total actions
```

That indented line above is the path to the "compiled" bats binary, which is in a special "bazel-bin" temporary directory that bazel manages, within the user's work directory.  Bazel names this binary based on the name of the sh_binary rule, and there is no way to make that name "bin/bats."  "bats" is the best that can be done.

For examples of all of this, I've contributed all of the bazel support files that you might need to run bazel:

WORKSPACES defines the bazel workspace (what dependencies bats needs to run.  In this case, we're just importing a standard bazel library that lets us do shell-escaping of command line arguments correctly.

BUILD.bazel defines the `sh_binary` rule I was talking about before that "compiles" the bats tool for use within bazel.

bats.bzl is a bazel library that users can import, to run bats tests from within their own BUILD files.  It defines a "bats_test" rule that makes use of the "bats" binary compiled in the prior BUILD.bazel file.

test/BUILD.bazel is an example of importing the bats_test rule and using it to run the entire bats regression suite.

These files mean that other bazel users who want to make use of bats won't have to roll-their-own solution: they'll be able to simply add bats_core to their WORKSPACES file and start writing bats_test rules.

To reproduce all of this yourself:
  1. check out this branch
  2. install bazel
  3. run "bazel test test:all"

> Which should deduce `BATS_ROOT=/usr`. From what you wrote I'd gather something like this:

In the bazel environment BATS_ROOT ends up being `$PATH_TO_GIT_REPO/bazel-bin/external/bats_core`

> Is that correct? If so, the problem seems to be that bazel changes the relative paths between bin/bats an the other scripts, right?

Very close.  bazel is "compiling" bin/bats and producing a bats executable in the BATS_ROOT directory.

I hope I managed to answer your questions.  bazel is really, really different and weird and underdocumented.  I would hate this tool if it wasn't for the fact that it actually does its job very well, once you get past the learning curve.

BUILD.bazel
jonathan-enf: Good catch, fixed.  My previous tests didn't have any dependencies there, but it's definitely needed.

test/BUILD.bazel
jonathan-enf: Normally, this would all be replaced with `glob(["**"])` which would capture everything.

Unfortunately, this can't be done because bazel's "glob" function will report an error if it globs a file containing whitespace in the filename.  I have worked around it by "exploding" the glob so that glob is only used for the directories that don't contain a file with whitespace in the name.

jonathan-enf: I've modified this rule to be a little easier to maintain.

martin-schulze-vireso:
Obviously, I would prefer to fix that via bazel not via bats. As you said, bazel seems to be a bit ideosyncratic about this, so this might not be an option. I am wondering if there are other tools with the same problem. Surely, we're not the only one who use relative paths from the entry point executable.

However, I'd like to avoid getting ever more complex workarounds if the next guy comes around with a new build system that thinks its okay to put everything into another folder. Your change already changes the details of BATS_PATH to BATS_ROOT conversion. That might trip up other systems that move around `bin/bats` (without breaking the existing code). I think a cleaner solution would be to allow setting BATS_ROOT when calling `bats`, to allow for the calling system to override the automatic detection. This allows for localized overrides by the affected systems.

However, for your usecase this leaves the questions: can bazel control the environment when bats is actually executed?

martin-schulze-vireso:
I'd like to get this moving again to avoid piling up too many conflicts.

After browsing the code it seems trivial to make bazel insert BATS_ROOT into the template. This leaves the question: do we know where bazel installed libexec/lib?

Assuming we can, I would propose following action plan:

1. replace your fix with `export BATS_ROOT=${BATS_ROOT:-${BATS_PATH%/*/*}}`
2. remove the tests, they don't serve a purpose if they will not showcase the actual problem
3. move the example project into its own repo, we might host it here or just reference yours
4. if possible, move the hosting of bazel related files into an extra repository similarly to [homebrew](https://github.com/bats-core/homebrew-bats-core).

This should leave us with only `bin/bats` (plus tests) being modified in this repo.

The goals of this approach are:

1. get a universal override mechanism that can accommodate other systems as well
2. avoid cluttering core with too many files

jonathan-enf:
Hi Martin,

> I'd like to get this moving again to avoid piling up too many conflicts.

I'd like to move on this as well, I'm just juggling a little work/life balance at the moment.

> After browsing the code it seems trivial to make bazel insert BATS_ROOT into the template. This leaves the question: do we know where bazel installed libexec/lib?

I had high hopes for this approach, but when I tried it I found that there was exactly one regression test that failed.  It seemed that even if bazel sets BATS_ROOT correctly, the environment variable gets lost in one of the nested invocations of bats.  I stared at the problem for a bit but I wasn't able to root-cause the issue.

On the other hand, the fix that just infers BATS_ROOT more flexibly worked correctly and all regressions passed.

> Assuming we can, I would propose following action plan:
> 
> 1. replace your fix with `export BATS_ROOT=${BATS_ROOT:-${BATS_PATH%/*/*}}`
> 2. remove the tests, they don't serve a purpose if they will not showcase the actual problem
> 3. move the example project into its own repo, we might host it here or just reference yours
> 4. if possible, move the hosting of bazel related files into an extra repository similarly to [homebrew](https://github.com/bats-core/homebrew-bats-core).

Sure, that's reasonable, if we can just fix that one regression test.  If that proves too difficult, I hope you'll consider accepting making the flexible inference of BATS_ROOT as an alternative.

I'm under the gun at work right now, but I can put some time into this next week.

Cheers,

- Jonathan.

martin-schulze-vireso:
Thanks for your response. I understand that, I just wanted to make sure this does not fall through the cracks.

Can you name the failing test, so I can have a look?

martin-schulze-vireso:
I added a test that should showcase the problem and works as intended. Can you verify that it works for your usecase?

martin-schulze-vireso:
I am closing this since there has been no feedback for a while. Feel free to reopen with new information.

jmelahman:
FYI, https://github.com/filmil/bazel-bats is a working example of integrating bats within the bazel build system. Feel free to file a ticket there as well if it's missing some feature (I only briefly skimmed this ticket, so not entirely sure what the issue is).

Also FWIW, filed https://github.com/bats-core/bats-core/issues/563 in case the bats-core team is interested in implementing this under the bats-core organization.

PR #574, opened 2022-05-05
Feature/option exclude

svensson84:
hi dear bats community

i like your testing tool. so i thought i could contribute as well. so here is my pull request / little feature :)

bats has been enhanced with the ability to exclude file(s)/directories for testing.
i do know, that's not the main purpose of a testing system... anyways, there are some viable use cases for this new option.

i tried to write clean code, hoping it can be easily reviewed and maintened. what is still missing, are some bats tests for the new option. i had my own test setup running. i will be happy to write bats tests for the new feature as soon as there is interest in merging my code.

impressions:
[screenshot of bats --exclude](https://pasteboard.co/fTxQOqXpVZ80.png)

greetings from switzerland
sven 

btw:
- no echo's
- printf -v applied for functions
- subshells prevented as much as possible
- performance was in focus
- man pages up to date
- workflow tests running

- [ x ] I have reviewed the [Contributor Guidelines][contributor].
- [ x ] I have reviewed the [Code of Conduct][coc] and agree to abide by it


libexec/bats-core/bats
```diff
@@ -80,6 +84,98 @@ expand_path() {
...
+}
+
+# argument suite_filename
+is_excluded() {
+  if [[ -z "$exclude_files" ]]; then
+    return 1
+  fi
+  
+  local suite_filename="$1"
+  read -ra EXCLUDE_FILES <<< "$exclude_files"
```
martin-schulze-vireso, line 99: I'd prefer explicitly passed parameters where possible. Can we use $2 onwards instead of $exclude_files?

libexec/bats-core/bats
```diff
@@ -80,6 +84,98 @@ expand_path() {
...
+
+  for exclude_file in "${EXCLUDE_FILES[@]}"
+  do
+    # 'file check' needed: $exclude_file can contain dir or file filter
+    if is_bats_file "$exclude_file"; then
```
martin-schulze-vireso, line 104: So what about the dirs?

libexec/bats-core/bats
```diff
@@ -363,13 +464,23 @@ for filename in "${arguments[@]}"; do
 
   if [[ -d "$filename" ]]; then
     shopt -s nullglob
-    if [[ "$recursive" -eq 1 ]]; then
+    if [[ "$recursive" == "0" ]]; then
```
martin-schulze-vireso, line 467: Why?

libexec/bats-core/bats
```diff
@@ -363,13 +464,23 @@ for filename in "${arguments[@]}"; do
...
+      done < <(eval "$find_command" -print0 | sort -z)
     else
       for suite_filename in "$filename"/*."${BATS_FILE_EXTENSION:-bats}"; do
-        filenames+=("$suite_filename")
+        if ! is_excluded "$suite_filename" ; then
+          filenames+=("$suite_filename")
+        fi
       done
     fi
     shopt -u nullglob
```
martin-schulze-vireso, line 486: It might be surprising that singly defined files are not subject to the filter. Arguably, one could expect users to not want to exclude explicitly named files. However, some users might do `bats test/*.bats` which would circumvent the `--exclude` mechanism.

libexec/bats-core/bats
```diff
@@ -363,13 +464,23 @@ for filename in "${arguments[@]}"; do
...
       while IFS= read -r -d $'\0' file; do
-        filenames+=("$file")
-      done < <(find -L "$filename" -type f -name "*.${BATS_FILE_EXTENSION:-bats}" -print0 | sort -z)
+        filenames+=("$file")   
+        # performance boost through piping by /dev/fd/63 instead of
+        # creating a temporary file for find's results.
+        # => pipe [<(..)] and redirect [<] '$find_command' to standard input.
+        # for `read` using c-style string termination '\0' [-print0]
+        # eval is a built-in command, no performance issue's expected
+      done < <(eval "$find_command" -print0 | sort -z)
```
martin-schulze-vireso, line 478: Can we have some performance data that drove this decision? While I am all for performance, I am even more for readable code. This is hardly on bats' hot path.

libexec/bats-core/bats
```diff
@@ -33,6 +33,10 @@ HELP_TEXT_HEADER
                             A two character string of code quote delimiters
                             or 'custom' which requires setting $BATS_BEGIN_CODE_QUOTE and 
                             $BATS_END_CODE_QUOTE. Can also be set via $BATS_CODE_QUOTE_STYLE
+  -e, --exclude <files>     Exclude file(s) from processing. File can also be a
+                            directory itself. <files> are separated by space 
+                            and enclosed in quotes, e.g. 
+                            `--exclude 'foo.bats bar baz qux.bats'`
```
martin-schulze-vireso, line 39: Maybe we should go for multiple --exclude statements to allow for spaces in filenames (although that is hopefully rare).

martin-schulze-vireso (changes requested):
Thanks for your contribution. 

I am unsure where exactly this fits with the current and planned filtering mechanisms. Maybe you can outline a concrete use case where you wanted this?

We already have:
- --filter (which does not really interact with this PR)
- skip in setup_file (which does not deal with excluding whole directories)
- specifying individual files

We also have planned a tagging mechanism that might solve some of your intended use cases. 

Additionally, I am toying with the idea of offering an interface where you can define a callback function that decides per test if it should be executed and should get all information pertaining to the test. This would be a catch all solution but is obviously quite verbose.

Finally, I must say that the 1.7.0 release is already behind schedule so I am afraid that we will have to mvo this PR into the next one. This can lead to merge conflicts although I hope there won't be any hairpullers

svensson84:

thanks a lot for your qualitative feedback and code review. i am gonna reply your code comments as well.

> I am unsure where exactly this fits with the current and planned filtering mechanisms. Maybe you can outline a concrete use case where you wanted this?

...for sure:

***main use case*** 
multi-module project  (pom's, git submodules, whatever...)

**setup**
* /project
   * core  
      * *invoke_bats.sh*  
   * batch
      * *invoke_bats.sh*  
      * core
         * *invoke_bats.sh*  
   * migration
      * *invoke_bats.sh*  
      * core
         * *invoke_bats.sh*  
   
**dependencies**
batch and migration both uses core module

**precondition**
each module has its own test suite.

**actor** 
batch module

**trigger**
calling test suite: batch/invoke_bats.sh

**main flow** 
1. all found tests starting from the very batch root directory should be processed (recursive manner)
2. all tests from core module should be left out


> 
> We already have:
> 
> * --filter (which does not really interact with this PR)
> * skip in setup_file (which does not deal with excluding whole directories)
> * specifying individual files

i am aware of the first two of these options. what do you mean exactly by the third one? so far, i do not think that they meet the use cases i was reconsidering.

> We also have planned a tagging mechanism that might solve some of your intended use cases.

sounds very interesting! cool.
 
> Additionally, I am toying with the idea of offering an interface where you can define a callback function that decides per test if it should be executed and should get all information pertaining to the test. This would be a catch all solution but is obviously quite verbose.

the following technical approaches could possibly also be reconsidered 
- configuration files in hierarchical manner (as git does with system, global, local). similar approach could be used for filter/exclusion setups, why not having a ***.batsconfig***
- on a lower architectural layer working with (parameterized) annotations instead of callbacks: **@skip**, **@filter[*-lib]**

imho, i think both approaches are very extendable/flexible for supporting the growth of the code base.

> Finally, I must say that the 1.7.0 release is already behind schedule so I am afraid that we will have to mvo this PR into the next one. This can lead to merge conflicts although I hope there won't be any hairpullers

no problem at all. i am very glad i got feedback from you. thanks again. 
if you need any help, maybe i can contribute/develop in other upcoming features/releases. let me know!

best wishes.
sven

libexec/bats-core/bats
```diff
@@ -80,6 +84,98 @@ expand_path() {
   printf -v "$result" '%s/%s' "$dirname" "${path##*/}"
 }
 
+# argument: exclude_literal
+is_bats_file() {
+  [[ "$1" == *".bats"* ]]
```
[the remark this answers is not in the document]
svensson84: yes, i agre with you.

libexec/bats-core/bats
svensson84: yes, we can :-) a matter of taste.

libexec/bats-core/bats
svensson84: they should survive this exclusion, else the whole construct of a test-suite could be skipped according the given filter.
keep in mind: this method takes a specific suite_filename as argument (also documented above method).

libexec/bats-core/bats
svensson84: i ask myself as well. it's a bad leftover of a small refactoring.
just wanted to make the global var **_recursive_** acting conform to shell standards. 0=true; other false
did not like that 0 is true if it is in recursive mode.

my preferred way would be:
recursive=1  # set globally to false

and the commented if written as
if  [[ "$recursive" -eq 0]]; then

libexec/bats-core/bats
svensson84: i am totally on your side with clean code (readability).

that's why i thought it would be useful to briefly document how this loop works. 
i have not changed this existing main construct. only the find is now dynamically built, which led to the use of eval. if there are better variants, i am very open and happy to learn. thanks.

libexec/bats-core/bats
svensson84: i prefer one exclude statement
we could introduce an escape character for spaces.

libexec/bats-core/bats
```diff
@@ -84,94 +84,111 @@ expand_path() {
...
-    return 1
-  fi
-  
-  local suite_filename="$1"
-  read -ra EXCLUDE_FILES <<< "$exclude_files"
+# arguments:
+#  filtered_suite_filenames ('out' param)
+#  exclude_files 
+#  suite_filenames
+filter_suite() {
```
svensson84, line 96: **naming:**
implemented the filter function of suite filenames more explicit.

**logic:**
almost the same as before except that EXCLUDE_FILES only need to be created once.
=> looping in line 480 moved one level up into this function (sorry cannot create a permalink to the line)

libexec/bats-core/bats
```diff
@@ -84,94 +84,111 @@ expand_path() {
...
+  local result="$1"
+  local exclude_files="$2"
```
svensson84, line 98: as requested... explicit parameters in functions and no more accessing of global vars.

libexec/bats-core/bats
```diff
@@ -84,94 +84,111 @@ expand_path() {
...
-# - using find combined with -prune instead of piping with grep/awk etc.
-# - use of sub processes limited to the absolute minimum.
-#
-build_find_command_string() {
+# arguments:
+#   find_command ('out' param)
+#   exclude_files
+build_find_command() {
   local result="$1"
+  # shellcheck disable=SC2178
```
svensson84, line 168: disabled because i don't really agree with the warning. maybe i do miss something?!

libexec/bats-core/bats
```diff
@@ -84,94 +84,111 @@ expand_path() {
...
+  local exclude_files="$2"
+  
+    # shellcheck disable=SC2128
```
svensson84, line 171: disabled because i don't really agree with the warning. maybe i do miss something?!

libexec/bats-core/bats
```diff
@@ -464,25 +481,26 @@ for filename in "${arguments[@]}"; do
...
+      build_find_command 'find_command' "$exclude_files"  
       while IFS= read -r -d $'\0' file; do
-        filenames+=("$file")   
-        # performance boost through piping by /dev/fd/63 instead of
-        # creating a temporary file for find's results.
-        # => pipe [<(..)] and redirect [<] '$find_command' to standard input.
-        # for `read` using c-style string termination '\0' [-print0]
-        # eval is a built-in command, no performance issue's expected
+        filenames+=("$file")
+        # using named pipe [<(..)] and stdin redirection [<] for handling
```
svensson84, line 490: doc simplified ;-)

svensson84:
for the sake of clean PR... even if it does not get merged because of future filtering plans... 15 bats tested added for --exclude feature:

**fixture/directory setup:**

exclude
 ├── a0
 │   └── baz.bats
 ├── b0
 │   └── b1
 │       ├── quux.bats
 │       └── qux.bats
 ├── c0
 │   └── c1
 │       ├── c2
 │       │   └── grault.bats
 │        └── corge.bats
 ├── bar.bats
 └── foo.bats 

well, have a good one. 
sven

martin-schulze-vireso:
Could you have a look how well the new tagging mechanism deals with your request? I think a combination of `# bats file_tags=` and `--filter-tags !file-tag` should come close to what you want to achieve here.

svensson84:
> Could you have a look how well the new tagging mechanism deals with your request? I think a combination of `# bats file_tags=` and `--filter-tags !file-tag` should come close to what you want to achieve here.

yeah. the tagging mechanism is much more verbose, hence much more flexible as well. i do like that mechanism, specially that is possible to declare conjunctions and disjunctions! great work, well done! 
... and yes, with the new tagging mechanism one can achieve the same goal as with my requested “--exclude" argument! 

once again, well done! 
have a good one, 
sven

martin-schulze-vireso:
Thanks, I understand that to mean we don't need this PR anymore.

PR #583, opened 2022-05-07
Fix: `run --keep-empty-lines` reported one line for empty output, should be zero

### LINKED ISSUE #573: run --keep-empty-lines true: `[ "${#lines[*]}" -eq 1 ]` instead of `[ "${#lines[*]}" -eq 0 ]`
**Describe the bug**

When using `run --keep-empty-lines` with a command that produces no output (e.g. `true`), `bats` reports a wrong line count of `1` instead of `0`.

**To Reproduce**

Copy the following lines to to a shell which has `bats` in PATH:
``` bash
cat >test.bats <<'EOF'
@test 'run, line count ok' {
    run true
    [ "${#lines[*]}" -eq 0 ] # ok!
}

@test 'run --keep-empty-lines, wrong line count' {
    run --keep-empty-lines true
    >&3 printf -- '# lines: %d\n' "${#lines[*]}"
    [ "${#lines[*]}" -eq 1 ] # should fail!
    [ "${#lines[*]}" -eq 0 ] # should pass!
}
EOF
bats test.bats
```
The 2nd test fails in the `[ "${#lines[*]}" -eq 0 ] # should pass!` line where it should have failed on `[ "${#lines[*]}" -eq 1 ] # should fail!`.

**Expected behavior**

Line count (`${#lines[*]}`) from `run --keep-empty-lines` called with command that has no output should be `0`.

 - Bats Version: `v1.6.0-42-g5c7a967` (from `git describe`)
 - OS: Linux (Debian 11)
 - Bash version: `5.1.16(1)-release (x86_64-pc-linux-gnu)`

martin-schulze-vireso:
Fixes #573

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 * don't show empty lines as `#` with pretty formatter  (#561)
 * prevent `teardown`, `teardown_file`, and `teardown_suite` from overriding bats'
   exit code by setting `$status` (e.g. via calling `run`) (#581, #575)
+* `run --keep-empty-lines` now reports 0 lines on empty `$output` (#583)
 
 #### Documentation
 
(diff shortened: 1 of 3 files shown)


ktetzlaff:
Thanks!

PR #628, opened 2022-08-08
Add default path for 'BATS_LIB_PATH' environment variable'

hwittenborn:

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -101,6 +101,7 @@ export BATS_TMPDIR="${TMPDIR:-/tmp}"
 BATS_TMPDIR=${BATS_TMPDIR%/} # chop off trailing / to avoid duplication
 export BATS_RUN_TMPDIR=
 export BATS_GUARANTEED_MINIMUM_VERSION=0.0.0
+export BATS_LIB_PATH=${BATS_LIB_PATH-/usr/lib/bats}
 
 if [[ ! -d "${BATS_TMPDIR}" ]];then
   printf "Error: BATS_TMPDIR (%s) does not exist or is not a directory" "${BATS_TMPDIR}" >&2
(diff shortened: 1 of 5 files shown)


martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for your contribution.

hwittenborn:
@martin-schulze-vireso: Any idea when the next release will be made? I'd like to start using this in my tests but I need to deploy it via my Debian package first.

martin-schulze-vireso:
Well, the deadline for 1.8.0 is this month. However, I want to close all issues in the milestone.

libexec/bats-core/bats
```diff
@@ -101,6 +101,7 @@ export BATS_TMPDIR="${TMPDIR:-/tmp}"
 BATS_TMPDIR=${BATS_TMPDIR%/} # chop off trailing / to avoid duplication
 export BATS_RUN_TMPDIR=
 export BATS_GUARANTEED_MINIMUM_VERSION=0.0.0
+export BATS_LIB_PATH=${BATS_LIB_PATH-/usr/lib/bats}
```
jcassee, line 104: Sorry for the drive-by comment, but shouldn't this be `${BATS_LIB_PATH:-/usr/lib/bats}`? (Note the colon.)

martin-schulze-vireso: I only translated this from the original Code but I assumed the intention is to allow for users to specify an empty search path. This would mit ve possible with colon.

jcassee: Sure, but the syntax for that is `${var:-default}`, not `${var-default}`, right? ([Bash docs](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html))

martin-schulze-vireso: No, :- is alternative value for unset or empty, - only replaces unset variables.

jcassee: I did not know that. Thank you for the explanation!

PR #641, opened 2022-08-24
ci(lint): add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR.

I saw that your scripts are in great shape and that you already have shell linter in place, but I still think that you might find `differential-shellcheck` action useful. It is able to produce reports in SARIF format, GitHub understands this format and is able to display it nicely as PR comment and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

---

### DIFF
--- a/.github/workflows/differential_shellcheck.yml
+++ b/.github/workflows/differential_shellcheck.yml
@@ -1,0 +1,26 @@
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
+      pull-requests: write
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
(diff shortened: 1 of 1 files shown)


martin-schulze-vireso:
Thanks for your contribution. I am afraid we won't be able to incorporate this into our project (yet?). While I see the benefit of having shellcheck annotations inlined automatically on PRs, there are several problems with incorporating this into our project right now:

1. it is a third party action, which opens up potential security holes
2. I don't have the time and expertise to vet it's code
3. judging form the amount of stars, it is not yet widely in use, so we cannot rely on it being vetted by a large community of users

How can these issues be addressed? The first one can only really be solved by GitHub, if at all but if it were the only issue, I would be more inclined to try your code. The second one might solve itself in time, either by me getting up to speed in this area or by another maintainer with more expertise. The third one should also solve itself over time, since I am sure many projects will find this useful.

With that, I congratulate you for your nice work and hope to see you again when these problems have sorted themselves out over time.

PR #646, opened 2022-09-06
Add github workflow to test-build RPM

hecko:
Builds on Ubuntu, but this should not affect the test build outcome.

### DIFF
--- a/.github/workflows/rpm_build.yml
+++ b/.github/workflows/rpm_build.yml
@@ -1,0 +1,15 @@
+name: Build RPM
+
+# Controls when the action will run. 
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  rpm:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build RPM on Ubuntu
+        run: |
+          sudo apt install rpm
+          sudo mkdir -p /root/rpmbuild/SOURCES/ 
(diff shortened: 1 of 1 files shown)


martin-schulze-vireso:
Please add install and bats test usage, so we see the packaged version works.

hecko:
> Please add install and bats test usage, so we see the packaged version works.

Not possible. Github does not offer rpm compatible OS. I can try to install rpm on ubuntu but this would be unreliable test.

martin-schulze-vireso:
Are there compatible docker images?

hecko:
https://hub.docker.com/_/almalinux

Im honestly not very sure whether this will give any additional value to the rpm build since the rpm is not signed and published anywhere. I understand your thought process, but Im unsure of the invested effort into this particular thing at this very moment. If you really want to publish the RPM and get the package into any of the major distributions the local Docker test will not suffice anyways.

martin-schulze-vireso:
Having a succesful package build is No guarantee that the packaged software works when installed. The Idea is to have an end to end test that shows the Bats installed from RPM works as intended.

hecko:
> Having a succesful package build is No guarantee that the packaged software works when installed. The Idea is to have an end to end test that shows the Bats installed from RPM works as intended.

Can you recommend a command to run to test the functionality of bats once installed on the system?

hecko:
migrated to #648

PR #645, opened 2022-09-06
Fix rpm build spec file to correctly build on Rocky Linux release 8.6

hecko:
I have fixed the RPM build as I really need the RPM for my dev env. This has been fully tested on Rocky Linux 8.6, which should be binary compatible with RHEL8.6 and Alma Linux 8.6

The patch fixes the following errors:
```
[root@roxana-dev rpm]# rpmbuild -v -bb bats.spec
...
RPM build errors:
    bogus date in %changelog: Tue Jul 08 2018 mbland <mbland@acm.org> - 1.1.0-1
    Installed (but unpackaged) file(s) found:
   /usr/lib/bats-core/common.bash
   /usr/lib/bats-core/formatter.bash
   /usr/lib/bats-core/preprocessing.bash
   /usr/lib/bats-core/semaphore.bash
   /usr/lib/bats-core/test_functions.bash
   /usr/lib/bats-core/tracing.bash
   /usr/lib/bats-core/validator.bash
   /usr/lib/bats-core/warnings.bash
```

- [x] I have reviewed the [Contributor Guidelines][contributor].
- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md
[coc]:         https://github.com/bats-core/bats-core/blob/master/docs/CODE_OF_CONDUCT.md

PR #648, opened 2022-09-07
RPM build check

hecko:
Workflow that builds, installs and test-runs bats on AlmaLinux 8

Migrates #646 from Ubuntu runner to AlmaLinux container
Also contains #645 


martin-schulze-vireso:
I made the install use the local version, because it would download the latest release otherwise. The test failures are due to install.sh not being installed (which is expected) and due to something with alma not working in the timeout test. This can be reproduced without using RPM, so I think it is unrelated.

hecko:
Is there anything else I can do here to support this merge?

martin-schulze-vireso:
I only need to find time to finish this. I am thinking about using the new tagging feature to exclude the setup tests.

martin-schulze-vireso:
Thanks for your contribution.

hecko:

Thanks for maintaining bats so well.

PR #657, opened 2022-09-25
Fix infinite wait when invoking "bats -j5" instead of "bats -j 5"

marc-hb:
bats's option parser translates "-j5" to "-j -5". This caused a negative number of "slots" (threads) to be available and the test to silently hang forever. Not nice.

Add a $num_jobs sanity check to fail immediately like this:

```
Invalid number of jobs: -5
tests.bats
   bats warning: Executed 0 instead of expected 3 tests

3 tests, 0 failures, 3 not run
```

Non-numeric arguments (e.g.: -jf) evaluate to zero which also fails the check.

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

### DIFF
--- a/test/parallel.bats
+++ b/test/parallel.bats
@@ -5,6 +5,9 @@ bats_require_minimum_version 1.5.0
 load test_helper
 fixtures parallel
 
+# shellcheck disable=SC2034
+BATS_TEST_TIMEOUT=10 # only intended for the "short form ..."" test
+
 setup() {
   type -p parallel &>/dev/null || skip "--jobs requires GNU parallel"
   (type -p flock &>/dev/null || type -p shlock &>/dev/null) || skip "--jobs requires flock/shlock"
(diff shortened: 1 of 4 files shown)


martin-schulze-vireso:
Thanks fior you report and suggested fix. This is a problem with the short flag unpacker and will cause trouble with other value arguments like -f as well. I believe we should also fix this for good.

I can take care of that. Meanwhile, can you please add a test for your fix?

marc-hb:
> This is a problem with the short flag unpacker and will cause trouble with other value arguments like -f as well. I believe we should also fix this for good.

Makes sense, indeed I thought about looking at the options parser. Then I remembered option parsers are always a complicated and ambiguous mess no matter how hard you try whereas I found this quick sanity check was dead simple to add (once I root-caused the issue). Also, I think a little bit of "defensive programming" is always good when infinite loops are involved!

> I can take care of that. 

Good! :-)


> Meanwhile, can you please add a test for your fix?

Depends how long it would take, BATS is completely new to me and I don't have a lot of spare time... Any starting point, similar test you can point me at?

test/parallel.bats
```diff
@@ -208,3 +211,10 @@ check_parallel_tests() { # <expected maximum parallelity>
 @test "BATS_NO_PARALLELIZE_WITHIN_FILE does not work from inside test function" {
   DISABLE_IN_TEST_FUNCTION=1 reentrant_run ! bats --jobs 2 "$FIXTURE_ROOT/must_not_parallelize_within_file.bats"
 }
+
+@test "Short form typo does not run endlessly" {
+  unset BATS_NO_PARALLELIZE_ACROSS_FILES
+  run bats -j2 "$FIXTURE_ROOT/../bats/passing.bats"
```
marc-hb, line 217: If you really want to test the check that was added in this PR then I think it should be:

```suggestion
  run bats -j -2 "$FIXTURE_ROOT/../bats/passing.bats"
```

Otherwise this test will switch to covering a completely different part of the code the moment you refactor the options parser.

marc-hb (commented):
Thanks for the test! Post-merge review...

test/parallel.bats
marc-hb: Additional test submitted in #693

PR #659, opened 2022-09-26
add two distros shipping bats-core to the docs

henning-schild:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 [kac]: https://keepachangelog.com/en/1.0.0/
 [semver]: https://semver.org/
 
+## [Unreleased]
+
+### Added
+
+* added installation instructions for Debian, Fedora, Gentoo, and OpenSUSE (#659)
+
 ## [1.8.2] - 2022-10-19
 
 ### Fixed
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -15,3 +15,4 @@ Versions before v1.2.1 are documented over `there <https://github.com/bats-core/
    gotchas
    faq
    warnings/index
+   support-matrix
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -2,44 +2,18 @@
 Installation
 ============
 
-Supported Bash versions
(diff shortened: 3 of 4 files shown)


henning-schild:
We could think about adding the name of the package the distros did choose, maybe even install instructions. But the already listed distros do not do that so i did not try.

henning-schild:
I bet that adds some value for users, seeing that they can install the package with their distro. It also should add value to this very project, seeing and tracking who ships and appreciates the code.

henning-schild:
in fact we also have opensuse Linux and fedora Linux, possibly more

i will try and complete the list if that gets merged or there are more questions, right now i care about Debian and Gentoo

martin-schulze-vireso:
I think we have to be careful with the wording here. While I assume Bats to be compatible with all major Linux distributions that provide Bash, we only have a small subset of CI tested versions, which are truly "supported".

All other distributions might be listed as providing a package but should not come with the expectation of being tested, at least not from our side.

henning-schild:
I see. In that case you decide what to do with that. Feel free to just close or throw commits on top that list distros in general and list the ones tested by bats itself.

The file is named "installation" but right it talks about tested and supported.

martin-schulze-vireso:
I restructured the text and added some of your suggestions. Please review the new layout to see whether it is still understandable.

henning-schild:
Looking good. You as the maintainer might find that support matrix and which one is tested how important, for a user install instructions are more relevant and the test matrix and coverage and exact versions might be something to look at when running into issues. Or maybe when having to manually install.

So all that should rather move down in the document, as additional but not primary information. Maybe all the way to the end. Hope that makes sense.

henning-schild:
The typical scenario is ... you want to install, you read that, see your distro ... stop reading and install via that distro. Trusting it will do things right for you.

martin-schulze-vireso:
>> Please review the [new layout](https://bats-core--659.org.readthedocs.build/en/659/installation.html) to see whether it is still understandable.
> So all that should rather move down in the document, as additional but not primary information. Maybe all the way to the end. Hope that makes sense.

I restructured a bit further and moved the support matrix into its own document. The link above points to the updated documentation.

henning-schild:
LGTM, thanks!

martin-schulze-vireso:
Thanks for you contribution. Sorry about the confusion, I accidentally closed instead of merging.

PR #665, opened 2022-10-11
Show file path URI in tracing messages

jbriales:

I'm considering a change to how the test error locations are printed in the trace,
so that one can jump to the location in an IDE with typical support.
We could make this a CLI option maybe, to keep things the same by default?

The new output would be:
```
 ✗ Prints full path URI with line location on error
   (in test file file:///home/jesusbriales/.dotfiles/test/misc/bats.bats:181)
```
instead of the current
```
 ✗ Prints full path URI with line location on error
   (in test file ./misc/bats.bats, line 181)
```

Before jumping into adding tests and any other extra work,
is this sth that you could consider merging to master?


martin-schulze-vireso:
I have been thinking about this for some time now. I think we should allow to parametrize this with a flag. There might be other formats as well.

jbriales:
Would you have any suggestions or hints on how to expose this as a flag?
I managed to find the relevant logic for the feature, but I'm not familiar with how flags cascade from the UI deep to this level :)

martin-schulze-vireso:
> Would you have any suggestions or hints on how to expose this as a flag? I managed to find the relevant logic for the feature, but I'm not familiar with how flags cascade from the UI deep to this level :)

Well, the flags are passed on again and again from (libexec/bats-core/) bats to bats-exec-suite to bats-exec-file to bats-exec-test. The parsing code consists of large `case` statements at the beginning of the files. If I find the time, I'll get around to do this myself eventually.

martin-schulze-vireso:
@jbriales Do you have feedback about the naming?

- `--file-reference-format`/`BATS_FILE_REFERENCE_FORMAT`
- `custom`, `comma_line`, `colon_separated`, `url_realpath`

I also had to work around missing realpath on MacOS.

jbriales:
> @jbriales Do you have feedback about the naming?
> 
> * `--file-reference-format`/`BATS_FILE_REFERENCE_FORMAT`
> * `custom`, `comma_line`, `colon_separated`, `url_realpath`

Sweet! Naming is hard, but I think it's good like that :D
I wondered for a moment if `uri` vs `url` was more appropriate here. I thought general non-web `file://`-like URLs tended to be referred to more as URIs (vs URLs). But after checking online, I'm not so sure of the difference anymore :)

Also I would consider if calling it `url`/`uri` only, vs `url_realpath`. What's the current expected behavior wrt e.g. symlinks?
If we keep it simple and just return a full absolute path w/o taking further care for e.g. symlinks (as realpath does by default IIRC), then maybe just `url` (or `url_abspath`) would be more appropriate?

martin-schulze-vireso:
Okay, I renamed:

- `--line-reference-format`/`BATS_LINE_REFERENCE_FORMAT`
- `colon_separeted` -> `colon`
- `url_realpath` -> `uri`

PR #670, opened 2022-10-19
Fix non zero exit code on successful retry

### LINKED ISSUE #660: For tests that pass but which were retried, bats returns non-zero exit code
**To Reproduce**
```sh
echo '
@test "fail once" {
    BATS_TEST_RETRIES=2
    ((BATS_TEST_TRY_NUMBER > 1)) || false
}
' > tst.bats
bats tst.bats
[ "$?" -eq 0 ] && echo 'reported success' || echo 'reported failure'
```

**Expected behavior**

`reported success` should be printed, as per the test output

 - Bats Version: 1.8
 - OS: MacOS
 - Bash version: [e.g. 5.1] 3.2

martin-schulze-vireso:
Fixes #660

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 [kac]: https://keepachangelog.com/en/1.0.0/
 [semver]: https://semver.org/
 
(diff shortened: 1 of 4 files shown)


Kbenge5:
Thank you Karen

PR #674, opened 2022-11-02
Update Dockerfile: add labels and bats org libs

### LINKED ISSUE #518: Include all support libraries in the Docker image
I am trying to setup a Docker-based validation process using `bats`. Our tests make use of `bats-assert` and `bast-support`, and those are not included in the `bats/bats` Docker image.

Include all optional libraries in the Docker image. Alternatively, provide an additional image with those included, if adding more libraries to the base one is not desirable.

We considered building our own image, but having all libraries in the official Docker image is cleaner and has less potential of becoming stale.

**Additional context**
Thanks for an awesome tool!

brokenpip3:

Fix #518 

- Add the most important opencontainer labels
- Do not use root as user
- Add bats-support, file, detik and assert libs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,14 +33,11 @@ RUN /tmp/docker/install_libs.sh detik ${LIBS_VER_DETIK}
 # context where it make sense to cite GNU Parallel).
 RUN apk add --no-cache parallel ncurses && \
     mkdir -p ~/.parallel && touch ~/.parallel/will-cite \
-    && addgroup -g "$GID" -S bats \
-    && adduser -u "$UID" -S bats -G bats -s "/bin/bash" \
-    && mkdir /code && chown bats:bats /code
+    && mkdir /code
 
 RUN ln -s /opt/bats/bin/bats /usr/local/bin/bats
 COPY . /opt/bats/
 
 WORKDIR /code/
-USER bats
 
 ENTRYPOINT ["/tini", "--", "bash", "bats"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ LABEL maintainer="Bats-core Team"
 LABEL org.opencontainers.image.authors="Bats-core Team"
 LABEL org.opencontainers.image.title="Bats"
 LABEL org.opencontainers.image.description="Bash Automated Testing System"
-LABEL org.opencontainers.image.url="https://hub.docker.com/_/bash"
+LABEL org.opencontainers.image.url="https://hub.docker.com/r/bats/bats"
 LABEL org.opencontainers.image.source="https://github.com/bats-core/bats-core"
 LABEL org.opencontainers.image.base.name="docker.io/bash"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,6 @@ ARG bashver=latest
 FROM bash:${bashver}
 ARG TINI_VERSION=v0.19.0
 ARG TARGETPLATFORM
-ARG TEMPDIR="/tmp/build"
-ARG BATS_LIBS_DESTDIR="/usr/lib"
-ARG BATS_LIBS_SUPPORT_VERSION=0.3.0
 
 # https://github.com/opencontainers/image-spec/blob/main/annotations.md
 LABEL maintainer="Bats-core Team"
@@ -19,6 +16,11 @@ LABEL org.opencontainers.image.base.name="docker.io/bash"
 COPY ./docker /tmp/docker
 # default to amd64 when not running in buildx environment that provides target platform
 RUN /tmp/docker/install_tini.sh "${TARGETPLATFORM-linux/amd64}"
+# Install bats libs
+RUN /tmp/docker/install_libs.sh support 0.3.0
+RUN /tmp/docker/install_libs.sh file 0.3.0
+RUN /tmp/docker/install_libs.sh assert 2.1.0
+RUN /tmp/docker/install_libs.sh detik 1.1.0
 
 # Install parallel and accept the citation notice (we aren't using this in a
 # context where it make sense to cite GNU Parallel).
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -3,21 +3,43 @@ ARG bashver=latest
...
 # Install parallel and accept the citation notice (we aren't using this in a
 # context where it make sense to cite GNU Parallel).
 RUN apk add --no-cache parallel ncurses && \
-    mkdir -p ~/.parallel && touch ~/.parallel/will-cite
+    mkdir -p ~/.parallel && touch ~/.parallel/will-cite \
+    && addgroup -S bats \
+    && adduser -S bats -G bats -s "/bin/bash" \
+    && mkdir /code && chown bats:bats /code \
+    # Bats support
+    && mkdir -p ${TEMPDIR}/bats-support ${BATS_LIBS_DEST_DIR}/bats-support/src \
```
martin-schulze-vireso, line 31: Maybe we should pull this out into a separate shell file.

Dockerfile
```diff
@@ -3,21 +3,43 @@ ARG bashver=latest
...
+    && wget -qO- https://github.com/bats-core/bats-support/archive/refs/tags/v${BATS_LIBS_SUPPORT_VERSION}.tar.gz | tar xvz -C ${TEMPDIR}/bats-support --strip-components 1 \
+    && install -Dm755 ${TEMPDIR}/bats-support/load.bash ${BATS_LIBS_DESTDIR}/bats-support/load.bash \
```
martin-schulze-vireso, line 33: I think it would be better to have dedicated install scripts in each project and rely on those for the packaging.
However, I am not even sure if we want to use curl here or use the [checkout action](https://github.com/actions/checkout).

martin-schulze-vireso (commented):
I'd like to use this as an opportunity to get some integration tests between the current bats-core (master) version and the libraries going.

Dockerfile
brokenpip3: >I think it would be better to have dedicated install scripts in each project and rely on those for the packaging.
However, I am not even sure if we want to use curl here or use the [checkout action](https://github.com/actions/checkout).

Honestly I prefer to address both of your comments in a separated bash script instead of using the github CI for a very important reason: adding an external ci step (GH actions) as prerequisite for a dockerfile will prevent to build the docker img in people laptop and from any kind other of CI system.

Let my take a tentative to create a shell script so we can make this Dockerfile more elegant and slim, will ping you asap

martin-schulze-vireso: You're right, I should not have mixed two concerns. This idea came from the wish for integration tests. But that should be solved in a different manner.

brokenpip3:
I have a working solution, will finalize and push later today

Dockerfile
```diff
@@ -4,20 +4,36 @@ FROM bash:${bashver}
...
+LABEL org.opencontainers.image.url="https://hub.docker.com/r/bats/bats"
+LABEL org.opencontainers.image.source="https://github.com/bats-core/bats-core"
+LABEL org.opencontainers.image.base.name="docker.io/bash"
+
 COPY ./docker /tmp/docker
 # default to amd64 when not running in buildx environment that provides target platform
 RUN /tmp/docker/install_tini.sh "${TARGETPLATFORM-linux/amd64}"
-    
+# Install bats libs
+RUN /tmp/docker/install_libs.sh support 0.3.0
```
brokenpip3, line 20: @martin-schulze-vireso  do you prefer the version hardcoded here or as args? (4 args..)

docker/install_libs.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
brokenpip3, line 1: This should be a good comprise to install the libs without pollute the dockerfile

Dockerfile
martin-schulze-vireso: I think we should have args but rely on defaults that are updated with each release. This allows for customization but a progressing default.

Dockerfile
```diff
@@ -4,20 +4,36 @@ FROM bash:${bashver}
...
+RUN /tmp/docker/install_libs.sh file 0.3.0
```
martin-schulze-vireso, line 21: C&P error?

docker/install_libs.sh
```diff
@@ -0,0 +1,48 @@
...
+
+set -o errexit
+set -o nounset
+
+
+LIBNAME="${1:-support}"
+LIVERSION="${2:-0.3.0}"
```
martin-schulze-vireso, line 8: I'd prefer `${1?}/${2?}` instead of a default value.

docker/install_libs.sh
```diff
@@ -0,0 +1,48 @@
...
+
+trap "test -d "${TMPDIR}" && rm -fr "${TMPDIR}"" EXIT
+
+_log() {
+    printf "$(date "+%Y-%m-%d %H:%M:%S") - %s - %s\n" "${1}" "${2}"
+}
+
+create_temp_dirs() {
+    mkdir -p "${TMPDIR}/${1}"
+    [[ ${LIBNAME} != "detik" ]] && mkdir -p "${DESTDIR}/bats-${1}/src" || _log INFO "Skipping src 'cause Detik does not need it"
```
martin-schulze-vireso, line 22: I think an `if else` would be more readable (and less subtle in its semantics)

docker/install_libs.sh
```diff
@@ -0,0 +1,48 @@
...
+
+_log INFO "Starting to install ${LIBNAME} ver ${LIVERSION}"
+_log INFO "Creating directories"
+create_temp_dirs "${LIBNAME}"
+_log INFO "Downloading"
+download_extract_source "${LIBNAME}" "${LIVERSION}"
+_log INFO "Installation"
+install_files "${LIBNAME}"
+_log INFO "Done, cleaning.."
+test -d "${TMPDIR}" && rm -fr "${TMPDIR}"
```
martin-schulze-vireso, line 48: Why duplicate the exit trap?

docker/install_libs.sh
martin-schulze-vireso: Scratch that. I'd just pull up the check for 2 parameters to the top.

docker/install_libs.sh
brokenpip3: done

docker/install_libs.sh
brokenpip3: sorry, done, also add more exit events to the trap

docker/install_libs.sh
brokenpip3: done

Dockerfile
brokenpip3: done

martin-schulze-vireso:
I decided to remove the unprivileged user from the Dockerfile. There are two main reasons for that:

1. this is going into a minor release and may well be a breaking change. (Furthermore, the content of this PR is not really about introducing this security feature.)
2. Dealing with the restricted user in a volume mount is not easy (see the many bats-file test failures above). I already prepared a new entrypoint.sh that would transfer the UID/GID of the volume mounted under `/code` to the bats user but I feel this automatism somewhat reduces the idea of restricting the user in the first place. I believe this must be thought through a bit more. Maybe @sublimino has some ideas on this?

martin-schulze-vireso:
Thanks for your contribution.

brokenpip3:

Sure yw

Issue #672, opened 2022-10-31
bats-core parallel tests assumes GNU parallel for the `--jobs` flag


filmil:
**Describe the bug**
https://github.com/bats-core/bats-core/blob/master/test/parallel.bats#L12 seems to assume that GNU parallel exists on the system. But it never checks that the `parallel` program
is indeed the GNU `parallel` program, vs the other `parallel` from `moreutils` in Ubuntu for example.  (History apparently here: https://www.gnu.org/software/parallel/history.html)

**To Reproduce**
1. See here: https://github.com/filmil/bazel-bats/actions/runs/3246348635/jobs/5567182637

**Expected behavior**
The test passes.

**Environment (please complete the following information):**
 - Bats Version `1.5.0`
 - OS: Ubuntu
 - Bash version: ?

**Additional context**

GNU `parallel` has a flag `--jobs` which the test uses. But, the `moreutils parallel` has only the short form `-j`. So when `test/parallel.bats` tests for the existence of the `parallel` binary, it never checks whether it also supports `--jobs`, leading to test failures if the `parallel` binary is not GNU `parallel`.

https://github.com/filmil/bazel-bats/pull/9#issuecomment-1279459828 explains the details.

PR #691, opened 2023-01-10
Stricter `parallel` Exists Check

qec-pconner:
There are 2 common implementations of `parallel`: GNU parallel (https://www.gnu.org/software/parallel/) and moreutils' parallel (https://manpages.debian.org/testing/moreutils/parallel.1.en.html, https://linux.die.net/man/1/parallel).

The former is much more feature rich and is the actual intended dependency of parallel tests in bats.

The latter is far more common on many operating systems (including Ubuntu).

Currently, when the `parallel.bats` file runs, it checks to see if `parallel` is available and will skip the test if not. Unfortunately, this only checks that there is _a_ `parallel` command available -- not necessarily the correct one. If moreutils parallel is installed, the skip will not occur and the test will later fail in "libexec/bats-core/bats-exec-suite", when trying to run `parallel --keep-order --jobs ...` -- these CLI args are unknown to moreutils `parallel`.

This PR improves the check such that not _any_ `parallel` command will work. This check is extended by checking that the `parallel` command supports `--version`, which distinguishes between GNU parallel (supported) and moreutils parallel (unsupported).

Fixes https://github.com/bats-core/bats-core/issues/672.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -26,6 +26,8 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ### Fixed
 
+* explicitly check for GNU parallel (#691)
+
 #### Documentation
 
 * improved clarity of section about output in free code (#671)
--- a/libexec/bats-core/bats-exec-suite
+++ b/libexec/bats-core/bats-exec-suite
@@ -97,7 +97,7 @@ while [[ "$#" -ne 0 ]]; do
 done
 
 if [[ "$num_jobs" != 1 ]]; then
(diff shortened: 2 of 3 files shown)


martin-schulze-vireso:
This check should also be updated in libexec/bats. And we should make a test for that.

qec-pconner:
Added to description that this fixes https://github.com/bats-core/bats-core/issues/672.

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for your contribution.

PR #693, opened 2023-01-24
test/parallel: add a test for negative jobs number arguments

marc-hb:
Before commit 48b0d2650a96, -j2 and -j -2 both triggered an infinite loop. That commit fixed the infinite loop but -j2 and -j -2 are still parsed the same. In the corresponding code review #657 Martin expressed an intent to change the parser and parse these differently in the future.

In commit 0ec3619426f5 Martin added a "no infinite loop" test for "-j2". However this test will stop testing negative arguments when the parser changes (the test will also have to be adjusted when the parser changes).

Add one additional test with a "-j -3" argument that will test a negative argument forever.

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

### DIFF
--- a/test/parallel.bats
+++ b/test/parallel.bats
@@ -218,3 +218,10 @@ check_parallel_tests() { # <expected maximum parallelity>
   (( SECONDS < 5 ))
   [ "${lines[1]}" = 'Invalid number of jobs: -2' ]
 }
+
+@test "Negative jobs number does not run endlessly" {
+  unset BATS_NO_PARALLELIZE_ACROSS_FILES
+  run bats -j -3 "$FIXTURE_ROOT/../bats/passing.bats"
+  (( SECONDS < 5 ))
+  [ "${lines[1]}" = 'Invalid number of jobs: -3' ]
+}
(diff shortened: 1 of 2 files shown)


test/parallel.bats
```diff
@@ -218,3 +218,10 @@ check_parallel_tests() { # <expected maximum parallelity>
   (( SECONDS < 5 ))
   [ "${lines[1]}" = 'Invalid number of jobs: -2' ]
 }
+
+@test "Negative jobs number does not run endlessly" {
+  unset BATS_NO_PARALLELIZE_ACROSS_FILES
+  run bats -j -3 "$FIXTURE_ROOT/../bats/passing.bats"
```
martin-schulze-vireso, line 224: I think we need a timeout if this really does trigger an infinite loop.

martin-schulze-vireso (changes requested):

test/parallel.bats
martin-schulze-vireso: Nevermind, there is already a timeout at the beginning of the file.

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thank you for your contribution.

PR #710, opened 2023-03-04
Test cat formatter and update usage string

andregri:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 * add `${BATS_TEST_TAGS[@]}` for querying the tags during a test (#705)
 * print tags on failing tests (#705)
+* add tests for cat formatter (#710)
 ### Fixed
 
 * fix `${BATS_TEST_NAMES[@]}` containing only `--tags` instead of test name since Bats v1.8.0 (#705)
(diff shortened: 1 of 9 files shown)


martin-schulze-vireso:
Thank for your suggestion. The cat formatter is intended as an internal debugging tool and as such not documented. Could you motivate why documentation and tests are warranted for this?

andregri:
Since the cat formatter is supported by bats, it should be documented for developers or contributors. Maybe it would be better to improve the PR by adding a bit of documentation to specify that is used as an internal tool.

Even if it is an internal / debugging tool, I think that tests acts like documentation for contributors.

Eventually i added tests for completeness.

martin-schulze-vireso (changes requested):
Following our previous discussion I reviewed the Code and have following suggestions:

1. Documentatiin of this formatter should only be provided in the contributor guide
2. AFAIK the only part of this formatter that is not already tested is that it prints the extended tap stream.

andregri:
> Documentation of this formatter should only be provided in the contributor guide

Should we document cat formatter in [CONTRIBUTING.md](https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md), for instance under "Testing" section ? Or should we add `cat` option in [USAGE.md](https://github.com/bats-core/bats-core/blob/master/docs/source/usage.md)?

> AFAIK the only part of this formatter that is not already tested is that it prints the extended tap stream

 To test the extended tap stream with the cat formatter I used the same input as the test "All formatters (except cat) implement the callback interface" in [formatter.bats](https://github.com/bats-core/bats-core/blob/master/test/formatter.bats#L55).

andregri:
I wrote few lines under the Testing section. Since it is a debugging tool, I think it may be useful to read about the cat formatter during testing.

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for your contribution.

PR #720, opened 2023-04-06
codespell: add config, action and fix some typos with it

yarikoptic:

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,3 @@
+[codespell]
+skip = .git,*.pdf,*.svg
+# ignore-words-list = 
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
 skip = .git,*.pdf,*.svg
-# ignore-words-list = 
+ignore-words-list = dne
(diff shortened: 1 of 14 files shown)


jasonkarns:
Can you provide an explanation of what codespell is and why we'd want it? Currently this PR has no description explaining the problem,  proposed solution or alternatives.

yarikoptic:
From the projects page https://github.com/codespell-project/codespell

Fix common misspellings in text files. It's designed primarily for checking misspelled words in source code, but it can be used with other files as well. It does not check for word membership in a complete dictionary, but instead looks for a set of common misspellings. Therefore it should catch errors like "adn", but it will not catch "adnasdfasdf". This also means it shouldn't generate false-positives when you use a niche term it doesn't know about.

As for efficiency and precision - diff in this PR speaks best of all. I have not investigated alternatives but did contribute to codespell while introducing it to all projects I participate in and dozens of projects I anyhow use our touch (eg as debian developer) in my life

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for your contribution.

PR #738, opened 2023-06-11
List test names 508 701

### LINKED ISSUE #508: Flag for enumerating, but not executing, tests
:wave: First just want to say thank you for the amazing project!

The more I've been using bats in my projects the more I've realized it would be beneficial to be able to wire it up in editors/IDEs to show in their test visualizers/explorers, and I've started experimenting with this type of integration. Something that would make that integration a lot easier would be if bats supported an option/flag (e.g. `-l | --list`) that would list out all the discovered tests without having to execute them, conceptually similar to the `-c | --count` flag.

Is there any existing option that supports this? The only path I was able to find today was to actually execute the full suite and then parse the results.

Assuming the option doesn't exist, is it something you'd be willing to consider? I'd be happy to take a pass at a PR, though I'd likely need some pointers because while a cursory glance of the code makes it seem pretty straightforward there does seem to be some internal encoding/representation of test names that's not terribly user friendly.

### LINKED ISSUE #701: List available test names
Running `bats -rc .` or `bats -c <somedir>` prevents for starting to much tests at once. Great.
After that I filter the test I'd like to run with `-f <pattern>`. But one have to lookup for the 'pattern' first.

`-l, --list      List test cases names without running any tests`
This should be compatible with `--recursive` and `--filter` 

`grep -r '^@test' *bats | grep <pattern>` ... but it's not nice

irkode:

Hi all,

I digged into that ones and hope I have found a valuable implementation providing a list in _bats processing order_ with or without filenames:

Close #508 and close #701

Hope you find it valuable. And in case there's something to add or change please let me now.

### Switches
  * `-l`, `--list`:
    list test names without running any tests
  * `-L`, `--list-with-filename`:
    Same as "-l" but print filenames, too

### Output

```bash
$ bin/bats test/fixtures/list_test_names/cryptic.bats -F tap
1..5
ok 1 nothing_special at-this-one
ok 2 don't fail on 'single' or Double quotes
ok 3 handles braces < [ } ({[]}) )(
ok 4 handles 
ok 5 handles \backslash incl sequences \n \r \t

5 tests, 0 failures

$ bin/bats test/fixtures/list_test_names/cryptic.bats -l
nothing_special at-this-one
don't fail on 'single' or "Double" quotes
handles braces < [ } ({[]}) )(
handles `backticks`
handles \backslash incl sequences \n \r \t

$ bin/bats test/fixtures/list_test_names/cryptic.bats -L
test/fixtures/list_test_names/cryptic.bats : nothing_special at-this-one
test/fixtures/list_test_names/cryptic.bats : don't fail on 'single' or "Double" quotes
test/fixtures/list_test_names/cryptic.bats : handles braces < [ } ({[]}) )(
test/fixtures/list_test_names/cryptic.bats : handles `backticks`
test/fixtures/list_test_names/cryptic.bats : handles \backslash incl sequences \n \r \t
```

### Implementation

looks like the only place where we have the original test name available is in bats-preprocess and so it would be extracted from there. Either by:

- creating a new intermediate file with mapping from encoded name to original name
- passing the original name via bats_test_function (additional argument)
- inject a special comment (# bats_test_name <original name>

all these will lead to maintenance of a separate list, keeping in sync when (tag) filtering tests, processing .....
Additionally original names can contain arbitrary characters which will break simple parsing or passing as arguments (Example: `@test "Don't find in empty array" {` even `@test " some "Quotes" inside"` will be a valid test name)

So I thought about decoding the name in case a list-only is requested and implemented the functionality with it.

It needs **no modification** to  the data structures or operating logic, but just add the new command line options and a section in bats-exec-suite to handle printing the list of names. I placed it where `-c` already is evaluated, that's after _gathering_ and _filtering_ (both tags and text) .

the bats _test suite_ is running fine and I added some tests and updated the manpage.

### Suggestion: 

- maybe the functions `bats_encode_test_name` (in `bats-preprocess`) and `bats_decode_test_name` (new in `bats-exec-suite`) could be externalized to `common.bash` or ...
- maybe a nice addition would be to print the line number when using `-L`
  ```bash
  test/fixtures/list_test_names/cryptic.bats [178]: nothing_special at-this-one
  test/fixtures/list_test_names/cryptic.bats [1356]: don't fail on 'single' or "Double" quotes
  ```

### DIFF
--- a/test/fixtures/list_test_names/decode.bats
+++ b/test/fixtures/list_test_names/decode.bats
@@ -69,6 +69,7 @@ bats_decode_test_name() {
 @test "decode" {
   BATS_TEST_PATTERN="^[[:blank:]]*@test[[:blank:]]+(.*[^[:blank:]])[[:blank:]]+\{(.*)\$"
   local count=0
+  local encoded decoded
   while read -r line; do
     if [[ "$line" =~ $BATS_TEST_PATTERN ]]; then
       (( count += 1 ))
@@ -78,5 +79,5 @@ bats_decode_test_name() {
       bats_decode_test_name "$encoded" 'decoded'
       [[ "$name" == "$decoded" ]]
     fi
-  done < <(find $BATS_ROOT/test/fixtures/list_test_names -type f -name '*.bats' -print0 | xargs -0 grep -h '^@test')
+  done < <(find "$BATS_ROOT"/test/fixtures/list_test_names -type f -name '*.bats' -print0 | xargs -0 grep -h '^@test')
 }
--- a/test/fixtures/list_test_names/decode.bats
+++ b/test/fixtures/list_test_names/decode.bats
@@ -1,0 +1,82 @@
+#!/usr/bin/env bats
+
+bats_require_minimum_version 1.5.0
+
+bats_encode_test_name() {
+  local name="$1"
+  local result='test_'
+  local hex_code
+
+  if [[ ! "$name" =~ [^[:alnum:]\ _-] ]]; then
+    name="${name//_/-5f}"
+    name="${name//-/-2d}"
+    name="${name// /_}"
+    result+="$name"
+  else
+    local length="${#name}"
+    local char i
+
+    for ((i = 0; i < length; i++)); do
+      char="${name:$i:1}"
+      if [[ "$char" == ' ' ]]; then
+        result+='_'
+      elif [[ "$char" =~ [[:alnum:]] ]]; then
+        result+="$char"
+      else
+        printf -v 'hex_code' -- '-%02x' \'"$char"
+        result+="$hex_code"
+      fi
+    done
+  fi
+
+  printf -v "$2" '%s' "$result"
+}
+
+bats_decode_test_name() {
+  local name="$1"
+  local length="${#name}"
+  name="${name#test_}"
+  local i=0
+  local result=''
+  local char=''
+  while (( i < length )); do
+    char="${name:$i:1}"
+    case $char in
+      '-')
+        (( i += 1 ))
+        if [[ "${name:$i:4}" == '2d5f' ]]; then
+          result+="_"
+          (( i += 4 ))
+        else
+          result+=$(printf "%b" "\x${name:$i:2}")
+          (( i += 2 ))
+        fi
+      ;;
+      '_')
+        result+=" "
+        (( i += 1 ))
+        ;;
+      *)
+        result+="$char"
+        (( i += 1 ))
+    ;;
+    esac
+  done
+  printf -v "$2" '%s' "$result"
+}
+
+# bats test_tags=bats:slow
+@test "decode" {
+  BATS_TEST_PATTERN="^[[:blank:]]*@test[[:blank:]]+(.*[^[:blank:]])[[:blank:]]+\{(.*)\$"
+  local count=0
+  while read -r line; do
+    if [[ "$line" =~ $BATS_TEST_PATTERN ]]; then
+      (( count += 1 ))
+      name="${BASH_REMATCH[1]#[\'\"]}"
+      name="${name%[\'\"]}"
+      bats_encode_test_name "$name" 'encoded'
+      bats_decode_test_name "$encoded" 'decoded'
+      [[ "$name" == "$decoded" ]]
+    fi
+  done < <(find $BATS_ROOT/test/fixtures/list_test_names -type f -name '*.bats' -print0 | xargs -0 grep -h '^@test')
+}
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -137,7 +137,7 @@ fi
 arguments=()
 
 # list of single char options that don't expect a value
-single_char_flags="hvcprtTx"
+single_char_flags="hvclLprtTx"
 
 # Unpack single-character options bundled together, e.g. -cr, -pr.
 for arg in "$@"; do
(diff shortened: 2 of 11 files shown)


irkode:
### Question:

If there are no tests to be printed I am not printing anything to stdout
* but still get one empty line if there are no tests.
* and an exit code 1 instead of 0.

root cause:

the output of _bats-exec-suite_ is piped to `validator.bash`. This expects to have some input in the pipe

```bash
bats_test_count_validator() {
  trap '' INT # continue forwarding
  header_pattern='[0-9]+\.\.[0-9]+'
  IFS= read -r header                                     #<- here 
  # repeat the header
  printf "%s\n" "$header"

```

I'm not sure how solve that consistent with the design:
* We could check for empty pipe and just do nothing in that case in the validator. This would change the behaviur to allow any empty output from `bats-exec-suite`. 
  Is it intention, that _bats_ treats empty output from `bats-exec-suite` in the validator as an error?

* to let empty output pass, we could print a dummy line like  `#~BATS_IGNORE_OUPUT~#` and if found by the validator, just do nothing

  ```bash
    IFS= read -r header
    [[ "$header" -eq "#~BATS_IGNORE_OUPUT~#" ]] && return 0
    # repeat the header
    printf "%s\n" "$header"
  ```

* we could skip the validator and formatter part in `bats` (line 516) by adding an extra `if`

  ```bash
  if [[ flag -l or -L is given ]]; then
    exec bats-exec-suite "${flags[@]}" "${filenames[@]}"
  ...
  ```

* ... 

Any suggestions or preferences from your side?

martin-schulze-vireso:
> * We could check for empty pipe and just do nothing in that case in the validator. This would change the behaviur to allow any empty output from `bats-exec-suite`.
>   Is it intention, that _bats_ treats empty output from `bats-exec-suite` in the validator as an error?

There has not been much thought about this. The validator has been moved into this position to allow for modular checking of the tap stream. Usage with non tap input has not been fleshed out except for not failing the `-c` switch.

I find it not necessarily bad that listing no tests results in an error exit code. This prevents errors where the test list is empty by accident.

irkode:
> I find it not necessarily bad that listing no tests results in an error exit code. This prevents errors where the test list is empty by accident.

I would consider a successful call should return 0. Especially in a command chain one relies on the exit code before checking for error details. 

For an empty test list it's ok to have no output. If the process exits with 1 this tells  _something failed_ but the caller has no clue if this is an 'accidential error' or just an empty list of tests.

bats in general allows execution without test and just prints tap header or a pretty message 
If `bats-exec-suite` exits non-zero without an error message it looks like someone found a bug :-)

I rechecked the alternatives (including a flag to the validator to allow empty input)  but for me it looks like the best way is to just skip formatting in case of -l - (saves two useless pipe commands at cost of a [[ test ]]) 

#### bats

```bash
# line 231
  -l | --list)
    flags+=('-l')
    skip_formatter=1 
    ;;
  -L | --list-with-filename)
    flags+=('-L')
    skip_formatter=1
    ;;
... # line 519++
# skip output formatting in case of we onyl print names (-l/-L)
if [[ -n "$skip_formatter" ]]; then
  exec bats-exec-suite "${flags[@]}" "${filenames[@]}"
elif [[ -n "$report_formatter" ]]; then
  exec bats-exec-suite "${flags[@]}" "${filenames[@]}" |
    bats_tee "${BATS_REPORT_OUTPUT_DIR}/${BATS_REPORT_FILENAME}" "$interpolated_report_formatter" "${report_formatter_flags[@]}" |
....

File: test/fixtures/list_test_names/decode.bats
Comment (martin-schulze-vireso) [inline test/fixtures/list_test_names/decode.bats:1]: Why is this test in a fixtures folder?

File: test/fixtures/list_test_names/decode.bats [hunk 1]
```diff
@@ -0,0 +1,83 @@
...
+        (( i += 1 ))
+        ;;
+      *)
+        result+="$char"
+        (( i += 1 ))
+    ;;
+    esac
+  done
+  printf -v "$2" '%s' "$result"
+}
```
martin-schulze-vireso, line 66: This seems like an awful amount of code duplication to me. Is this really necessary? Can't we just hardcode the expected test results below?

test/fixtures/list_test_names/decode.bats
irkode: mmh, not for any special reason, just reused the structure with having subtests for `list_test_names` in a common folder and the fixture layout seems to match . The defined "test_names" of this subfolder also make up the test names for the `decode_bats_test_name`  function.

If fixture is reserved for bug fixes only we could think about another subfolder structure.

currently the functions are defined (runtime) in `bats-preprocess` and `bats-exec-suite` the definition is a duplicate in the test-suite as it is impossible to source these files. That's why I suggested to exeternalize both functions to a _sourceable_ library file to be included at runtime and test.

see also my comment on the duplication below.

test/fixtures/list_test_names/decode.bats
irkode: First duplication of both functions is due to the fact, that the files where they are defined at bats runtime is not sourceable. This in fact should be eliminated by externalizing the functions to a core library.

If you think about duplication between encode and decode, I would not consider that as a duplicate encode and decode are inverse functions.

The tests written are _black box_ tests not caring about the actual encoded result but just checking that after encode and decode the original name  is returned by decode.

I did not add a test for the _encode_ function itself becuse the function was already there - might be nice to have some there, too
Imho the encode could be tested by defining a hard coded name -> encoded_name list. For the decode I would consider the name ->  encode =>  decode -> name the better approach because it ensures the result is the same.

If we use hardcoded names, a breaking change to the encode function one woul dhave to update *both* not only the tests for encode, but also for decode (at least the list). If not the decode tests would still pass.
using the encode -> decode method will make the decode method fail on braking changes.

PR #873, opened 2024-02-21
Check Bash version

akinomyoga:
I here suggest including the check for the Bash version at the beginning of `bats`.

Using Bats, I would like to test if Bash frameworks would correctly work in different versions of Bash. I tried to run the test for the devel version of Bash, the latest release 5.2, down to Bash 3.0. However, the test fails because of syntax errors for Bash 3.0 and 3.1. It turned out that the failure is not caused by the Bash frameworks that are tested but by the `bats` test framework itself. I'd like to request Bats to check the Bash version first and output an error message if the minimum Bash version requirement by Bats is not satisfied.

I tried to find an explicit notice about the required Bash version, but I couldn't find any information. What is the minimum Bash version supported by Bats? I here assume it is 3.2 because Bats produce syntax errors in Bash 3.0 and 3.1, while some files seem to contain checks for 3.2 (e.g. in [`test/trace.bats`](https://github.com/bats-core/bats-core/blob/18eca2496cbd80ac97c73a70002e7ab4fb2179cf/test/trace.bats#L72)).

  - The guideline says `[[ ... ]]` should be always used, but can I use `[ ... ]` for the very initial test for the shell? Relying on the Bash feature for the test of whether the Bash feature is available doesn't seem to make sense.
  - Also, the guideline says "*Always use [[ and ]] for evaluating variables*", but what "*evaluating variables*" means is unclear. I assume it considers just the two variations of `[[ ... ]]` and `[ ... ]`. However, technically, it might also exclude the uses of the arithmetic commands of the form `(( ... ))`. I used the arithmetic command in this PR because arithmetic commands seem to be already used in the codebase.
  - Unrelated to this PR, but the "Current Maintainers" section seems outdated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,9 @@
 
 [![Join the chat in bats-core/bats-core on gitter](https://badges.gitter.im/bats-core/bats-core.svg)][gitter]
 
-Bats is a [TAP](https://testanything.org/)-compliant testing framework for Bash.  It provides a simple
-way to verify that the UNIX programs you write behave as expected.
+Bats is a [TAP](https://testanything.org/)-compliant testing framework for Bash
+3.2 or above.  It provides a simple way to verify that the UNIX programs you
+write behave as expected.
 
 A Bats test file is a Bash script with special syntax for defining test cases.
 Under the hood, each test case is just a function with a description.
--- a/bin/bats
+++ b/bin/bats
@@ -2,6 +2,19 @@
 
 set -euo pipefail
 
+# Note: We first need to use POSIX's `[ ... ]' instead of Bash's `[[ ... ]]'
+# because this is the check for Bash, where the shell may not be Bash.  Once we
+# confirm that we are in Bash, we can use [[ ... ]] and (( ... )).  Note that
+# these [[ ... ]] and (( ... )) do not cause syntax errors in POSIX shells,
+# though they can be parsed differently.
+if [ -z "${BASH_VERSION-}" ] ||
+    [[ -z "${BASH_VERSINFO-}" ]] ||
+    ((BASH_VERSINFO[0] < 3 || (BASH_VERSINFO[0] == 3 && BASH_VERSINFO[1] < 2)))
+then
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -10,6 +10,7 @@
 
 Bats is a [TAP](https://testanything.org/)-compliant testing framework for Bash.  It provides a simple
 way to verify that the UNIX programs you write behave as expected.
+Bats itself is also written in Bash and requires Bash 3.2 or above.
```
martin-schulze-vireso, line 13: Maybe we should just bot "3.2 or above" into the first sentence on line 11? Anyways, did you see https://bats-core.readthedocs.io/en/stable/support-matrix.html?

bin/bats
```diff
@@ -2,6 +2,20 @@
 
 set -euo pipefail
 
+# Note: We need to use POSIX's `[ ... ]' instead of Bash's `[[ ... ]]' because
+# this is the check for Bash, where the shell may not be Bash.
+if [ -z "${BASH_VERSION-}" ]; then
```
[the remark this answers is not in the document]
akinomyoga: OK, it seems to work. I thought the latter `if` statement may cause a syntax error in shells that do not support Bash syntaxes, but after thinking about it again, it seems fine even though it's syntactically interpreted differently.

README.md
akinomyoga: > Maybe we should just bot "3.2 or above" into the first sentence on line 11?

I'll update it in that way.

> did you see https://bats-core.readthedocs.io/en/stable/support-matrix.html?

I'm sorry. I tried to find any version information about Bash using `git grep version`, but I missed it. (I again tried to look it up in the result of `git grep version`, but I couldn't find it until I rerun it as `git grep 'Bash version'`...)

martin-schulze-vireso:
Thanks for your contribution. I slightly reworded the changelog entry and signed the commit.

akinomyoga:
Thank you!

PR #882, opened 2024-03-07
Add variable to track bats call arguments

soda480:

Having such a variable is useful for use cases where downstream functions can call bats with the original arguments, for example I am defining a setup-suite to automatically generate a coverage report and badge for bats, I can use this variable like:
```
coverage() {
    if which kcov; then
        echo "# generating coverage report using kcov" >&3
        # strip --setup-suite-file argument to prevent infinite loop condition
        kcov --include-path=$WORK_DIR $WORK_DIR/coverage bats ${BATS_ARGS%--setup-suite-file*}
        coverage_badge
    else
        echo "# unable to generate coverage report because kcov is not installed" >&3
    fi
}
```

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -112,6 +112,7 @@ BATS_LIBEXEC="$(
   pwd
 )"
 export BATS_LIBEXEC
+export BATS_ARGS="$@"
 export BATS_CWD="$PWD"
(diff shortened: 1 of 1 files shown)


martin-schulze-vireso:
I think we have to take more care with separators. Maybe `printf -v BATS_ARGS '%q ' "$@"`?

PR #934, opened 2024-06-05
Optimization function method

ut003460:
1. Merge the local declaration with variable initialization to make the code more concise.
2. Use ${# array [@]} directly to obtain the length of the array, without using eval.
3. Declare array variables outside the loop instead of declaring them every time in the loop

 [✔ ] I have reviewed the [Contributor Guidelines][contributor].
 [✔] I have reviewed the [Code of Conduct][coc] and agree to abide by it

### DIFF
--- a/lib/bats-core/common.bash
+++ b/lib/bats-core/common.bash
@@ -197,22 +197,23 @@ bats_trim() {                                            # <output-variable> <st
 # a helper function to work around unbound variable errors with ${arr[@]} on Bash 3
 bats_append_arrays_as_args() { # <array...> -- <command ...>
   local -a trailing_args=()
-  while (($# > 0)) && [[ $1 != -- ]]; do
-    local array=$1
+  local array
+  while (( $# > 0 )) && [[ $1 != -- ]]; do
+    array=$1
     shift
 
-    if eval "(( \${#${array}[@]} > 0 ))"; then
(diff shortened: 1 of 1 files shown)


lib/bats-core/common.bash
```diff
@@ -197,22 +197,23 @@ bats_trim() {                                            # <output-variable> <st
...
-  while (($# > 0)) && [[ $1 != -- ]]; do
-    local array=$1
+  local array
+  while (( $# > 0 )) && [[ $1 != -- ]]; do
+    array=$1
     shift
 
-    if eval "(( \${#${array}[@]} > 0 ))"; then
-      eval "trailing_args+=(\"\${${array}[@]}\")"
+    if (( ${#array[@]} > 0 )); then
```
martin-schulze-vireso, line 205: This totally changes the meaning. I think the documentation of this function could be improved but the tests and usage should make it clear:

```
array_1=(a b c)
array_2=(1 2)
bats_append_arrays_as_args array_1 array_2 -- echo
````
will ultimately call
```
echo a b c 1 2
```
so it is a bit like `xargs` but instead of consuming everything from stdin, the arguments before `--` are arrays whose elements should be appended as parameters to the command.

martin-schulze-vireso:
What exactly are you trying to achieve with this PR? The title sounds like you are trying to improve the performance. 

If so, did you measure this part of the code to be a performance bottleneck?

PR #979, opened 2024-08-22
Recursively call setup_suites in subdirectories

gwangmu:
- [O] I have reviewed the [Contributor Guidelines][contributor].
- [O] I have reviewed the [Code of Conduct][coc] and agree to abide by it

Thank you for maintaining BATS. :)

I modified `bats-exec-suite` so as to execute {setup,teardown}_suite on every subdirectory on a recursive run. The basic idea is to replace the file execution loop (calling `bats-exec-file`) with a tail recursion and to enter a subshell whenever the next test is in the subdirectory. In the subshell, it calls `setup_suite` of that subdirectory (if it exists), executes the test, and calls `teardown_suite` before changing the directory.

The motivation was that, if somebody (like, me) tries to organize their tests using the directory structure, the tests in the subdirectory probably need to assume the test environment of their parent, PLUS their own environment if needed. This is doable if BATS recursively recognizes {setup,teardown}_suites in the subdirectories.

That said, there are some concerns even from my side. I think these need to be addressed before this PR is eventually merged into the master.

- Overhead

This is the most obvious concern. I totally agree with BATS's basic policy of being lightweight. I ensured this modification causes no additional overhead when it's not recursive, but I don't know whether a tail recursive is sizably heavier than a loop. For reference, the `time` difference between the unmodified and modified ones was negligible using the test directory structure in `test/fixture/suite/recursive_setup`.

- The term 'suite' in `setup_suite`

I think this can be a little misnomer when this recursive setup takes effect. A 'suite' arguably means the set of tests, namely all the tests that BATS will execute. From this perspective, 'setup_suite' should be called only once before the entire test begins. However, I need to clarify that the motivation signified something "between" setup_file and setup_suite at a directory level. Maybe there can be several solutions: sticking to `setup_suite` for backward compatibility, newly creating the concept of `setup_dir`, or aggressively changing the name `setup_suite` to `setup_dir`.

- Parallel execution

My implementation doesn't support parallel execution. There's nothing impossible to have this "feature" in parallel execution, but it'll definitely need some engineering effort.

I also need to make it clear that I ran all tests in `test` (i.e., `test.bats` and `suite.bats`) and confirmed that they all passed. I also added a new test for this PR in the `test` directory.

Thank you. :)


martin-schulze-vireso:
First of all let me say that I am sorry for taking so long to respond.

I already thought about this problem before you posted this PR and while its simplicity has a lot going for it I feel that we need a more thorough approach.

My design goal with the setup/teardown functions was to keep their changes local to their execution context. This means a `setup_file` should not be able to change (env) variables such that another file might see the variables defined there, to avoid "cross-talk". This is achieved by having each setup function run in their own subshell/process and only inheriting from the "parent contexts" (setup_file, setup_suite).

Transferring this to the domain of arbitrarily nested, folder based `setup` functions would require to mimic the folder structure in the process/subshell hierarchy. This has ramifications for the code structure and probably complicates the parallelization code.

In many cases this setup/teardown firewall (of sorts) might be overkill and I toyed with the idea of making it optional for performance reasons (the many subprocesses can slow you down on some platforms). However, I have not found the time to think about this in a structured manner so far, let alone spell it out or try some implementation.

PR #985, opened 2024-09-04
fix: use $BATS_LIBDIR instead of hardcoded lib in bats-gather-tests

### LINKED ISSUE #982: libexec/bats-core/bats-gather-tests should use $BATS_LIBDIR not /lib
**Describe the bug**
bats-gather-tests has a hard coded reference to lib and does not use $BATS_LIB when sourcing common.bash and others.

```
$ bats .
/usr/libexec/bats-core/bats-gather-tests: line 10: /usr/lib/bats-core/common.bash: No such file or directory
```

**To Reproduce**

1. Install bats-core to lib64
2. run `bats .` in a directory containing multiple bats files

**Expected behavior**
bats should run all tests

 - Bats version [e.g. 1.4.0 or commit hash if applicable]:  master
 - operating system (including version): [e.g. Linux (please name your distribution!), FreeBSD, MacOS]: CentOS 7.9
 - `bash --version`: [e.g. 5.1] 4.2.46
 - Install method: [e.g. git submodule, distribution package manager, npm, homebrew, ...] bash install.sh /usr lib64

Cheers,
  Peter

pscn:
- [ I'm an idiot] I have reviewed the [Contributor Guidelines][contributor].
- [PR without testing...  PFFF... ] I have reviewed the [Code of Conduct][coc] and agree to abide by it

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -197,6 +197,21 @@ jobs:
           docker run -it "bash:${{ matrix.version }}" --version
           time docker run -it ${{ matrix.env_vars }} "bats/bats:bash-${{ matrix.version }}"  --print-output-on-failure --tap /opt/bats/test
 
+  lib64-install:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Install in lib64
+        run: sudo ./install.sh /usr/ lib64
+      - name: Run from env
+        run: time bats test --print-output-on-failure --report-formatter junit --output test-results
+      - name: Publish Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v2
+        if: always()
+        with:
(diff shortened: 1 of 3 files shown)


brokenpip3:
I started the tests, can you drop a line in the changelog? check the already merged pr to check how to do it.
Thanks!

martin-schulze-vireso:
I wonder why the tests did not catch this? IIRC there is a test with different lib path.

martin-schulze-vireso:
I cleaned up the history and added another test that should reproduce the issue. Thanks for your contribution!

martin-schulze-vireso (approved):

brokenpip3:
> I wonder why the tests did not catch this? IIRC there is a test with different lib path.

maybe in our github action we always run bats from the repo and when we use the nodejs install it will always it in the same path? we may want to add a different path matrix by testing the install.sh installation method

martin-schulze-vireso:
I added the Tests here and fixed them in #990

PR #1017, opened 2024-11-19
Upgrade detik to 1.3.2

### LINKED ISSUE #1018: bats/bats docker image is using older version of detik
**Describe the bug**
The Dockerfile references 1.3.1.  This version has a [bug](https://github.com/bats-core/bats-detik/issues/46) that was fixed in 1.3.2.

**To Reproduce**

1. use `docker run -it -v "${PWD}:/code" bats/bats:latest ...` to run a test that uses detik and a regular expression in the name of an object (`at most 1 times every 30s to get kustomization named 'infrastructure-stage[[:digit:]]+' and verify that 'status.lastAppliedRevision' is '$LAST_APPLIED_REVISION'`).
2. test won't find the object.

**Expected behavior**
Latest version of detik will be included in `:latest` docker image of bats-core.

 - operating system (including version): docker image
 - `bash --version`: 1.11.0 (bats/bats:latest)
 - Install method: docker image.

youngderekm:
Upgrade reference to detik in Dockerfile from 1.3.1 to 1.3.2 (released July 17). closes #1018


brokenpip3:
can you please add a line in the changelog? like [here](https://github.com/bats-core/bats-core/pull/932)

youngderekm:
I updated the CHANGELOG.md in the unreleased section.  It already listed an upgrade to detik 1.3.1 so I replaced that line with an upgrade to 1.3.2.
However, from what I can tell, the docker image (bats/bats) is using bats-core 1.11.0 but detik 1.3.1 (which must be the unreleased version?).

youngderekm:
> > However, from what I can tell, the docker image (bats/bats) is using bats-core 1.11.0 but detik 1.3.1 (which must be the unreleased version?).
> 
> what do you mean by this? bats is the binary itself and detik is one of the libraries (plugins or ext if we want to call it in another way) that have different version and the 1.3.2 is the latest one of detik while 1.11 is the latest of bats. Maybe I'm misunderstood your question :)

What I mean is that bats-core is the source of `bats` (the binary) and also the docker image `bats/bats:latest` (the Dockerfile).  I was wondering if `bats/bats:latest` was built/pushed from the master branch, rather than the v1.11.0 tag, and built after v1.11.0 was released.  It looks to me like the md5sum of detik.bash that is in the container matches the v1.3.1 version from bats-detik, even though bats-detik was upgraded to v1.3.1 after the 1.11.0 release of bats-core (Dockerfile from v1.11.0 specifies detik v1.3.0).  Maybe I'm mistaken though.  Either way, the Dockerfile should now have the latest detik.

martin-schulze-vireso (approved):

youngderekm:
Thanks for looking into this.  I see what happened on my end now - detik.bash didn't change between v1.3.0 and v1.3.1, so the sha1sum from the docker image correctly matched by tags.  Looking at the other files like you did would have cleared it up for me.  All good now.

PR #1031, opened 2024-11-30
feat: add bats::on_failure hook

### LINKED ISSUE #756: Add on_failure callback
Sometimes we want to react to test failures in a specific way, doing special cleanup, inspecting in an interactive shell, ... currently, we can only use `teardown` and check whether the test completed.

An `on_failure` callback function that gets called immediately, when the test is known to have failed. As a bonus this might allow for overriding the failure.

Special "failure only" cleanup can be done in `teardown` but is a bit more cumbersome. Dropping into an interactive shell could be solved with a specialized flag.

martin-schulze-vireso:
Contrary to setup/teardown this has the bats:: prefix to avoid name collisions due to t he generic name.

In the long run, the other functions should be named bats::setup/bats::teardown/bats::... as well.

TODOs:
- [x] documentation

Closes #756

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ## [Unreleased]
 
+### Added 
+
+* `bats::on_failure` hook that gets called when a test or `setup*` function fails (#1031)
+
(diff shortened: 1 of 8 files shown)


martin-schulze-vireso:
@brokenpip3 do you want to review?

brokenpip3 (approved):

Issue #976, opened 2024-08-19
Dynamic registration overwrites files and fails with noclobber


someone-stole-my-name:
**Describe the bug**

After https://github.com/bats-core/bats-core/commit/9d5ecdbe319bd0e39ef2637806832bb07768fb23 when calling `bats` with a list of files and a shell with `noclobber` it fails because it seems that bats now runs the [preprocess](https://github.com/bats-core/bats-core/blob/master/lib/bats-core/preprocessing.bash#L8) step N times instead of a single run like it used to.

**To Reproduce**

1. Create two `.bats` files with dummy tests
2. Create a wrapper script to run them:
```sh
set -C
export SHELLOPTS
find . \
        -type f \
        -name '*.bats' \
        -print0 \
        | sort -z \
        | xargs -0 -r bats --tap
```
3. Run the script and output should be similar to:

```
/usr/lib/bats-core/preprocessing.bash: line 13: /tmp/bats-run-jf8poy/bats.64.src: cannot overwrite existing file
1..1
not ok 1 bats-gather-tests
...
```

**Expected behavior**

Running the tests

**Environment (please complete the following information):**
 - Bats version: 1.11.0
 - operating system (including version): Alpine 3.20.2
 - `bash --version`: 5.2.26(1)-release (aarch64-alpine-linux-musl)
 - Install method: package manager

martin-schulze-vireso:
Fixed in #1042

PR #1042, opened 2024-12-18
fix gather-tests breaking under noclobber

### LINKED ISSUE #1039: Evaluating each test file in its own shell overwrites files and fails with noclobber
**Describe the bug**

After release [v1.11.1](https://github.com/bats-core/bats-core/releases/tag/v1.11.1), at [commit](https://github.com/bats-core/bats-core/commit/5c52dc304f13593f48ccf97a070a0e57032c97fd), when calling bats with a list of files and a shell with noclobber (Example provided below) it fails because it seems that bats runs the [bats-test-gatehering](https://github.com/bats-core/bats-core/blob/5c52dc304f13593f48ccf97a070a0e57032c97fd/libexec/bats-core/bats-gather-tests#L292) step N times instead of a single run like it used to. The failure happens as it tries to overwrite the output file (`$BATS_OUT`) each time.

A similar problem with other file was reported in [issue #976](https://github.com/bats-core/bats-core/issues/976) for the previous version.

**To Reproduce**

1. Create two `.bats` files with dummy tests

2. Create a wrapper script to run them:

   ```bash
    set -C
    export SHELLOPTS
    find . \
        -type f \
        -name '*.bats' \
        -print0 \
        | sort -z \
        | xargs -0 -r bats --tap
    ```

3. Run the script and output should be similar to:

   ```bash
    /usr/libexec/bats-core/bats-gather-tests: line 301: /tmp/bats-run-ymad5G/gather-tests.out: cannot overwrite existing file
    1..1
    not ok 1 bats-gather-tests
    ...
    ```

**Expected behavior**

It should be running the test without erroring.

 - Bats version: v1.11.1
 - operating system (including version): Alpine Linux v3.21.0
 - `bash --version`: 5.2.37(1)-release (aarch64-alpine-linux-musl)
 - Install method: package manager

**Additional context**

Issue may be solved by allowing to overwrite the file with the clause `>|` when redirecting the output.

martin-schulze-vireso:
improve naming and structure of BATS_OUT/src files in BATS_RUN_TMPDIR

Fixes #1039 and #976

### DIFF
--- a/.github/workflows/set_nounset.bash
+++ b/.github/workflows/set_nounset.bash
@@ -1,1 +1,1 @@
-set -u
+set -uo noclobber
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 ### Fixed
 
 * increase compatibility with Solaris (#1043)
+* fix `noclobber` breaking `bats-gather-tests` (#1042)
 
 ## [1.11.1] - 2024-11-29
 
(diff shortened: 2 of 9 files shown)


brokenpip3 (approved):
:rocket:

martin-schulze-vireso:
Thanks for the review @brokenpip3

Issue #209, opened 2019-04-03
Add an option to fail the entire run as soon as first test fails


filmil:
Sometimes I just want to get an error report if the test is very broken, as soon as possible, instead of waiting for the entire test suite to finish running.

abbbi:
hi,

that would be a nice feature. My testsuite basically depends on wether if the first test finishes succesfully. If this is not the case then other tests may fail too. Of course one could add the first test in the first bats file and call the other bats files only if the first test works out.. but well.

One way to workaround that missing option would be to kill $PPID (bats) if one test condition
fails. Then the entire run exits after the failed function. Its not very clean but seems to work.
Example:

```
checkstate() {
    if [ $1 == 0 ]; then
        return 0
    else
        kill $PPID
        return 1
    fi
}
@test "foo_fails" {
    run cat /etc/nonexistant
    checkstate $status
}
@test "bar" {
    run cat /etc/fstab
    checkstate $status
}
@test "baz" {
    run cat /etc/fstab
}

```

sublimino:
There is another type of workaround on the original `bats` repo: https://github.com/sstephenson/bats/issues/135#issuecomment-242366074

A pull request for a `--fail-fast` flag based on the above approach would be welcome if anybody wants to add it.

carlosonunez-wiz:
cc: @carlosonunez

[This](https://github.com/sstephenson/bats/issues/135#issuecomment-242366074) workaround still works, but only if you have all of your tests in a single file.

Here's a modified version of this workaround that should work across test files. I'm assuming that your tests are in a directory called `tests` at the root of your project:

```sh
setup() {
  [ ! -f ${BATS_RUN_TMPDIR}/.skip ] || skip "skip remaining tests" ]
}
teardown() {
  [ -n "$BATS_TEST_COMPLETED" ] || touch ${BATS_RUN_TMPDIR}/.skip
}
```

Also, if you run BATS within Docker Compose, you can use the modification shown below to automatically inject the statements above into every `.bats` test file (so that you don't have to repeat yourself any time you create a new test):

```yaml
services:
  bats:
    depends_on:
      add-fail-fast-blocks-to-tests: # This mutates your test files before BATS starts whenever you run 'docker compose run --rm bats' or similar.
        condition: service_completed_successfully
    images: bats-core/bats
    working_dir: /tests
    volumes:
      - modified-tests:/tests
    command:
      - .
  add-fail-fast-blocks-to-tests:
    image: bash:5
[...]

PR #1101, opened 2025-06-06
Add section about exiting on test failure to FAQ

kravciak:
Thanks to Carlos Nunez (carlosonunez-wiz) for [initial idea](https://github.com/bats-core/bats-core/issues/209#issuecomment-2444878157)


martin-schulze-vireso:
Thank you for work. While this seems to be a clever workaround, I would prefer a less "hacky" solution in the official documentation.

Seeing that it might take a while until that is in place we need a better place. We could add a community wiki page for these kinds of workarounds or put that into the gotchas section with a link to the appropriate ticket requesting/tracking the actual solution.

Issue #1108, opened 2025-06-17
bats add --skip cli option


Luap99:
**Is your feature request related to a problem? Please describe.**

Consider the use case where upstream tests are run downstream in many different environments. Sometimes one more tests might fail in that downstream environment, for example a new dependency update broke something there. The downstream run just reuses the test files and as such it would be much nicer if we could skip a test via a cli option instead of requiring us to patch the test files everytime that happens. In our case we are talking about (hopefully) short lived issues that will be fixed quickly again so adding a cli option to the bats run in the testing env seems easier than patching 1+ file contents.

**Describe the solution you'd like**

Add a --skip option to bats that acts basically inverse to what --filter is doing.

**Describe alternatives you've considered**

Alternative for us to hack up some preprocessor with sed that looks for the test name and then inserts a `skip` function call in the test block.

**Additional context**

We are running podman bats tests in fedora openQA infrastructure and that just reuses the same test files but testing that against each new fedora update. It is likely that some test breaks there due new dependencies and in that case we like to have an easy way to skip a failing test so we don't get failure reports on every new update that just shows the same already known issue.

https://pagure.io/fedora-qa/os-autoinst-distri-fedora/pull-request/389#comment-220261

PR #1114, opened 2025-07-16
Add a --negative-filter option (#1108)

AdamWill:
See https://pagure.io/fedora-qa/os-autoinst-distri-fedora/pull-request/389 . The FAQ claimed "So you could try to invert the regex", but as far as I can tell, there's no practical way to do this with bash regexes, if you want to express "run all tests except foo_in_a" or something like that. You can do it in some regex flavors using negative lookahead matches, but bash's regex flavor does not seem to support that, and you can't just stick a `!` on the front of the regex or anything like that.

So, this adds --negative-filter which just does the opposite of --filter. The implementation is intentionally 'dumb' because handling 'what if one or both is not defined?' gets a bit awkward if you try to handle both in one function.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -18,8 +18,7 @@ If you want to see this output, you need to print it yourself, or use functions 
 Can I use `--filter` to exclude files/tests?
 --------------------------------------------
 
-No, not directly. `--filter` uses a regex to match against test names. So you could try to invert the regex.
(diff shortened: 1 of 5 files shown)


martin-schulze-vireso:
> Are you planning a release soon, btw?

The new Version was just released @AdamWill . Sorry about the delay and late reply. I could not tell how long it would take and did not want to commit to any schedule.

AdamWill:
awesome news, thanks!

PR #1118, opened 2025-07-27
Add --errexit flag to enable errexit in run commands

### LINKED ISSUE #719: Allow for `set -e` in `run`
**Describe the bug**
The `run` changed resets the `set -e` that change the behaviors of the function.

**To Reproduce**
1. Here is an example script that should fail without print anything to console:
``` shell
set -e

function main() {
    my_command
}

function my_command() {
    false # failing command
    echo 'FOOBAR'
}

if [ "${BASH_SOURCE[0]}" == "$0" ]; then
    main
fi
```
2. But this test will fail, because the `set -e` is not set as the source :
``` shell
setup() {
    source my_script.sh
}

@test "test my_command" {
    run my_command
    [ "$output" != '' ]
}
```

**Expected behavior**
The sourced `set` flags have to be set under testing too.

 - Bats version: 1.8.2
 - OS: Linux (nixos)
 - Bash version: 5.1.16

**Additional context**
I use `run` to have `$status` and `$output`.

I have read the https://bats-core.readthedocs.io/en/stable/writing-tests.html#when-not-to-use-run part of the documentation, that mentions the code will run with `set -e` in a subshell. This would be right for me, but my function works differently.

### LINKED ISSUE #748: errexit is ignored in functions executed via run
**Describe the bug**

Not sure if this is a bug, but it was a surprise to me, and not clear from reading the docs. If this is by-design, then maybe it should be listed under "Gotchas":

When you call a function via `run` then `set -o errexit` inside the function has no effect, and the function is executed to the last statement.

**To Reproduce**

1. Create example `file.bats` with following contents

```bash
load ./bats-support/load.bash
load ./bats-assert/load.bash

foo() {
    set -o errexit
    false
    true
}

@test 'without run' {
    foo
}

@test 'with run' {
    run foo
    assert_success
}
```

2. run `bats file.bats`

```
file.bats
 ✗ without run
   (from function `foo' in file file.bats, line 6,
    in test file file.bats, line 11)
     `foo' failed
 ✓ with run

2 tests, 1 failure
```

**Expected behavior**

I expected that the `false` command in `foo` would terminate the function and set `$status` to `1` when invoked via `run`. So I expected `with run` to fail as well.

 - Bats version [e.g. 1.4.0 or commit hash if applicable]

    ```
    commit f4552f12e8c1800bd2285c6aaec5e5e91d2afe76 (HEAD -> master, origin/master, origin/HEAD)
    Merge: 26e7408 596315d
    Author: Martin Schulze <37703201+martin-schulze-vireso@users.noreply.github.com>
    Date:   Tue Jun 13 01:20:56 2023 +0200

        Merge pull request #736 from martin-schulze-vireso/fix/mocking_date_breaks_timing

        Fix overriding date breaks timing
    ```

 - operating system (including version): [e.g. Linux (please name your distribution!), FreeBSD, MacOS]

    ```
    $ sw_vers
    ProductName:	macOS
    ProductVersion:	12.6.7
    BuildVersion:	21G651
    ```

 - `bash --version`: [e.g. 5.1]

    GNU bash, version 5.2.15(1)-release (x86_64-apple-darwin21.6.0)

 - Install method: [e.g. git submodule, distribution package manager, npm, homebrew, ...]

    git clone of all repos, just for this test

jandubois:
* Set BATS_RUN_ERREXIT environment variable when flag is provided.
* Write `$?` to a temp file when errexit is enabled.

I've tried for several hours, but have been unable to find a solution without using a temp file. I don't quite understand why, but I couldn't make it work.

Fixes #719
Fixes #748

---

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,9 +11,13 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+* `--errexit` flag to enable errexit (set -e) behavior for commands run in `run` (#1118)
+
 ### Fixed
 
-* junit formatter: 
+* junit formatter:
   * avoid interference between env and internals (#1175)
   * remove control characters (\x00-\x08\x0B\x0C\x0E-\x1F) (#1176)
   * don't report (skipped) last test as failed when `teardown_suite` generates FD3 output (#1181s)
@@ -29,7 +33,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ## [1.13.0] - 2025-11-07
 
-### Added 
+### Added
 
 * use the [`syntax`](https://docs.docker.com/reference/dockerfile/#syntax) parser directive to declare the Dockerfile syntax version (#1127)
 * Negative test filtering via `--negative-filter` - tests matching the filter are *excluded* (#1114)
@@ -37,7 +41,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ### Fixed
 
-* unset `output`, `stderr`, `lines`, `stderr_lines` at the start of `run` to avoid crosstalk 
+* unset `output`, `stderr`, `lines`, `stderr_lines` at the start of `run` to avoid crosstalk
   between successive invocations (#1105)
 * junit:
   * XML escape fully removes ANSI sequences, e.g. color codes, cursor movements (#1103)
@@ -56,7 +60,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ## [1.12.0] - 2025-05-18
 
-### Added 
+### Added
 
 * `bats::on_failure` hook that gets called when a test or `setup*` function fails (#1031)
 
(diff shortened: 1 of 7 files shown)


brokenpip3:
> I believe I have addressed all failures found by CI related to the PR.
> 
>     * The Windows tests are cancelled because the windows-2019 runner has been retired.
> 
>     * The Dockerfile test failure seems unrelated to the changes in this PR and fails on other PRs too.

yes indeed, will take a look

martin-schulze-vireso:
> Is there anything else I can do to help get this merged?

I don't think so. I just need to find enough time to review this en bloc, as I already had to abort twice.

martin-schulze-vireso:
I think I finally found a clean way to deal with the tracing problems due to the restructuring triggering the ERR trap inside `run`. This even fixed some edge cases for the sigint trap with run.

The restructuring also should not suppress `set -e` within functions under `run` anymore. This is a behavior change and we might want to make this optin. Then again it might be regarded as a fix?

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for contributing.

jandubois:
> Thanks for contributing.

🙏 Thank you for all your work improving the PR!

PR #1160, opened 2025-10-15
fix: timeout not working with run

martin-schulze-vireso:
TODOS:
- [x] test timeout kills all descendant processes 
- [x] CTRL+C still works
- [x] CTRL+C and timeout works in parallel mode too

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -110,7 +110,7 @@ jobs:
         run: bash -c "! bin/bats test/fixtures/bats/failing.bats"
       - run: |
           bash --version
-          bash -c "time bin/bats  --print-output-on-failure --formatter tap test"
+          bash -c "time bin/bats -T --print-output-on-failure --formatter tap test"
 
   npm_on_windows:
     strategy:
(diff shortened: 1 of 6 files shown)


martin-schulze-vireso:
This is getting more complex than I thought. The main problem of the linked issue is: using `run` adds another subprocess layer, which is shielded from the `kill child processes` command in the timeout watchdog. 

I tried to circumvent that by enabling job control before starting `exec-test` to force all processes of one test into the same process group and sending the signal to the whole group. However, that breaks CTRL+C because the test process group is now separated from the `bats`/`exec-suite`/`exec-file` process group.

Then I tried forwarding the signal from `exec-file` to `exec-test` via a `trap SIGINT` in `exec-file`. However, in the parallel case there might be multiple `exec-test` processes running at the same time, which means I need to either send to all child processes' groups or have each of them listen to the signal individually.

I already had problems with putting the `exec-test` call in the background for the sequential execution to make it easier to send signals to the process.

The old timeout code used SIGABRT which leads to coredumps in some processes. Using SIGINT might be more fitting but makes it harder to distinguish between CTRL+C vs timeout.

martin-schulze-vireso:
Made more progress but for some reason the skip_in_setup_file test is blocked by the background process that should forward sigints to the test subprocesses.

martin-schulze-vireso:
I tried to use SIGINT+TERM as signals but that did not work, so back to SIGABRT.

PR #1172, opened 2025-11-12
Removed duplicate `test` suite runs

fearphage:
Currently every pull request runs an identical test suite _twice_. Once for the `pull_request` event and once for the `push` event. This change reduces it only one once for PRs that will be merged into the default branch. And the `push` runs are limited to the default branch as well. 

> [!NOTE]
> 
> I saw some repositories in this org are named with the new default branch name (`main`) and this one is still using the legacy name (`master`) so I added both so it would continue to work if this default branch were renamed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,7 +1,16 @@
 name: Tests
 
 # Controls when the action will run.
-on: [push, pull_request, workflow_dispatch]
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks, that has been bugging me for a while but I always got distracted with other issues.

PR #1183, opened 2025-12-31
Fix --gather-test-outputs-in fails when tests change directory

### LINKED ISSUE #1132: gather-test-outputs-in fails if test cd's somewhere
**Describe the bug**
If one uses bats --gather-test-outputs-in out ... and the test case cd's somewhere, it fails to create the output file with:
cp: cannot create regular file 'out/1-name-of-test.log': No such file or directory

**To Reproduce**

1. Create example `file.bats` with following contents
```
 @test "test_that_does_cd" {
     mkdir test
     cd test
     echo yey from test directory
 }
```
2. create a output directory:
`mkdir out`
2. run `bats --gather-test-outputs-in out file.bats`
3. See the error in the output:
```file.bats
 cp: cannot create regular file 'out/1-test_that_does_cd.log': No such file or directory                                                                   1/1
 ✓ test_that_does_cd

1 test, 0 failures
```
4. See no log file in out:
```ls -al out/
total 8
drwxr-xr-x 2 user group 4096 Aug 19 19:17 ./
drwxr-xr-x 4 user group 4096 Aug 19 19:18 ../
```

**Expected behavior**
I expect what the test does to not affect what bats does.

 - Bats 1.11.1
 - Debian GNU/Linux 12
 - GNU bash, version 5.2.15(1)-release (x86_64-pc-linux-gnu)
 - custom dir installed bats.

**Additional context**
I'm trying to figure out a way to have jenkins parse any ThreadSanitizer, LeakSanitizer and MemorySanitizer output from when our test cases run. That's why I'd would like to be able to dump any output from tests into log files so I can parse those and try to find those issues.

The way --gather-test-outputs-in works makes it kinda hard, when it uses a directory that's relative to the .bats file, and not relative to where bats is run from, like how `--report-formatter junit -o <dir>` works, but that's another issue.

jnasbyupgrade:
Fixes #1132

## Problem

When using `--gather-test-outputs-in` with a relative path, if a test changes directory using `cd`, the `cp` command fails with "No such file or directory" because relative paths become invalid when the current working directory changes.

## Solution

Convert relative paths to absolute paths when processing the `--gather-test-outputs-in` option in the `bats` script, ensuring the path remains valid regardless of directory changes within tests.

Also add a defensive check to ensure `BATS_OUT` exists before copying, preventing errors if a test produces no output.

## Changes

- `libexec/bats-core/bats`: Convert relative output directory paths to absolute paths
- `libexec/bats-core/bats-exec-test`: Check `BATS_OUT` exists before copying
- `test/fixtures/bats/cd_in_test.bats`: Add test fixture that changes directory
- `test/bats.bats`: Add test case to verify fix works with relative paths

## Testing

- ✅ Test passes with relative paths when tests change directory
- ✅ Test passes with absolute paths (backward compatible)
- ✅ Existing tests continue to pass

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 * junit formatter: 
   * avoid interference between env and internals (#1175)
   * remove control characters (\x00-\x08\x0B\x0C\x0E-\x1F) (#1176)
+* fix failures with `--gather-test-outputs-in` when tests change directory (#1183)
 
 ### Changed
 


martin-schulze-vireso:
Thank you for your contribution.

PR #1185, opened 2026-01-18
Revert "Removed duplicate `test` suite runs"

jasonkarns:
Reverts bats-core/bats-core#1172

It was reported that the test suite runs "duplicate" test runs. However, the configuration was intentional. https://github.com/bats-core/bats-core/pull/1172#issuecomment-3573593526

The pull_request event runs the suite against the PR's MERGE_HEAD. The push event runs against the PR's branch HEAD. These are not the same commits, and they don't include the same code. (The branch head will not include any code that has landed in main since the branch diverged, and the merge_head _does_.)

So first and foremost, having both runs sends a valuable signal when only one of the suites fails. If the suite fails only on push, then that means that `main` contains code that resolves the failure. Alternatively, if it fails only on pull_request, then that means that whatever causes the failure exists in within the code being merged from main (or the merge is bad for some other reason).

Beyond the fact that both events run against different source, there is value in the test suite being run more often. Test suites that have flaky tests can exhibit failures that seem random. Resolving these failures is difficult in repos that receive frequent activity, but is drastically more difficult when the "random" failure occurrence is infrequent (as is common in relatively low-activity OSS repos like bats). So having the suite run more frequently decreases the overall length of time that flaky tests can exist within the codebase. (As they would ideally be noticed and resolved sooner.)

Lastly, the change to push events failed to include tag push events. While this could have been addressed by explicitly include tags in the push-event trigger, this uncovers the more insidious problem with configuration: the more configuration that exists, the more likely for errors to creep in due to oversight.

Simplicity in configuration might not (on its own) be enough to keep the "duplicate" suites running. However, taken with all the other benefits outlined above, and the fact that the simple configuration also runs the suite on tags, it is surely an added benefit.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,16 +1,7 @@
 name: Tests
 
 # Controls when the action will run.
-on:
-  pull_request:
-    branches:
-      - main
-      - master
-  push:
-    branches:
-      - main
-      - master
-  workflow_dispatch:
+on: [push, pull_request, workflow_dispatch]
 
 permissions:
   contents: read


jasonkarns:
Catching flaky tests is only the edgiest of side benefits. The main concern is that push and pr don't run against the same code so they aren't duplicate runs.

But I spend less time in this project than others so if the predominate contributors are fine with that, I'll close this. 

Though we should at least get tests running on tags again. That's been broken since the change.

martin-schulze-vireso:
I did not mean to close this, just to discuss potential downsides.

I am just one of many contributors, so we should not focus solely on my workflow.

That being said, I see value in simpler configuration. I also understand why additional signals might shorten the feedback loop.

Maybe we can offer a choice: If the branch name contains e.g.  `no-ci-on-push` we only run CI on the PR?

PR #1203, opened 2026-04-27
Fix #1199: error when first argument is empty string

### LINKED ISSUE #1199: bats exits zero when first argument is empty string
**Describe the bug**
When invoked with an empty first parameter (instead of a file or directory that exists), bats collects 0 tests and exits 0.

**To Reproduce**

1. invoke `bats ""`

**Expected behavior**
Exit 1 (or otherwise nonzero) as it would if the first argument were a file path or directory that did not exist.

 - Bats version: 1.12.0
 - operating system (including version): macOS 15.5
 - `bash --version`: `GNU bash, version 5.3.3(1)-release (aarch64-apple-darwin24.6.0)`
 - Install method: nixpkgs

**Additional context**
The invocation is, of course, nonsensical. But we ran into this in nixpkgs testing when a change caused something like `bats "$varName"` to start expanding to `bats ""`, and we didn't notice since the tests were still passing.

Jai2082001:
When invoked with bats "" (empty string as the first argument), bats previously silently collected 0 tests and exited 0, hiding the missing input from CI scripts. This is particularly dangerous in shell pipelines where bats "$var" expands to bats "" when $var is empty, leading to passing CI with no tests actually running.

This change adds an explicit empty-string check in the argument-processing loop, producing a clear error and exit code 1.

Fixes #1199

### DIFF
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -471,6 +471,11 @@ fi
 
 filenames=()
 for filename in "${arguments[@]}"; do
+  
+  if [[ -z "$filename" ]]; then
+    printf "Error: filename argument cannot be empty\n" >&2
+    exit 1
+  fi
   expand_path "$filename" 'filename'
 
   if [[ -z "$setup_suite_file" ]]; then
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -22,6 +22,12 @@ setup() {
   [ "${lines[1]%% *}" == 'Usage:' ]
 }
 
+@test "empty filename prints an error" {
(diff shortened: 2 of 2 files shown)


martin-schulze-vireso:
While this fixes the problem at hand, I would prefer a more robust solution or at least prove that this is already robust. I wonder why passing an empty string as "filename" does not lead to `file not found` errors?

martin-schulze-vireso:
Closing this in favor of #1212. Thanks for your suggestion, the code is very similar but I wanted more "defense in depth".

PR #1208, opened 2026-05-31
Avoid overwriting $_ by the DEBUG and ERR traps set by Bats

akinomyoga:

**Summary**

In short, the current DEBUG trap set by Bats breaks the value of `$_`, which becomes an issue in testing a shell function that uses the value of `$_`. The DEBUG and ERR traps should preserve the value of `$_`.

**Details**

I want to test a function that uses `$_` in https://github.com/rcaloras/bash-preexec/pull/187. However, the `DEBUG` trap set by Bats seems to overwrite `$_` with a string like `/tmp/bats-run-lfvLSG/bats.201138.src` between the execution of the line to set `$_` and the line to use the value of `$_`. There doesn't seem to be a way to test this function (except for unsetting the trap, though I'm not sure what would be a side effect of removing the hooks installed by Bats in the middle of the test).

For example, see the test case included in this PR. This test case doesn't pass with the `master` branch because the DEBUG trap string `bats_debug_trap "$BASH_SOURCE"` set by Bats will overwrite `$_` to be `"$BASH_SOURCE"`. To reproduce the value of `$_` at the end of the evaluation of the function `bats_debug_trap`, the function in the DEBUG trap should receive an extra dummy argument `"$_"` as in the suggested change. With this change, the test case passes.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
   * **ATTENTION**: In previous versions this was suppressed unintentionally.
     While it might constitute a breaking change for some, we decided the new behavior should be the default because it might uncover hidden errors.
     If you need the old behavior, you can use this wrapper function `suppress_errexit() { "$@" || return $?; }` like `run suppress_errexit <your command...>`
+* avoid overwriting `$_` by the DEBUG and ERR traps set by Bats (#1208)
 
 ### Changed
 
--- a/lib/bats-core/tracing.bash
+++ b/lib/bats-core/tracing.bash
@@ -382,8 +382,8 @@ bats_setup_tracing() {
   done
 
   # turn on traps after setting excludes to avoid tracing the exclude setup
-  trap 'bats_debug_trap "$BASH_SOURCE"' DEBUG
-  trap 'bats_error_trap' ERR
+  trap 'bats_debug_trap "$BASH_SOURCE" "$_"' DEBUG
+  trap 'bats_error_trap "$_"' ERR
 }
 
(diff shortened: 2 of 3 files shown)


akinomyoga:
The Shellcheck errors in the CI are false positives for the special shell variable `_`. I reported this to the upstream Shellcheck at  https://github.com/koalaman/shellcheck/issues/3478. For now, I manually added `disable=SC2030` and `disable=SC2031` to the respective lines.

The following test failure on "npm_on_windows (windows-2025)" doesn't seem to be related to the present change.

```
not ok 77 CTRL-C aborts and fails the current setup_file in 11693ms
# (from function `single-use-latch::wait' in file test/concurrent-coordination.bash, line 44,
#  in test file test/bats.bats, line 950)
#   `single-use-latch::wait hang_in_setup_file 1 10' failed
# ERROR: single-use-latch hang_in_setup_file timed out
# mv: cannot stat '/tmp/bats-run-3kJm1y/suite/latch-hang_in_setup_file': No such file or directory
```

In fact, in another run of the CI test (with essentially no change in the code), test case 77 passed. However, this time, another test case 74 failed:

```
not ok 74 CTRL-C aborts and fails the current run in 1013ms
# (in test file test/bats.bats, line 846)
#   `[ "${lines[1]}" == "not ok 1 test" ]' failed
```

I suspect that those tests are unstable even if the implementation is fine. Can I ignore those test failures? Or should I resolve those issues to make this PR reviewed?

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for this fix.

akinomyoga:
Thank you!

PR #1209, opened 2026-06-21
Fix setup_suite PATH changes breaking internals

c0da-ricon:
## Summary

Fix `bats-exec-suite` so `setup_suite` can mutate `PATH` without breaking Bats' own internal helper lookup.

`setup_suite` runs in the suite process before `bats-exec-suite` invokes `bats-exec-file`. If `setup_suite` replaces `PATH`, for example by evaluating a tool-version manager environment, the unqualified internal command can disappear and the run fails before executing any tests.

## Maintainer-requested revision

This revision centralizes internal executable lookup through `bats_internal_path` and `bats_execute_internal`, based on `BATS_LIBEXEC` rather than `PATH`.

It routes the built-in formatter, suite, gather-tests, exec-file, exec-test, and preprocess paths through that boundary. Direct `bats-exec-suite` invocation derives `BATS_LIBEXEC` from its own location. GNU Parallel still receives a shell-quoted absolute `bats-exec-file` path, including when the libexec path contains spaces. Absolute external formatter paths keep their existing behavior.

## Reproducer

With Bats 1.13.0:

```bash
cat > setup_suite.bash <<'EOF'
setup_suite() {
  export PATH=/usr/bin:/bin
}
EOF

cat > sample.bats <<'EOF'
#!/usr/bin/env bats
@test "one" {
  [ 1 -eq 1 ]
}
EOF

bin/bats sample.bats
```

Before this change, that fails with:

```text
bats-exec-suite: line 323: bats-exec-file: command not found
# bats warning: Executed 0 instead of expected 1 tests
```

## Validation

- all 488 current test cases passed locally across bounded suite shards
- focused Bash 3.2 run passed 31/31, including serial and parallel `setup_suite` PATH mutation
- `./shellcheck.sh` passed with ShellCheck 0.11.0
- `git diff --check` passed
- upstream `Tests` run `30710912970` is waiting for maintainer approval (`action_required`)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ### Fixed
 
+* resolve Bats internal tools independently of `PATH` mutations in `setup_suite` (#1209)
 * pretty formatter was not the default on interactive shells anymore (#1220)
 
 ### Documentation
--- a/lib/bats-core/common.bash
+++ b/lib/bats-core/common.bash
@@ -271,6 +271,18 @@ bats_format_file_line_reference_uri() {
   printf -v "$output" "file://%s:%d" "$filename" "$line"
 }
 
+bats_internal_path() { # <output-variable> <internal-command>
+  local output=$1 command=$2
+  printf -v "$output" '%s/%s' "${BATS_LIBEXEC?}" "$command"
+}
+
+bats_execute_internal() { # <internal-command> <args...>
+  local command=$1 path
+  shift
+  bats_internal_path path "$command"
+  "$path" "$@"
+}
+
 # execute command with backed up path
 # to prevent path mocks from interfering with our internals
 bats_execute() { # <command...>
--- a/lib/bats-core/preprocessing.bash
+++ b/lib/bats-core/preprocessing.bash
@@ -9,7 +9,7 @@ bats_export_preprocess_source_BATS_TEST_SOURCE() {
 bats_preprocess_source() { # index
   bats_export_preprocess_source_BATS_TEST_SOURCE
   # shellcheck disable=SC2153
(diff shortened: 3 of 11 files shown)


martin-schulze-vireso (changes requested):
While I appreciate the suggestion and see some merit in this, it is far from enough. There are many more places where Bats internals are called that would fail even with the proposed fix.

My preferred solution would be a wrapper function (e.g. `call-bats-internal`) that is fully independent of the path. I  like the idea of deriving `BATS_LIBEXEC` from `bats-exec-suite`.  With that in place we would have to call all Bats tools (e.g. `exec-file`,`exec-test`, `gather-tests`) via that wrapper.

c0da-ricon:
Thanks for the review, @martin-schulze-vireso. I’ve reworked the PR based on your feedback. Would you be willing to take another look?

martin-schulze-vireso (approved):

martin-schulze-vireso:
It seems the new test fails on freebsd, probably due to a different PATH where bash is found:
```
  not ok 454 PATH mutations in setup_suite do not break Bats internals
  # (from function `reentrant_run' in file test/test_helper.bash, line 79,
  #  in test file test/suite_setup_teardown.bats, line 68)
  #   `reentrant_run -0 bats "$FIXTURE_ROOT/path_mutation/"' failed, expected exit code 0, got 1
  # 1..1
  # env: bash: No such file or directory
  # # bats warning: Executed 0 instead of expected 1 tests
  # Last output:
  # 1..1
  # env: bash: No such file or directory
  # # bats warning: Executed 0 instead of expected 1 tests
  not ok 455 parallel PATH mutations in setup_suite do not break Bats internals
  # (from function `reentrant_run' in file test/test_helper.bash, line 79,
  #  in test file test/suite_setup_teardown.bats, line 99)
  #   `BATS_LIBEXEC="$libexec_with_spaces" \' failed, expected exit code 0, got 1
  # ERROR: flock/shlock is required for parallelization within files!
  # Last output:
  # ERROR: flock/shlock is required for parallelization within files!

PR #1226, opened 2026-07-22
fix: timeout watchdog killer 

### LINKED ISSUE #1206: BATS_TEST_TIMEOUT hangs full duration on fast-failing tests in 1.13.0 (related to #1067, but for fail not skip)
## Summary

`BATS_TEST_TIMEOUT=N` makes bats main process hang for the full `N` seconds when a test exits non-zero (false, exit 1, failed assertion). The test process itself completes immediately, but bats blocks waiting on the watchdog subshell's pipe to close. The "`kill: ... - No such process`" stderr line is the smoking gun — it's the watchdog firing AFTER its full sleep completes, by which time the test process tree is gone.

This is similar to but distinct from #1067 (which was about *skipped* tests). #1067 closed 2025-11-06 just before 1.13.0 released 2025-11-07; that fix appears to address the SKIP path but not the FAIL path. Verified empirically against released 1.13.0 (2026-05-09).

## Reproduction

```bash
$ bats --version
Bats 1.13.0

$ cat > /tmp/fail.bats <<'EOT'
#!/usr/bin/env bats
@test "passing test"      { true; }
@test "fast failing test" { false; }
@test "another fail"      { exit 1; }
EOT

$ time BATS_TEST_TIMEOUT=8 bats /tmp/fail.bats
1..3
ok 1 passing test
not ok 2 fast failing test
# (in test file /tmp/fail.bats, line 3)
#   `@test "fast failing test" { false; }' failed
not ok 3 another fail
# (in test file /tmp/fail.bats, line 4)
#   `@test "another fail" { exit 1; }' failed
/usr/lib/bats-core/bats-exec-test: line 263: kill: (PID) - No such process

real    0m8.137s

$ time bats /tmp/fail.bats   # baseline, no BATS_TEST_TIMEOUT
real    0m0.116s
```

`BATS_TEST_TIMEOUT=8` adds ~8 seconds to a sub-second test run that should complete in ~0.1s. The watchdog from `bats_start_timeout_countdown` does its full `sleep $timeout` and only THEN tries to `kill` a process that no longer exists.

## Expected behavior

`BATS_TEST_TIMEOUT=N` should bound MAX wall-clock per test. A test that returns in 0.05s (whether passing, failing, or skipping) should not delay subsequent test scheduling for N seconds.

## Verified scenarios

| Scenario                          | BATS_TEST_TIMEOUT | Wall clock | Verdict |
|-----------------------------------|-------------------|------------|---------|
| Single passing test               | 10s               | 0.07s      | ✓ correct |
| Slow passing, timeout fires       | 2s                | 2.12s      | ✓ correct |
| Single skipped test               | 10s               | 0.07s      | ✓ correct (was #1067) |
| **Single failing test (`false`)** | **8s**            | **8.14s**  | ✗ hangs |
| **Multiple failing tests**        | **8s**            | **8.14s**  | ✗ hangs |
| **`exit 1`**                      | **8s**            | **~8s**    | ✗ hangs |

## Root cause hypothesis

From reading `bats-exec-test:267-289` and `:336-338`: `bats_start_timeout_countdown` forks an outer subshell that holds open inherited FDs to bats main's read pipe. On a failing test, the SIGTERM sent by `bats_abort_timeout_countdown` does not actually interrupt the watchdog's `wait` (the subshell stays alive holding the pipe open), so bats main blocks on pipe-EOF for the full timeout. The watchdog exits naturally only after its `sleep $timeout` completes, at which point the kill at line 263 finds no process to signal — hence the "No such process" stderr.

This may be the same FD-leak class as #1020 (`run …` hangs), where the fix needs to ensure the watchdog releases all inherited FDs before going to `wait`/`sleep`.

## Workaround

Don't set `BATS_TEST_TIMEOUT`. Use `timeout(1)` for suite-level wall-clock bound:

```bash
timeout 1800 bats --tap *.bats || exit 2
```

Loses per-test granularity (the failing test's identity in a hung suite is hard to pin down) but suite cap keeps things bounded. We're shipping this workaround in our `phased-breakout` orchestrator (reverie-orchestrate). It's load-bearing for our CI not hanging.

## Environment

- `bats --version`: 1.13.0
- `bash --version`: GNU bash 5.x on Linux x86_64
- Install method: npm (`npm install -g bats`)
- bash variant: GNU bash via npm-shipped libexec

Happy to dig further into the FD-leak hypothesis if the team can point at the right place to look. The fix may share structure with the #1020 → #1067 closeout series.

martin-schulze-vireso:
The BATS_killer_pid variable was not global, leading to `kill ""`. The error message was suppressed.

fixes #1206


martin-schulze-vireso:
Thanks for the reproducer in #1207 @HonorCodes

PR #1236, opened 2026-08-10
fix(bats-preprocess): make test-name encoding locale-independent

### LINKED ISSUE #1233: Tests with non-ASCII names are silently never executed under a UTF-8 locale (locale-dependent [[:alnum:]] in bats_encode_test_name)
## Summary

`bats_encode_test_name` in `libexec/bats-core/bats-preprocess` decides between two
encoding strategies with a **locale-dependent** regex:

```sh
if [[ ! "$name" =~ [^[:alnum:]\ _-] ]]; then
```

Under `LC_ALL=C` this is false for any non-ASCII name, so the safe hex-encoding branch
runs and the generated function name is pure ASCII.

Under a UTF-8 locale, `[[:alnum:]]` matches CJK, Cyrillic, Hangul, fullwidth and accented
letters. The condition becomes **true**, the "fast" branch runs, and the raw multibyte
bytes are embedded directly into the generated shell function name.

For a subset of those names, `declare -F` in `bats_perform_test`
(`libexec/bats-core/bats-exec-test:320`) then fails to find the function, and the test
**never runs** — while the TAP plan line still counts it.

## Impact

The `1..N` plan line reports the declared count. The tests that ghosted produce no
`ok`/`not ok` line at all. On a suite where this went unnoticed, we measured
**889 declared / 804 executed — 85 tests that had not run for an unknown length of time.**

bats does print `# bats warning: Executed N instead of expected M tests` and exits
non-zero, so this is not *fully* silent. But the diagnostic is easy to lose in practice:
it goes to stderr as a comment line, and any pipeline that ends in `| tail`, a formatter,
or a CI step that only inspects the last lines will show a plan of N and a screen of
`ok` lines. The tests are gone, and the report looks healthy.

This is a false-green failure mode in a test runner, which is why we are filing it even
though a warning exists.

## Reproduction

```sh
cat > t.bats <<'EOF'
@test "pure ascii" { true; }
@test "あ" { true; }
@test "中" { true; }
@test "件" { true; }
@test "Кириллица" { true; }
@test "한국어" { true; }
@test "café" { true; }
@test "Ａ" { true; }
EOF

LC_ALL=C           bats t.bats; echo "rc=$?"
LC_ALL=en_US.UTF-8 bats t.bats; echo "rc=$?"
```

### Actual

```
$ LC_ALL=C bats t.bats
1..8
ok 1 pure ascii
ok 2 あ
ok 3 中
ok 4 件
ok 5 Кириллица
ok 6 한국어
ok 7 café
ok 8 Ａ
rc=0

$ LC_ALL=en_US.UTF-8 bats t.bats
1..8
ok 1 pure ascii
bats: unknown test name `$'test_\343\201\202''
bats: unknown test name `$'test_\344\270\255''
ok 4 件
bats: unknown test name `$'test_\320\232\320\270...''
bats: unknown test name `$'test_\355\225\234...''
ok 7 café
ok 8 Ａ
# bats warning: Executed 4 instead of expected 8 tests
rc=1
```

Same binary, same file, same bash. Only `LC_ALL` differs.

## Confirming the branch flip directly

Extracting `bats_encode_test_name` verbatim and calling it on the same inputs:

| input | `LC_ALL=C` | `LC_ALL=en_US.UTF-8` |
|---|---|---|
| `あ` | slow path → `test_-ffff…e3-ffff…81-ffff…82` | **fast path → `test_あ`** |
| `件` | slow path → `test_-ffff…e4-ffff…bb-ffff…b6` | **fast path → `test_件`** |
| `café` | slow path (ASCII-safe) | **fast path → `test_café`** |

So the locale controls which branch is taken. That part is fully determined.

## What we could NOT determine (open question for maintainers)

Under a UTF-8 locale **every** one of the names above takes the fast path and gets a raw
multibyte function name — yet only some of them fail to dispatch:

| name | codepoint | UTF-8 | result |
|---|---|---|---|
| `件` | U+4EF6 | `E4 BB B6` | **runs** |
| `中` | U+4E2D | `E4 B8 AD` | ghosts |
| `亜` | U+4E9C | `E4 BA 9C` | ghosts |
| `あ` | U+3042 | `E3 81 82` | ghosts |
| `ア` | U+30A2 | `E3 82 A2` | ghosts |
| `К` | U+041A | `D0 9A` | ghosts |
| `한` | U+D55C | `ED 95 9C` | ghosts |
| `é` | U+00E9 | `C3 A9` | **runs** |
| `Ａ` | U+FF21 | `EF BC A1` | **runs** |

`件` and `中` are both CJK ideographs in the same block and differ only in codepoint, so
**the boundary is per-codepoint, not per-script and not per-length.** We tested and ruled
out: number of multibyte characters, script/block, and mixing with ASCII.

Separately, bash itself has no problem with these names — defining and calling
`test_あ`, `test_中`, `test_한` etc. directly all succeed in the same shell. So the loss
happens somewhere between the preprocessor emitting the definition and
`declare -F "${BATS_TEST_NAME%% *}"` resolving it, and we did not isolate where.

We are reporting the boundary as **uncharacterized** rather than guessing.

## Why the fix does not depend on that open question

Regardless of which raw-multibyte names happen to survive, the fast path exists as an
optimisation for names that are *already* safe as shell identifiers. Non-ASCII names are
not. Making the guard locale-independent restores the hex-encoding path for every
non-ASCII name and removes the whole class:

```diff
-  if [[ ! "$name" =~ [^[:alnum:]\ _-] ]]; then
+  if [[ ! "$name" =~ [^a-zA-Z0-9\ _-] ]]; then
```

An explicit ASCII class is preferable to wrapping the comparison in `LC_ALL=C`, because
it states the intent (shell-identifier-safe) rather than depending on ambient locale.

The same locale sensitivity applies to the `[[ "$char" =~ [[:alnum:]] ]]` test on line 22
inside the slow path, which should get the same treatment for consistency.

## Environment

- bats-core **1.13.0** (Homebrew), macOS 26.x, arm64, `bash` 3.2.57
- Also reproduced on **bats-core 1.14.0** on a second macOS machine
  (21-case matrix: `declared=21 / executed=16 / 5 ghosts / rc=1` under `en_US.UTF-8`;
  `declared=21 / executed=21 / 0 ghosts / rc=0` under `LC_ALL=C`)

The version is not the variable — the locale is. A machine whose ambient environment has
`LC_ALL`/`LC_CTYPE`/`LANG` unset behaves like `LC_ALL=C` and will not show the bug, which
makes it easy for this to reproduce on one developer's machine and not another's.

## Detection for existing suites

For anyone who wants to check whether their suite is already affected, comparing the
declared count against the executed count per file is sufficient:

```sh
bats tests/ > out.tap 2>&1
declared=$(/usr/bin/grep -oE '^1\.\.[0-9]+' out.tap | /usr/bin/grep -oE '[0-9]+$')
executed=$(/usr/bin/grep -cE '^(ok|not ok) ' out.tap)
[ "$declared" = "$executed" ] || echo "shortfall: $((declared - executed))"
```

## Detection tooling

For anyone who wants to check an existing suite, I published the two checks I used while
investigating this: https://github.com/KinohTaGo/bats-ghost-test-check

The static one (scan `@test` names for non-ASCII) is the one worth wiring into CI — the
dynamic declared-vs-executed check runs bats and therefore inherits the same locale
condition that triggers the bug, so it reports clean on a runner with `LC_ALL` unset.

vjymisal0:
## Bug

`bats_encode_test_name()` in `libexec/bats-core/bats-preprocess` picks between two
encoding strategies for a test name using a **locale-dependent** regex:

```sh
if [[ ! "$name" =~ [^[:alnum:]\ _-] ]]; then
```

Under `LC_ALL=C` this is false for any non-ASCII name, so the safe hex-encoding
branch runs and the generated shell function name stays pure ASCII.

Under a UTF-8 locale, `[[:alnum:]]` matches CJK, Cyrillic, Hangul, fullwidth and
other multibyte "letters", so the condition flips to true, the "fast" branch runs
instead, and the raw multibyte bytes get embedded directly into the generated
function name. For a subset of codepoints this causes `declare -F` (in
`bats_perform_test`, `libexec/bats-core/bats-exec-test`) to fail to find the
generated function, so the test **silently never runs** — while the `1..N` TAP
plan line still counts it as declared. This is a false-green failure mode: bats
does print `# bats warning: Executed N instead of expected M tests` to stderr,
but that's easy to miss in a CI log that only surfaces the last lines or the exit
code of a formatter/pipeline.

This is #1233, which already narrows the root cause to this exact line with a
detailed repro (`889 declared / 804 executed` on a real suite) and proposes
this exact fix. No PR had been opened for it, so I opened this one.

## Fix

Use an explicit ASCII character class (`a-zA-Z0-9`) instead of the
locale-dependent POSIX `[[:alnum:]]` class, both in the branch-selection regex
and in the per-character check inside the hex-encoding loop, so any non-ASCII
name always takes the safe hex-encoding path regardless of the caller's locale.

## Testing

- Verified the branch-selection bug directly: running
  `libexec/bats-core/bats-preprocess` on a fixture with CJK/Cyrillic/Hangul test
  names under `LC_ALL=C` vs `LC_ALL=en_US.UTF-8` produces different generated
  function names before the fix (safe hex-encoded under `C`, raw multibyte
  under UTF-8), and identical (always hex-encoded) after the fix.
- I could not reproduce the downstream `declare -F` failure itself in my dev
  environment (bash 5.2 on MSYS/Windows) — #1233 reports the exact
  ghosting boundary as codepoint- and bash-build-dependent, reproduced on two
  separate macOS/bash-3.2 machines. The fix removes the locale-dependent branch
  selection that #1233 identifies as the root cause regardless of which
  specific codepoints ghost on a given platform.
- Added `test/fixtures/bats/non_ascii_test_names.bats` and a regression test in
  `test/bats.bats` ("test names with non-ASCII characters run under a UTF-8
  locale") that runs this fixture under `LC_ALL=en_US.UTF-8` and asserts all 6
  tests execute (`1..6` and 6 `ok` lines). It's skipped if `en_US.UTF-8` isn't
  installed on the runner, following the existing skip pattern used for the
  unicode `BATS_CODE_QUOTE_STYLE` test.
- Ran the full `test/bats.bats` suite (127 tests) before and after the change;
  all pass, no regressions.

Closes #1233.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 ### Fixed
 
 * pretty formatter was not the default on interactive shells anymore (#1220)
+* `bats_encode_test_name()` used the locale-dependent `[[:alnum:]]` character class, causing test names with non-ASCII characters to be silently skipped under some locales (#1236)
 
 ### Documentation
 
--- a/libexec/bats-core/bats-preprocess
+++ b/libexec/bats-core/bats-preprocess
@@ -6,7 +6,7 @@ bats_encode_test_name() {
   local result='test_'
   local hex_code
 
-  if [[ ! "$name" =~ [^[:alnum:]\ _-] ]]; then
+  if [[ ! "$name" =~ [^a-zA-Z0-9\ _-] ]]; then
     name="${name//_/-5f}"
     name="${name//-/-2d}"
     name="${name// /_}"
@@ -19,7 +19,7 @@ bats_encode_test_name() {
       char="${name:$i:1}"
       if [[ "$char" == ' ' ]]; then
         result+='_'
-      elif [[ "$char" =~ [[:alnum:]] ]]; then
+      elif [[ "$char" =~ [a-zA-Z0-9] ]]; then
         result+="$char"
       else
         printf -v 'hex_code' -- '-%02x' \'"$char"
--- a/test/bats.bats
+++ b/test/bats.bats
@@ -529,6 +529,24 @@ END_OF_ERR_MSG
   [[ "${lines[3]}" == "ok 3 test 3" ]]
 }
 
+@test "test names with non-ASCII characters run under a UTF-8 locale (see #1233)" {
+  if ! locale -a 2>/dev/null | grep -qi '^en_US\.utf-\?8$'; then
+    skip "en_US.UTF-8 locale is not available on this system"
+  fi
+
+  # shellcheck disable=SC2030,SC2031
+  REENTRANT_RUN_PRESERVE+=(LC_ALL)
+  LC_ALL=en_US.UTF-8 reentrant_run --separate-stderr bats "$FIXTURE_ROOT/non_ascii_test_names.bats"
+  [ "$status" -eq 0 ]
+  [ "${lines[0]}" = "1..6" ]
+  # all 6 tests must actually run (none may be silently skipped/"ghosted"
+  # due to locale-dependent [[:alnum:]] matching in bats_encode_test_name)
+  local ok_count
+  ok_count="$(grep -c '^ok ' <<<"$output")"
+  [ "$ok_count" -eq 6 ]
+  [ -z "$stderr" ]
+}
+
 @test "report correct line on unset variables" {
   LANG=C reentrant_run bats "$FIXTURE_ROOT/unbound_variable.bats"
   [ "$status" -eq 1 ]
--- a/test/fixtures/bats/non_ascii_test_names.bats
+++ b/test/fixtures/bats/non_ascii_test_names.bats
@@ -1,0 +1,23 @@
+@test "pure ascii" {
+  true
+}
+
+@test "あ" {
+  true
+}
+
+@test "中" {
+  true
+}
+
+@test "Кириллица" {
+  true
+}
+
+@test "한국어" {
+  true
+}
+
+@test "café" {
+  true
+}


martin-schulze-vireso (approved):

vjymisal0:
The PR is approved; the remaining red job is the FreeBSD/flock CI path. I don't have a FreeBSD environment locally to reproduce that one, so I'll wait for maintainer guidance or a rerun if it's infra-flaky.

martin-schulze-vireso:
Thanks for contributing.

PR #1238, opened 2026-08-15
fix: negate the whole GNU parallel availability check

### LINKED ISSUE #1237: parallel guard in bats-exec-suite can never fire: `!` binds only to the first clause
### Summary

The guard in `libexec/bats-core/bats-exec-suite` that is meant to refuse `--jobs` without GNU parallel can never fire, because `!` binds only to the first clause of the `&&` chain. As a result, bats hands moreutils' `parallel` a set of GNU flags and the run dies with a confusing message instead of the clear abort the guard was written to give.

`master`, [`libexec/bats-core/bats-exec-suite` line 110](https://github.com/bats-core/bats-core/blob/master/libexec/bats-core/bats-exec-suite#L110):

```bash
if ! type -p "${parallel_binary_name}" >/dev/null && "${parallel_binary_name}" --version &>/dev/null && [[ -z "$bats_no_parallelize_across_files" ]]; then
  abort "Cannot execute \"${num_jobs}\" jobs without GNU parallel"
```

The condition reads as `(! type -p parallel) && (parallel --version) && ...`, i.e. *"parallel is absent **and** parallel answers `--version`"*. Those are contradictory, so the abort is unreachable in every configuration.

This came in with #691, which fixed #672. That PR got the grouping right in `test/parallel.bats`:

```bash
(type -p parallel &>/dev/null && parallel --version &>/dev/null) || skip "--jobs requires GNU parallel"
```

but the parentheses did not make it into `bats-exec-suite`.

### Reproduction

Debian trixie, bats 1.11.1, moreutils' `parallel` first on PATH (`/usr/bin/parallel` from the `moreutils` package, which Debian's `parallel` package normally diverts to `/usr/bin/parallel.moreutils`):

```console
$ printf '@test "one" { true; }\n@test "two" { true; }\n' > t.bats
$ PATH="/path/to/moreutils-parallel:$PATH" bats --jobs 2 t.bats
1..2
parallel: invalid option -- '-'
parallel [OPTIONS] command -- arguments
	for each argument, run command with argument, in parallel
parallel [OPTIONS] -- commands
	run specified commands in parallel
# bats warning: Executed 0 instead of expected 2 tests
$ echo $?
1
```

**Expected:** `Cannot execute "2" jobs without GNU parallel`.

**Actual:** moreutils' usage text and `Executed 0 instead of expected 2 tests`. The exit status is non-zero, so this is not a silent pass — but the diagnosis is exactly the one the abort message exists to prevent, and it is what #672 was reported as in the first place.

### Suggested fix

Group the two clauses the way `test/parallel.bats` already does:

```bash
if ! { type -p "${parallel_binary_name}" >/dev/null && "${parallel_binary_name}" --version &>/dev/null; } && [[ -z "$bats_no_parallelize_across_files" ]]; then
```

The discriminator itself is sound, for the record: moreutils' `parallel` exits 1 on `--version` (verified against `moreutils` 0.69-1), so `--version` succeeding does distinguish GNU parallel from it. Only the precedence is wrong.

I'm happy to open a PR with that one-liner if it would help.

### Context

Found while making a project's pre-commit bats gate portable across a fleet that includes Debian and Mint boxes with `moreutils` installed. We work around it by asking the binary what it is — `parallel --version | head -1 | grep -q '^GNU parallel'` — rather than relying on bats to refuse.

### Versions

- bats 1.11.1 (Debian trixie) for the reproduction; the condition is byte-identical on `master` as of today.
- moreutils 0.69-1.

VXNCXNX:
Fixes #1237.

## What's broken

The guard that refuses `--jobs` without GNU parallel can never fire. `!` binds only to the first command of an `&&` chain, so when parallel is absent, `! type -p parallel` is true but the next clause runs the missing binary and fails, and the chain is false. When parallel is present the first clause is false and the chain short-circuits. Both directions skip the abort.

## Repro

```
$ if ! type -p nonexistent_xyz >/dev/null && "nonexistent_xyz" --version &>/dev/null && [[ -z "$x" ]]; then echo FIRED; else echo "DID NOT FIRE"; fi
DID NOT FIRE
```

End to end on a machine without GNU parallel, `bats --jobs 2 test/` runs zero tests and dies on `parallel: command not found` instead of printing the guard's message.

## The fix

Group the availability check and negate it as a whole, the same shape `test/parallel.bats` already uses:

```
(type -p parallel &>/dev/null && parallel --version &>/dev/null) || skip ...
```

The `bats_no_parallelize_across_files` clause stays outside the group, so its behaviour is unchanged.

## Verification

Truth table over all four combinations of binary present/absent and the flag set/empty: the new condition fires in exactly one case, parallel absent with the flag empty, and the old one fires in none. `bin/bats test/bats.bats` passes 126/126.

One thing I could not verify: GNU parallel is not installed on this machine, so all 18 tests in `test/parallel.bats` skip here, before and after. That suite gives no direct coverage of this change and it is worth one run on a box that has parallel.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 ### Fixed
 
 * pretty formatter was not the default on interactive shells anymore (#1220)
+* `--jobs` now aborts when GNU parallel is unavailable instead of failing later (#1237, #1238)
 
 ### Documentation
 
--- a/libexec/bats-core/bats-exec-suite
+++ b/libexec/bats-core/bats-exec-suite
@@ -111,7 +111,7 @@ while [[ "$#" -ne 0 ]]; do
 done
 
 if [[ "$num_jobs" != 1 ]]; then
-  if ! type -p "${parallel_binary_name}" >/dev/null && "${parallel_binary_name}" --version &>/dev/null && [[ -z "$bats_no_parallelize_across_files" ]]; then
+  if ! (type -p "${parallel_binary_name}" >/dev/null && "${parallel_binary_name}" --version &>/dev/null) && [[ -z "$bats_no_parallelize_across_files" ]]; then
     abort "Cannot execute \"${num_jobs}\" jobs without GNU parallel"
   fi
   # shellcheck source=lib/bats-core/semaphore.bash


martin-schulze-vireso (approved):

VXNCXNX:
This has been approved since the 5th. Anything blocking it, or anything you would like changed?

martin-schulze-vireso:
No, I was just lacking the time ti have a look again.

Thanks for your contribution.

PR #1240, opened 2026-08-17
feat: add BATS_ALLOW_EMPTY_SUITE environment variable

### LINKED ISSUE #1239: --allow-empty-suite forces a hard dependency on bats >= 1.14
### Describe the bug

Since v1.14.0 (PR #1211) an empty test suite results in an error, which is incompatible with older bats versions. It breaks some existing setups (see below for an example). The suggested workaround is to add `--allow-empty-suite` to get the old behavior back.

Unfortunately, the workaround itself is also incompatible with older bats versions, as any old version will error out with `Bad command line option '--allow-empty-suite'`.

So for a project that legitimately produces an empty suite in some configuration, the choice is:

* add `--allow-empty-suite` and drop support for every bats before 1.14, or
* pin bats to <= 1.13 and stay on an old version.

### To Reproduce

Real-world example, https://github.com/containers/conmon/pull/672: CRI-O's test runner always makes two bats passes, the second one selecting the serial tests:

```
execute bats --jobs "$JOBS" --tap "${TESTS[@]}" --filter-tags '!crio:serial'
execute bats --tap "${TESTS[@]}" --filter-tags 'crio:serial'
```

When the requested test set contains no serial test, the second pass legitimately has nothing to run, and with v1.14.0 CI fails:

```
+ bats --tap critest.bats --filter-tags crio:serial
1..0
ERROR: Found no tests. (Try `--allow-empty-suite`?)
```

Adding `--allow-empty-suite` there would have broken the distro-provided older bats that other jobs use, so we had to pin bats to v1.13.0 instead.

### Expected behavior

A way to opt into the old behavior that older bats versions silently ignore, i.e. an environment variable — `BATS_ALLOW_EMPTY_SUITE=1` — with the same effect as `--allow-empty-suite`. Older versions ignore an unknown `BATS_*` variable and keep exiting 0 on an empty suite, which is exactly the behavior being asked for, so one setting works across versions.

There is precedent for flags having an environment variable equivalent, e.g. `--code-quote-style` / `$BATS_CODE_QUOTE_STYLE`.

### Environment

* bats 1.14.0

kolyshkin:
Fixes #1239.

`--allow-empty-suite` (added in v1.14.0) can only be used by projects that no longer need to run with bats < 1.14, since older versions abort on the unknown flag. This adds `BATS_ALLOW_EMPTY_SUITE`: set it to any non-empty value and an empty suite exits 0, just like the flag. Older bats versions ignore the variable and already exit 0 on an empty suite, so the same setting works across versions — which is what motivated this, see https://github.com/containers/conmon/pull/672 where pinning bats to v1.13.0 was the only option.

The second commit (feel free to drop if you don't like it) rewords the error message to say what to do instead of asking a question, and mentions the variable next to the flag:

```
ERROR: Found no tests. Use `--allow-empty-suite` or `BATS_ALLOW_EMPTY_SUITE=1` to suppress this error.
```

Tests cover the variable enabling an empty suite, an empty value not enabling it, and no effect when tests exist; the flag's own positive case was missing and is added too. Docs (`--help`, man page, CHANGELOG) updated.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+* `$BATS_ALLOW_EMPTY_SUITE` environment variable as an alternative to `--allow-empty-suite`, which keeps test setups compatible with Bats versions before 1.14 that don't know the flag (#1240)
+
 ### Fixed
 
 * pretty formatter was not the default on interactive shells anymore (#1220)
--- a/libexec/bats-core/bats
+++ b/libexec/bats-core/bats
@@ -36,6 +36,7 @@ HELP_TEXT_HEADER
 
   --abort                   Stop execution of suite on first failed test
   --allow-empty-suite       Exit with code 0 (instead of the default code 1) when no tests are found.
+                            Can also be set via $BATS_ALLOW_EMPTY_SUITE
   -c, --count               Count test cases without running any tests
   --code-quote-style <style>
                             A two character string of code quote delimiters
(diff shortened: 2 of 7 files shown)


kolyshkin:
@martin-schulze-vireso PTAL (this would be nice to have in the next dot release)

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for contributing.

PR #1241, opened 2026-08-17
CI refresh

kolyshkin:
Update the actions that are behind their latest upstream releases:

| Action | From | To |
|---|---|---|
| `actions/setup-node` | v6.0.0 | v7.0.0 |
| `ossf/scorecard-action` | v2.4.3 | v2.4.4 |
| `github/codeql-action/upload-sarif` | v4.30.8 | v4.37.8 |
| `EnricoMi/publish-unit-test-result-action` | v2.21.0 | v2.24.0 |
| `acj/freebsd-firecracker-action` | v0.10.1 | v0.11.0 |
| `docker/login-action` | v4.4.0 | v4.6.0 |
| `docker/setup-buildx-action` | v4.2.0 | v4.3.0 |

All pins are SHAs of the corresponding tags (annotated tags dereferenced to
the underlying commit), matching the existing convention.

The FreeBSD guest image shipped with `freebsd-firecracker-action` v0.11.0 no
longer provides the `C.UTF-8` locale that the environment asks for, so perl
(which GNU parallel is built on) prints locale warnings to stdout. Those
warnings land in the captured output and shift the line indices some tests
assert on, e.g. `--abort prevents further tests from running in parallel mode`.
The FreeBSD job now asks for `LANG=C LC_ALL=C`, a locale the image actually
has.

Also adds the missing `# v4.6.0` version comment to the second
`docker/login-action` pin in `release_dockerhub.yml`, so dependabot picks it
up like the other one.

`actions/setup-node` v7 is a major bump and the three call sites request
`node-version: v18.20.2` (EOL); the npm jobs pass on all of Linux, macOS and
Windows with it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -13,7 +13,7 @@ permissions:
 jobs:
   codespell:
     name: Check for spelling errors
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
 
     steps:
       - name: Checkout
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -14,7 +14,7 @@ permissions:
 
 jobs:
   dependency-review:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: 'Checkout Repository'
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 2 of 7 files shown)


Wuodan:
The open PRs by dependabot cover most of these, see:
https://github.com/bats-core/bats-core/pulls?q=is%3Apr+state%3Aopen+author%3Adependabot%5Bbot%5D

Just my 2 cents :-)

martin-schulze-vireso:
Thanks for your suggestions. However, I prefer to update individually.

PR #1244, opened 2026-09-10
test: eliminate SIGINT-after-run race

Wuodan:
## Summary

Fix an intermittent Windows failure in `CTRL-C aborts and fails after run`.

The fixture signalled readiness before `run true` completed. The parent could
therefore send SIGINT before the intended post-`run` phase, resulting in a
different TAP result. Signal readiness after `run true` so the interrupt always
targets the following sleep.

## Validation

- `bin/bats --filter 'CTRL-C aborts and fails after run' test/bats.bats`

- [x] I have reviewed the [Contributor Guidelines][contributor].
- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it

[contributor]: https://github.com/bats-core/bats-core/blob/master/docs/CONTRIBUTING.md
[coc]:         https://github.com/bats-core/bats-core/blob/master/docs/CODE_OF_CONDUCT.md

PR #1245, opened 2026-09-10
test: allow parallel suite startup on loaded macOS runners

Wuodan:
## Summary

Fix an intermittent macOS timeout in `parallel suite execution with --jobs`.

The test requires 12 Bash workers to reach a barrier but inherited the
10-second watchdog. On loaded runners, process startup can exceed that limit
and exhaust the retries. Give this test a local 120-second timeout; its TAP,
ordering, and concurrency assertions remain unchanged.

## Validation

- `bin/bats --filter 'parallel suite execution with --jobs' test/parallel.bats`

### DIFF
--- a/test/parallel.bats
+++ b/test/parallel.bats
@@ -70,6 +70,9 @@ check_parallel_tests() { # <expected maximum parallelity>
 @test "parallel suite execution with --jobs" {
   # shellcheck disable=SC2034
   BATS_TEST_RETRIES=2 # be more robust against flaky MacOS runners
+  # Starting 12 Bash processes can take longer than a minute on loaded macOS runners.
+  # shellcheck disable=SC2034
+  BATS_TEST_TIMEOUT=120
   # shellcheck disable=SC2031,SC2030
   export FILE_MARKER
   # shellcheck disable=SC2030


Wuodan:
Similar to #1244 this error also seemed to pop up in my fork when running tests under load (the full test suite). And a re-run of just failing jobs mostly worked then.

## Error Output

```
not ok 375 parallel suite execution with --jobs # timeout after 10s
# (from function `reentrant_run' in file test/test_helper.bash, line 77,
#  in test file test/parallel.bats, line 83)
#   `reentrant_run bash -c "bats --jobs $PARALLELITY \"${FIXTURE_ROOT}/suite/\" 2> >(grep -v '^parallel: Warning: ')"' failed due to timeout
# Last output:
# 1..12
```

Wuodan:
Update: 
Timeout increased from 60 to 120 seconds.

I still had the error sometimes with a 60 second timeout when the macOS test was running in the full test suite.

Wuodan:
I just converted this to draft status.

I have not seen this test failing on the "polished" main branch of my fork.
But I added this change to temporary branches to test PRs and there it has popped up.

So I'm not sure if this change alone fixes the issue - hard to tell with something that randomly appears.
Also this might depend on general load of GH infrastructure - I've seen it have really bad days where everything is slow and such a timeout can be affected. Plus macOS runners are a bit exotic/special there anyways.

martin-schulze-vireso:
If it helps with pinning this down, we could strip off all other OSes and tests, and only run many parallel macos tests and retry it several times. This should at least help with measuring the defect frequency. If it is in a noticable range, we should be able to see differences.

However, I am also inclined to make the test more deterministic. I am open for suggestions.

Wuodan:
Closed: Replaced by PR #1259, which has a solution for the underlying issue.

PR #1246, opened 2026-09-10
ci: use available locale in FreeBSD tests

Wuodan:
The Firecracker FreeBSD guest inherits LANG=C.UTF-8 from the Linux runner, but that locale is not installed in the guest. GNU parallel's Perl processes then print locale warnings to stdout, interleaving with TAP output and causing parallel-test assertions to fail.

Set LC_ALL=C before running the suite. The C locale is always available on FreeBSD, preventing the warnings and preserving the expected TAP output.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -259,6 +259,7 @@ jobs:
         with:
           run-in-vm: |
             pkg install -y bash parallel ${{ matrix.packages }}
+            export LC_ALL=C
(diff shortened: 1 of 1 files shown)


Wuodan:
This caused the `FreeBSD` tests to fail **sometimes**, which is worse than **always**.

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for this fix.

Issue #1036, opened 2024-12-10
Time to complete empty tests increases quadratically (~O(n^2)) as the number of tests grows


cherouvim:
I am on a fresh Ubuntu 24 VM and have installed `Bats 1.11.1` using:
```bash
git clone https://github.com/bats-core/bats-core.git /tmp/bats
sudo /tmp/bats/install.sh /usr/local
```

I have empty tests of this form with unique titles so bats will not complain:
```bash
#!/usr/bin/env bats

@test "test 00481268-6f00-460e-b5e9-ed31520f5869" { }
@test "test 00a062d7-473a-4549-988d-ac7585110c89" { }
@test "test 0168e5d7-231b-4a19-b652-364948971c7e" { }
@test "test 027d81cb-7e48-42d7-b7bb-9d8d0ebab2da" { }
@test "test 029ef1b9-fe4f-4081-8780-5552d1a5a662" { }
...
```

Running the above with just 1 test is fast. Doing so with 10 is fast as well. Doing so with 100 or 500 is slow, and doing so with 1000 takes ages. And not only does each test statement take longer, but script initialization (even before the first test runs) is very slow on 500 or 1000 tests. So, the total time to complete all tests does not seem to be linear:

1 test:
```
...
1 test, 0 failures


real    0m0.241s
user    0m0.199s
sys     0m0.050s
```

10 tests:
```
...
10 tests, 0 failures


real    0m0.719s
user    0m0.426s
sys     0m0.298s
```

100 tests:
```
...
100 tests, 0 failures


real    0m7.413s
user    0m5.834s
sys     0m1.531s
```

500 tests:
```
...
500 tests, 0 failures


real    1m12.522s
user    0m53.700s
sys     0m18.415s
```

1000 tests:
```
...
1000 tests, 0 failures


real    4m2.437s
user    3m4.033s
sys     0m57.090s
```

Thus bats is not usable at all for large tests. For extra context, I want to test around 5000 configured http redirects on a local apache server.

martin-schulze-vireso:
I assume the tests are all in one file? What happens when you split them into multiple files?

cherouvim:
@martin-schulze-vireso yes, I am talking about 100, 500, and 1000 tests in a single file. Splitting them to multiple files solves the issue.

martin-schulze-vireso:
Without having looked into the code, my guess would be the duplicate test check. We can probably improve it via sort and uniq.

martin-schulze-vireso:
[...]

PR #1249, opened 2026-09-10
Fix false duplicate detection for dynamic tests

### LINKED ISSUE #1235: `bats_test_function` false "Duplicate test name(s)" when one dynamic test's args are a prefix of another's
**Component:** `libexec/bats-core/bats-gather-tests`
**Version:** bats-core 1.13.0 (also present at `d9faff0`, i.e. `v1.13.0-33-gd9faff0`)
**Affected function:** `bats_test_function` (dedup check added at gather time)

## Summary

When a test file registers dynamically-parametrized tests via `bats_test_function`
(the pattern used for data-driven tests, e.g. one function run with different
argument tuples), bats aborts the whole file with a false
`ERROR: Duplicate test name(s) in file "...": <name>` whenever one registered
call's `function name + args` string happens to be a literal **prefix** of another
call's string for the *same* function. This is a false positive: the two calls have
different arguments and are not duplicates. Whether the bug triggers depends on the
**registration order** of the two calls within the file, which for callers building
the argument list from an associative array (e.g. iterating `"${!some_assoc_array[@]}"`
to call `bats_test_function` in a loop) is effectively bash hash-order-dependent and
not under the test author's control.

## Root cause

In `bats_test_function` (`libexec/bats-core/bats-gather-tests`):

```bash
line="$BATS_TEST_FILENAME"
line+=$'\t'
line+="$*"
...
# dereferencing ${test_names[*]} fails on older bash versions when empty -> check before
if [[ ${#test_names[@]} -gt 0 && " ${test_names[*]} " == *" $line "* ]]; then
    test_dupes+=("$line")
fi
test_names+=("$line")
```

Each test's identity is `<filename>\t<function-name-and-args>`. Duplicate detection
joins all previously-seen identities with spaces and does a **substring** glob
match (`*" $line "*`) instead of an exact per-element comparison. Because `IFS`
(space) also appears *inside* `$*` whenever a call has multiple arguments, a
shorter identity string can end up embedded, padded by spaces, inside a longer one
— even though the two are genuinely different `line` values with different
`$*` args.

Concretely: registering `f --format plain` and then `f` (bare, no args) produces
these two `line` values (tab-separated filename omitted for brevity):

- `f --format plain`
- `f`

Joining as `" f --format plain f "` and searching for `*" f "*` matches, because
`f ` appears right before `--format` — the search string `" f "` is a substring of
`" f --format plain f "` purely by coincidence of word boundaries, not because `f`
(no args) was ever actually registered before.

## Reproduction

```bash
# prefix.bats — longer args registered FIRST, shorter (bare) SECOND
f() { :; }
bats_test_function --description "long one" -- f --format plain
bats_test_function --description "short one" -- f
```

```
$ bats prefix.bats
ERROR: Duplicate test name(s) in file "prefix.bats": f
```

Swapping the registration order makes the false positive disappear, confirming the
substring/order dependency and that this is not a real duplicate:

```bash
# prefix-reversed.bats — same two calls, shorter (bare) registered FIRST
f() { :; }
bats_test_function --description "short one" -- f
bats_test_function --description "long one" -- f --format plain
```

```
$ bats prefix-reversed.bats
1..2
ok 1 short one
ok 2 long one
```

Both files declare the exact same two logical tests; only the registration order
differs, and only one order is (falsely) rejected.

## Impact

Any file that dynamically parametrizes tests over a set of CLI-flag-shaped argument
strings — e.g. building test cases from an associative array and calling
`bats_test_function` in a `for description in "${!cases[@]}"` loop — can hit this
whenever one case's argument string happens to be a prefix of another's for the same
underlying function (a very natural situation: a "default"/bare case next to a case
that adds one more flag on top). Because the loop iterates over hash-table keys,
whether the bug fires can flip between runs/bash versions with no change to the test
file itself, making it read as flaky/nondeterministic when it is really a
registration-order artifact of a substring-match dedup check.

## Related: bats-core#1036

[bats-core#1036](https://github.com/bats-core/bats-core/issues/1036) (open,
confirmed) reports that gathering empty tests scales ~O(n²) with test count (0.24s
at 1 test, 242s at 1000). That's this same dedup check: on every
`bats_test_function` call it rebuilds `" ${test_names[*]} "` (already O(n) in the
tests seen so far) and glob-scans it, so the file-level cost is O(n²) even before
considering correctness. #1036's repro uses unique per-test titles, so it isn't
hitting the prefix false-positive described here — it's the same root design (an
ever-growing collection re-scanned from scratch on every call) surfacing as a
performance bug instead of a correctness one. Worth linking the two when filing.

## Suggested fix

Compare identities exactly instead of via substring/glob match. An associative
array would be the obvious tool, but this file explicitly supports Bash 3 (see the
"avoid unbound variable with set -u Bash 3" comments a few lines above the bug, at
lines 71 and 88) and `declare -A` is Bash-4+ only, so the fix needs to stay
Bash-3-compatible.

**Minimal fix (correctness only):** a plain exact-match scan over the existing
array, no new data structure:

```bash
local is_dupe=false existing
for existing in "${test_names[@]}"; do
    if [[ "$existing" == "$line" ]]; then
        is_dupe=true
        break
    fi
done
"$is_dupe" && test_dupes+=("$line")
test_names+=("$line")
```

This preserves the existing behaviour for genuine duplicates (identical filename +
function + args registered twice) while eliminating false positives from prefix
relationships between distinct argument strings, and removes the order-dependence —
all without raising the Bash version floor. It's still O(n²) per file overall
(one O(n) scan per call), so on its own it does **not** fix #1036 — merely stops it
from also being wrong.

**Fix that also addresses #1036:** move duplicate detection out of the per-call hot
path entirely. Keep `test_names+=("$line")` as the only per-call work, then once at
the end of gathering (e.g. in `bats_gather_tests`, after the file has been fully
sourced), sort the collected array once and scan for adjacent equal entries:

```bash
local sorted_names
IFS=$'\n' sorted_names=($(sort <<<"${test_names[*]}")) # or `sort -z` with a NUL-joined build if names may contain newlines
unset IFS
local i
for ((i = 1; i < ${#sorted_names[@]}; i++)); do
    [[ "${sorted_names[i]}" == "${sorted_names[i-1]}" ]] && test_dupes+=("${sorted_names[i]}")
done
```

This is O(n log n) per file (dominated by `sort`), fixes the same false-positive
(exact-value comparison, no substring search) and directly addresses #1036's
complaint, at the cost of a slightly larger change (duplicate detection moves from
"as you register" to "after the file is fully gathered").

fzlzjerry:
- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it.

## Summary

Confirm duplicate candidates against complete array elements instead of treating a substring match as a duplicate. Keep the cheap negative check for space-joined arrays, and use the exact scan directly for other IFS separators.

This preserves existing test identities, duplicate messages, registration order and status-log format, without adding a subprocess per registration or requiring Bash 4.

## Validation

- Seven regression tests: four fail on the original gatherer; all seven pass after the fix on Bash 5.2.37 and Bash 3.2.57.
- Full Bash 5.2 suite: 494 passed, 2 existing skips, no failures.
- Full Bash 3.2 suite: 493 passed, 3 existing/platform skips, no failures. The original suite has the same skips.
- Coverage includes both orders, multi-word/glob/empty arguments, actual argument execution, genuine duplicates, IFS/nounset cases, separate files and filtering.
- Whole-project ShellCheck passes; changed regions match shfmt 3.5.1. Existing repository-wide formatting differences remain outside this patch.
- Paired 1,000-registration count checks had local medians of 6.08 s before and 6.05 s after. This is a correctness fix, not a fix for the scaling issue in #1036.

Fixes #1235.


test/bats.bats
```diff
@@ -1530,6 +1530,104 @@ END_OF_ERR_MSG
   [ "${#lines[*]}" -eq 17 ]
 }
 
+@test "dynamic argument prefixes are distinct in either registration order" {
+  local order expected_log
+  local log="$BATS_TEST_TMPDIR/registration.log"
+  for order in long-first short-first; do
+    : >"$log"
```
Wuodan:
```suggestion
    : >|"$log"
```

test/bats.bats
```diff
@@ -1530,6 +1530,104 @@ END_OF_ERR_MSG
...
+    fi
+    [ "$(<"$log")" == "$expected_log" ]
+  done
+}
+
+@test "dynamic argument prefixes preserve spaces glob characters and empty values" {
+  local registration_case expected_log
+  local log="$BATS_TEST_TMPDIR/registration.log"
+  for registration_case in multi-word glob empty; do
+    : >"$log"
```
Wuodan:
```suggestion
    : >|"$log"
```

test/bats.bats
```diff
@@ -1530,6 +1530,104 @@ END_OF_ERR_MSG
...
+    [[ ${lines[0]} == 'ERROR: Duplicate test name(s) in file '* ]]
+    [ ! -s "$log" ]
+  done
+}
+
+@test "dynamic argument prefixes work with nounset and unset IFS" {
+  local registration_case
+  local log="$BATS_TEST_TMPDIR/registration.log"
+  for registration_case in nounset unset-ifs; do
+    : >"$log"
```
Wuodan:
```suggestion
    : >|"$log"
```

Wuodan (commented):
I suggest changing `: >"$log"` to `: >|"$log"` in the added tests or they will fail with `set -uo noclobber` used by the test GitHub Actions `unset_variables` job.

To see for yourself run this in your branch and you will see `cannot overwrite existing file` in the 3 affected tests: 
```
BASH_ENV="$PWD/.github/workflows/set_nounset.bash" \
  bin/bats --filter 'dynamic argument prefixes (are|preserve|work)' test/bats.bats
```

**Reason:**
The newly added tests reuse the same registration.log across loop iterations. CI enables noclobber through BASH_ENV, so after the first iteration creates the file, >"$log" refuses to truncate it and the test aborts before the next iteration. Since overwriting the reused log is intentional, the three reset operations need to use >|"$log" to explicitly override noclobber. This is a test-harness issue; the nested Bats behavior being tested is not what fails.

test/bats.bats
fzlzjerry: Fixed in `d654694`: all three intentional log resets use `>|`. I reproduced the three `cannot overwrite existing file` failures on the original head with the CI BASH_ENV inherited by the test processes; all three pass after the fix, both with and without that environment. The complete test/bats.bats run under nounset/noclobber had 137 test observations, 5 skips, and no failures. The repository-pinned shfmt check passes; its whitespace normalization also touched two existing command substitutions.

test/bats.bats
fzlzjerry: Applied here as well in `d654694`. This multi-iteration test passes with the CI nounset/noclobber environment; the original redirection failed on the second iteration.

test/bats.bats
fzlzjerry: Applied here in `d654694`, retaining the nounset/unset-IFS coverage while explicitly allowing the reused log to be truncated. Verified with the CI BASH_ENV.

Wuodan:
@martin-schulze-vireso how comfortable are you with having `awk` as dependency?

---

The change in [c632ffb](https://github.com/bats-core/bats-core/commit/c632ffb8af50fa85c9eb143290395ee1fb93b32b) by @fzlzjerry does indeed fix #1235 .

I found alternative solutions which also address #1036 "O(n^2) time increase with many tests in one file".

Code for a hybrid solution (with check if awk is present) can be seen at [this commit](https://github.com/bats-hardened/bats-core/commit/a430b312e9cedca0546685027fd0e6c410616c96):

1. A nicely readable O(n) solution using `awk`
2. A not so readable O(n log n) solution with `sort`

Pros/Cons:

- `awk`:
  - Pros:  readable and fast code
  - Cons: Adds `awk` as dependency
- `sort`: 
  - Pros:  not having `awk` as dependency, `sort` is already used by `bats-core`
  - Cons: harder to read algorithm. Slightly slower with many tests in one file (+1s with 4k tests in a file)

The real drawback of the sort solution to me is readability of code.

---

My performance benchmark results for the `sort` vs. `awk` solutions are here:
- [performance-results.md](https://github.com/bats-hardened/bats-core/blob/PR-review/fork/fzlzjerry/fix/1235-exact-test-name-membership.benchmark/docs/PR-review/1249/performance-results.md)
- [benchmark-duplicate-detection.sh](https://github.com/bats-hardened/bats-core/blob/PR-review/fork/fzlzjerry/fix/1235-exact-test-name-membership.benchmark/docs/PR-review/1249/benchmark-duplicate-detection.sh): Many tests in one file, compares 2 commits
- [benchmark-many-files.sh](https://github.com/bats-hardened/bats-core/blob/PR-review/fork/fzlzjerry/fix/1235-exact-test-name-membership.benchmark/docs/PR-review/1249/benchmark-many-files.sh): Many files with few tests, compares 2 commits

martin-schulze-vireso:
To be honest I am not sure about `awk`. We don't have good data about the fleet of different systems running bats out there.
I would not avoid it at all cost but if we can do without additional dependencies, I would prefer not using `awk`.

Also, I must say I don't find the numbers very compelling for any optimization at all. 20s vs 10s is a nice speedup but I expect actually running 4k tests to dominate this. 

I would prefer to focus this PR on fixing the correctness issue and having the performance discussion in the other issue.

docs/CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ The format is based on [Keep a Changelog][kac] and this project adheres to
 
 ### Fixed
 
+* avoid false duplicate errors for dynamic tests with argument prefixes, and
+  detect duplicates with non-space `IFS` separators (#1235)
```
Wuodan, line 21:
```suggestion
  detect duplicates with non-space `IFS` separators (#1249)
```
The changelog must reference the PR, not the issue, please replace `1235` with `1249`.

Wuodan (changes requested):

Wuodan:
> I would prefer to focus this PR on fixing the correctness issue and having the performance discussion in the other issue.

You're right. I support this PR as is with the small change to the CHANGELOG.md suggested above.

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thank you for contributing.

PR #1250, opened 2026-09-12
fix: preserve stack traces for failed helper loads

### LINKED ISSUE #1186: If a `load` call fails immediately after a `load` call that is successful, the stack trace incorrectly indicates the first `load` failed
**Describe the bug**
If a `load` fails immediately after a `load` that was successful, the stack trace indicates the wrong line failed.

i.e.

```bash
❯ cat test/load-trace-failure-example.bats
#!/usr/bin/env bats

@test "load trace failure example" {
  load test_helper
  load this_does_not_exist
}

❯ ./bin/bats test/load-trace-failure-example.bats
load-trace-failure-example.bats
 ✗ load trace failure example
   (from function `source' in file test/test_helper.bash, line 58,
    from function `bats_internal_load' in file lib/bats-core/test_functions.bash, line 67,
    from function `bats_load_safe' in file lib/bats-core/test_functions.bash, line 98,
    from function `load' in file lib/bats-core/test_functions.bash, line 148,
    in test file test/load-trace-failure-example.bats, line 4)
     `load test_helper' failed
   bats_load_safe: Could not find '/Users/levi.noecker/workspace/bats-core/test/this_does_not_exist'[.bash]

1 test, 1 failure
```

You will note that one line indicates `load test_helper' failed`, while the line below is the actual output from the failed command, indicating the second `load` failed: 
`bats_load_safe: Could not find '/Users/levi.noecker/workspace/bats-core/test/this_does_not_exist'[.bash]`

**To Reproduce**

1. From the root of the bats-core repo, create a test `test/load-trace-failure-example.bats` with the following contents  
```bash
#!/usr/bin/env bats

@test "load trace failure example" {
  load test_helper
  load this_does_not_exist
}
```
2. run `bats test/load-trace-failure-example.bats`  

**Expected behavior**
I would expect the tracing output to correctly indicate the line that failed. In this case it indicates the previous line, which was successful, failed.

**Environment:**
```bash
❯ printf 'Bats Version: \n%s\n\nOS Information:\n%s\n\nBash Version:\n%s' "$(./bin/bats --version)" "$(uname -a)" "$(bash --version)"
Bats Version:
Bats 1.13.0

OS Information:
Darwin TC-4001192 25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:30:01 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6031 arm64

Bash Version:
GNU bash, version 5.3.9(1)-release (aarch64-apple-darwin25.1.0)
Copyright (C) 2025 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```
Installed via `git clone git@github.com:bats-core/bats-core.git` and running `./bin/bats` directly

**Additional context**
I spent some time trying to debug this, and got to this section of [lib/bats-core/tracing.bash](https://github.com/bats-core/bats-core/blob/5f12b3172105a0f26bce629ff8ae0ac76c4bd61e/lib/bats-core/tracing.bash#L28-L42):

```bash
bats_get_failure_stack_trace() {
  local stack_trace_var
  # See bats_debug_trap for details.
  if [[ -n "${BATS_DEBUG_LAST_STACK_TRACE_IS_VALID:-}" ]]; then
    stack_trace_var=BATS_DEBUG_LAST_STACK_TRACE
  else
    stack_trace_var=BATS_DEBUG_LASTLAST_STACK_TRACE
  fi
  # shellcheck disable=SC2016
  eval "$(printf \
    '%s=(${%s[@]+"${%s[@]}"})' \
    "${1}" \
    "${stack_trace_var}" \
    "${stack_trace_var}")"
}
```

`BATS_DEBUG_LAST_STACK_TRACE` and `BATS_DEBUG_LASTLAST_STACK_TRACE` point to temporary files holding the trace information. The trace pointed to by `BATS_DEBUG_LAST_STACK_TRACE` does indicate the correct line that failed. The trace pointed to by `BATS_DEBUG_LASTLAST_STACK_TRACE` contains, as one might expect, the trace from immediately before the failure. After adding some debug print lines it was clear that the branch with `BATS_DEBUG_LASTLAST_STACK_TRACE` is the one followed when my example failure test runs.

I then tried to trace down how and why `BATS_DEBUG_LAST_STACK_TRACE_IS_VALID` gets set, which led me to the `bats_debug_trap()` function, also in [bin/bats-core/tracing.bash](https://github.com/bats-core/bats-core/blob/5f12b3172105a0f26bce629ff8ae0ac76c4bd61e/lib/bats-core/tracing.bash#L228-L300). From the big block of text above the function I gathered this could be an issue associated with the reliability of `$BASH_LINENO`, and decided this was about as far as I was comfortable exploring at this time.

aprylewu:
After `load test_helper` succeeds, a subsequent `load nonexistent` reports the successful first call as the failure. The same problem affects `bats_load_library`.

These helpers exit directly on failure, so the ERR trap does not run. Their internal calls are excluded from DEBUG tracing, leaving the current stack trace valid; falling back to the preceding trace points at the wrong command. Mark the current trace as valid before exiting. This also retains the helper's source location when sourcing returns a nonzero status.

Adds four regression cases for both loaders, covering missing helpers and nonzero returns after a successful load. Updates the suite-loading assertion to expect the retained helper frame on newer Bash versions.

Fixes #1186.

Validation on macOS arm64:
- Bash 3.2.57 and Bash 5.3.0: all four new regressions fail on the original code and pass with the fix.
- Full suite on each Bash version: 470 passed, 23 existing dependency/intentional skips.
- Full ShellCheck 0.11.0 check and `git diff --check` pass.
- shfmt 3.5.1 passes for the changed tests. The loader file has existing formatting differences; this change introduces none.

- [x] I have reviewed the [Code of Conduct][coc] and agree to abide by it.

AI assistance: OpenAI Codex.


Wuodan:
I recommend approving this PR. The implementation fixes the reported stack traces, and the updated Bash-version expectations match the behavior I observed. I found one small correction needed in the test suite.

## Suggested correction: respect `BATS_LIBDIR`

The repository's `.github/workflows/tests.yml` includes a `lib64-install` job that installs Bats with `./install.sh /usr/ lib64` and then runs the test suite using that installation. The new expectations in `test/suite.bats` hard-code `lib`, so this job fails: the produced trace is correctly rooted at `/usr/lib64`, while the assertion expects `/usr/lib`.

The three assertions should replace `${RELATIVE_BATS_ROOT}lib/` with `${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/`:

```diff
-  [ "${lines[3]}" = "#  from function \`bats_internal_load' in file ${RELATIVE_BATS_ROOT}lib/bats-core/test_functions.bash, line 67," ]
-  [ "${lines[4]}" = "#  from function \`bats_load_safe' in file ${RELATIVE_BATS_ROOT}lib/bats-core/test_functions.bash, line 98," ]
-  [ "${lines[5]}" = "#  from function \`load' in file ${RELATIVE_BATS_ROOT}lib/bats-core/test_functions.bash, line 150," ]
+  [ "${lines[3]}" = "#  from function \`bats_internal_load' in file ${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/bats-core/test_functions.bash, line 67," ]
+  [ "${lines[4]}" = "#  from function \`bats_load_safe' in file ${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/bats-core/test_functions.bash, line 98," ]
+  [ "${lines[5]}" = "#  from function \`load' in file ${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/bats-core/test_functions.bash, line 150," ]
```

With this correction, the source-tree tests pass 47/47 and the same suite passes 24/24 from a temporary `lib64` installation.

## What I checked

I reproduced #1186 by running the new fixture before and after the change:

```console
bin/bats test/fixtures/load/failing_load_after_success.bats
```

The fixture intentionally fails all four tests. This is the relevant output diff:

```diff
 not ok 1 missing helper after successful load
-# (in test file test/fixtures/load/failing_load_after_success.bats, line 2)
-#   `load test_helper' failed
+# (in test file test/fixtures/load/failing_load_after_success.bats, line 3)
+#   `load nonexistent' failed

 not ok 2 helper returning nonzero after successful load
-# (in test file test/fixtures/load/failing_load_after_success.bats, line 8)
+# (from function `source' in file test/fixtures/load/return1.bash, line 1,
+#  from function `bats_internal_load' in file lib/bats-core/test_functions.bash, line 67,
+#  from function `bats_load_safe' in file lib/bats-core/test_functions.bash, line 98,
+#  from function `load' in file lib/bats-core/test_functions.bash, line 150,
+#  in test file test/fixtures/load/failing_load_after_success.bats, line 8)
 #   `load return1' failed

 not ok 3 missing library after successful load
-# (in test file test/fixtures/load/failing_load_after_success.bats, line 12)
-#   `bats_load_library "$BATS_TEST_DIRNAME/test_helper.bash"' failed
+# (in test file test/fixtures/load/failing_load_after_success.bats, line 13)
+#   `bats_load_library "$BATS_TEST_DIRNAME/nonexistent.bash"' failed

 not ok 4 library returning nonzero after successful load
-# (in test file test/fixtures/load/failing_load_after_success.bats, line 18)
+# (from function `source' in file test/fixtures/load/return1.bash, line 1,
+#  from function `bats_internal_load' in file lib/bats-core/test_functions.bash, line 67,
+#  from function `bats_load_library_safe' in file lib/bats-core/test_functions.bash, line 135,
+#  from function `bats_load_library' in file lib/bats-core/test_functions.bash, line 141,
+#  in test file test/fixtures/load/failing_load_after_success.bats, line 18)
 #   `bats_load_library "$BATS_TEST_DIRNAME/return1.bash"' failed
```

This fixes both problems I could reproduce: missing loads no longer blame the preceding successful load, and a helper returning nonzero retains the source location where the status originated.

## `test/suite.bats` changes

I initially wasn't sure why this part of the test changed so much, so I ran the old and new code under Bash 3.2.57, 4.0.44, and 4.1.17:

| Bash | Before | After |
|---|---|---|
| 3.2 | Full helper trace; status 127 | Same trace and status |
| 4.0 | Full helper trace | Same trace |
| 4.1+ | Only the outer `load` frame | Full helper trace |

The old conditional was recording a real Bash-version difference. The PR makes Bash 4.1+ produce the same full trace as 3.2 and 4.0, so moving the common assertions out of the conditional is correct. The remaining branch covers the Bash 3 status-127 exit versus Bash 4+'s `Error while sourcing library loader`.

I also traced why the two assignments are sufficient:

- The loader handles the failure under `if !` and then calls `exit`, so no `ERR` trap marks the latest trace as valid.
- `bats_get_failure_stack_trace` therefore falls back to `BATS_DEBUG_LASTLAST_STACK_TRACE`.
- Bats' internal paths do not update the DEBUG trace, so explicitly selecting `BATS_DEBUG_LAST_STACK_TRACE` at this controlled exit preserves the correct user-code trace.

test/suite.bats
```diff
@@ -59,31 +59,21 @@ setup() {
...
-  # bash < 4.0 does not handle the status on source, it fails through the ERREXIT instead, which creates another trace
-  # bash == 4.0 seems to be sonwhere in between
-  if (( BASH_VERSINFO[0] > 4 )) || (( BASH_VERSINFO[0] == 4 && BASH_VERSINFO[1] > 0 )); then
-    [ "${lines[2]}" = "# (in test file $RELATIVE_FIXTURE_ROOT/errors_in_multiple_load/a.bats, line 1)" ]
-    [ "${lines[3]}" = "#   \`load test_helper' failed" ]
-    [ "${lines[4]}" = "# $FIXTURE_ROOT/errors_in_multiple_load/test_helper.bash: line 1: call-to-undefined-command: command not found" ]
-    [ "${lines[5]}" = "# Error while sourcing library loader at '$FIXTURE_ROOT/errors_in_multiple_load/test_helper.bash'" ]
-    [ "${#lines[@]}" -eq 6 ]
+  [ "${lines[2]}" = "# (in file $RELATIVE_FIXTURE_ROOT/errors_in_multiple_load/test_helper.bash, line 1," ]
+  [ "${lines[3]}" = "#  from function \`bats_internal_load' in file ${RELATIVE_BATS_ROOT}lib/bats-core/test_functions.bash, line 67," ]
```
Wuodan:
```suggestion
  [ "${lines[3]}" = "#  from function \`bats_internal_load' in file ${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/bats-core/test_functions.bash, line 67," ]
```

test/suite.bats
```diff
@@ -59,31 +59,21 @@ setup() {
...
+  [ "${lines[4]}" = "#  from function \`bats_load_safe' in file ${RELATIVE_BATS_ROOT}lib/bats-core/test_functions.bash, line 98," ]
```
Wuodan:
```suggestion
  [ "${lines[4]}" = "#  from function \`bats_load_safe' in file ${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/bats-core/test_functions.bash, line 98," ]
```

test/suite.bats
```diff
@@ -59,31 +59,21 @@ setup() {
...
+  [ "${lines[5]}" = "#  from function \`load' in file ${RELATIVE_BATS_ROOT}lib/bats-core/test_functions.bash, line 150," ]
```
Wuodan:
```suggestion
  [ "${lines[5]}" = "#  from function \`load' in file ${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/bats-core/test_functions.bash, line 150," ]
```

Wuodan (commented):
Suggested changes

test/suite.bats
aprylewu: Changed to `${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/` in 8714023.

test/suite.bats
aprylewu: Applied in 8714023.

test/suite.bats
aprylewu: Applied in 8714023; all three expectations now use `BATS_LIBDIR`.

aprylewu:
Thanks for catching this and for the detailed review. Updated all three assertions in 8714023 to use `${RELATIVE_BATS_ROOT}${BATS_LIBDIR}/`. I reproduced the failure with a temporary `lib64` installation; after the change, the source-tree load/suite tests pass (47/47), and the installed suite passes (24/24). The affected test also passes under Bash 3.2 in both layouts.

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks again for another fix contribution. :)

PR #1251, opened 2026-09-12
Support -- as the CLI option terminator

### LINKED ISSUE #867: Make -- a valid command line option and follow the convention of it being the terminator for all options

This relates to #863. I encountered it when working around the issue of dotfiles not being shown by setting `GLOBIGNORE=".:.."`, then invoking `bats *.bats`.

ShellCheck rightly pointed out that confusion could occur between command line options and files starting with `--` when using a bare glob.

The convention in most GNU utils these days is to allow `--` and assume that this terminates the list of options. Everything past that point is treated as a list of files/directories/items ...

I would propose to allow for `--` and have it signal the end of the list of options.

Currently I am simply using `./*.bats`, which also dodges the issue.

aprylewu:
Fixes #867.

`bats -- -example.bats` currently rejects `--`, and bundled-option preprocessing can also split dash-prefixed filenames. This change stops option parsing at `--` and keeps the remaining arguments as test paths. Short options are unpacked as they are parsed, preserving option arguments such as `-f --`. Path expansion also accepts parent directories whose names start with `-`.

The regression cases cover literal and option-like filenames, multiple paths, bundled options, recursive directories, nested paths, a missing test argument, and `--` used as both an option value and the terminator. CLI help, usage documentation, and the changelog describe the new syntax.

Validation:
- Full suite on macOS with Bash 3.2.57 and Bash 5.3.0: 475 passed, 23 skipped on each.
- The 24 `--`-related cases also completed under the CI `nounset` configuration on both Bash versions: 23 passed, 1 skipped.
- ShellCheck passed. shfmt reports only differences already present on the base branch.

AI assistance was used for implementation and test development.


Wuodan:
Looks good to me — I support this change as it is.

### Testing

I ran the full GitHub Actions test workflow (different Bash versions, macOS and Windows runners). `./shellcheck.sh` also passed.

### Implementation

I considered a few alternative ways to structure the short-option unpacking, but none looked better than the current implementation. In detail:

- Keeping the old preprocessing pass unchanged would still unpack filenames after `--`, defeating the purpose of the terminator.
- Teaching that preprocessing pass to stop at the first `--` looks simpler, but it cannot tell a terminator from an option value. For example, in `bats -f -- -ct test.bats`, the first `--` belongs to `-f`, and `-ct` still needs to be parsed as bundled options.
- Handling bundles in another `case` branch is possible, but introduces subtle pattern-order and control-flow requirements and makes the already substantial option switch harder to follow.

Expanding only the argument currently being parsed avoids those problems. The current placement immediately before the `case` looks like the clearest and least error-prone option: normalize one argument, then dispatch it.

### Why this is useful

`--` is the standard way to separate options from operands, and it fixes some real edge cases for Bats users:

- A glob such as `bats -- *.bats` can safely include a file named `-regression.bats`.
- Option-looking filenames can be run directly, for example `bats -- --help` or `bats -- -tr`.
- Dash-prefixed directories work too, including `bats -r -- -nightly-tests`.
- Wrapper scripts can safely use `bats -- "$test_path"` without first checking whether a user-provided or generated path starts with `-`.

The implementation, documentation, and regression coverage all look solid. Thanks for the contribution!

aprylewu:
Thanks for the thorough review and for walking through why this `--` expansion approach fits. Really appreciate the support for merging.

Wuodan (approved):

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for you contribution!

PR #1256, opened 2026-09-17
bats-format-tap - show test number out of total as a prefix

yogch:


martin-schulze-vireso:
No, the tap format is standardized and intended for programmatic ingestion.

 The pretty formatter offers a running count in interactive mode.

Please describe the problem you want to solve.

yogch:
I want to know the current status - how many tests ran out of X.
Without the need to scroll up and search for how many tests should run.
That's just easier and gives a better status than having only the test number.

I'm aware about the pretty formatter and its status line, but I'm running tests also with `CI=1`.

I think having additional number, for example: `[4/10]`, or `4/10`, will still work with "programmatic ingestion".

martin-schulze-vireso:
You could put your patched version of the tap formatter into your repo and use it via `--formatter /absolute/path/to/your/patched/version`.

yogch:
@martin-schulze-vireso yes I know. I actually did that.
But besides my own needs I thought it would be best to have in the real code. there's no harm.
Just more info regarding the status of the tests.

Also, if you change the actual 'tap' formatter I need to be aware and follow along...

martin-schulze-vireso:
> I think having additional number, for example: `[4/10]`, or `4/10`, will still work with "programmatic ingestion".

I don't think it is impossible to parse the additional number but I think this does not conform to the TAP specification, so it is not compatible with tools that expect TAP and might be used by the diverse userbase of Bats. 


> ... I thought it would be best to have in the real code. there's no harm. Just more info regarding the status of the tests.

We could add it behind a flag but I don't feel like the additional code is worth it. There already is a workaround and the TAP formatter is not the best target.

I could be persuaded to put something like this into the pretty formatter in non-interactive/CI mode though.
(I would also be keen to hear from @brokenpip3 and @Wuodan on this.)

brokenpip3:
Agree on keeping TAP standard the total and test numbers are already in the output so parsers already have what they need

yogch:
ok, I'm closing.

PR #1259, opened 2026-09-18
parallel: claim semaphore slots atomically

Wuodan:
## Problem

Parallel test jobs fail intermittently on macOS, while rerunning the same job usually succeeds. These failures occur frequently enough to make macOS CI unreliable during development.

Semaphore slots are currently represented by files and claimed through separate operations:

1. Count the occupied slots.
2. Find a slot whose file does not exist.
3. Create the file with `touch`.

This is not an atomic claim. Another process could select the same slot between the existence check and `touch`, and `touch` succeeds whether it creates the file or finds it already present. The complete sequence therefore has to be serialized with a global `flock` or `shlock`.

On macOS, the `shlock` fallback retries lock contention at one-second intervals. In observed failures, starting 12 workers took 7–9 seconds, leaving only 1–3 seconds for the barrier, result collection, and teardown before the enclosing 10-second test timeout.

## Solution

Represent each semaphore slot with a directory and claim it directly using `mkdir`.

`mkdir` combines the existence check and creation: it succeeds for the process that creates the slot and fails when the slot already exists. POSIX requires directory operations to be atomic and serializable (see [4.4 Directory Operations][posix-dir]), so competing workers cannot both create the same slot directory.

Slots are released with `rmdir`. This removes the separate slot counting and existence checks, the global lock, polling for lock contention, and the runtime dependency on `flock` or `shlock`.

The remaining one-second wait occurs only when all semaphore slots are occupied.

## Tests

- Keep the existing parallel and concurrency-limit coverage enabled without requiring `flock` or `shlock`.
- Remove the obsolete test expecting parallel execution to fail when neither command is installed.

[posix-dir]: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap04.html


Wuodan:
Following the [suggestion on #1245](https://github.com/bats-core/bats-core/pull/1245#issuecomment-5688979635), I ran a focused macOS stress test for the `mkdir`-based semaphore implementation in this PR.

The workflow was reduced to macOS 14 and macOS 15 with `rush`, repeatedly running only `test/parallel.bats` for five hours. Each iteration had its own 10-minute timeout, and the loop recorded failures and continued instead of stopping at the first one.

[The complete stress-test run](https://github.com/bats-hardened/bats-core/actions/runs/35305798266) finished successfully:

- [macOS 14](https://github.com/bats-hardened/bats-core/actions/runs/35305798266/job/105477620660): 304 iterations, 304 succeeded, 0 failed, 0 timed out
- [macOS 15](https://github.com/bats-hardened/bats-core/actions/runs/35305798266/job/105477620564): 251 iterations, 251 succeeded, 0 failed, 0 timed out
- Total: 555 complete executions of `test/parallel.bats` without a failure or timeout

This exercises the previously unreliable parallel path repeatedly under the affected macOS configurations and provides evidence that atomic slot claiming removes the intermittent failures.

martin-schulze-vireso:
We should also run this stress test with the old code to verify it can reproduce the issue. I also remember having tried mkdir and rejecting it. However, the only reason I can find now is that mkdir might not be atomic if used on a networked directory.

Wuodan:
> We should also run this stress test with the old code to verify it can reproduce the issue.

Coming, I just forgot to run it overnight.

---

> mkdir might not be atomic if used on a networked directory.

I looked into the network-filesystem case a bit more.

For network filesystems that provide normal POSIX-style namespace semantics, such as NFS, `mkdir`/`rmdir` are still server-side modifying operations. The client does not decide the result from a potentially stale directory cache. Concurrent `mkdir` calls for the same path are serialized by the server, so one succeeds and the other gets `EEXIST`.

There are also mounted "network filesystems" that do not really provide POSIX semantics, for example FUSE adapters over object storage or similar backends. In those cases, `mkdir` may only be an emulated operation and its atomicity depends on that particular implementation.

But I don't think that makes the proposed approach worse than the current one. The current implementation relies on more filesystem semantics: cross-client `flock`/`shlock`, directory reads/counting, existence checks, and `touch`, with correctness depending on the lock making that whole sequence atomic. On a backend that does not provide reliable POSIX-style semantics, those assumptions are not generally safer either.

Also, in normal use these semaphore directories are created below `/tmp`. `/tmp` is normally backed by a local filesystem or `tmpfs`; putting it on a network filesystem is possible, but unusual. Pointing the internal test-only temp-directory override at some exotic non-POSIX mount is of course also possible, but that is an even less typical case.

So my current understanding is:

* on normal local `/tmp` filesystems, `mkdir`/`rmdir` provide the atomic primitive we need;
* on normal POSIX-style network filesystems, `mkdir`/`rmdir` still provide the atomic primitive we need;
* on non-POSIX filesystem adapters, neither implementation can generally be assumed to provide correct cross-client locking, and the current implementation depends on at least as many filesystem guarantees as the `mkdir` approach.

Wuodan:
[The stress-test run with current locking solution](https://github.com/bats-hardened/bats-core/actions/runs/35451774995) had 7 failures on macOS 15, none on macOS14:

- [macOS 14](https://github.com/bats-hardened/bats-core/actions/runs/35451774995/job/105920074214): 298 iterations, 298 succeeded, 0 failed, 0 timed out
- [macOS 15](https://github.com/bats-hardened/bats-core/actions/runs/35305798266/job/105477620564): 227 iterations: 220 succeeded, 7 failed, 2 timed out

Only 7 failures of 227 tests on macOS15 is less than I experience.
But that maybe due to the test setup of letting 2 jobs run for 5 hours and not starting new job VMs for each iteration.

The failures all printed `kill: (12345) - No such process` before the `not ok` line.
6 failures were: `not ok 3 parallel suite execution with --jobs # timeout after 10s`,
1 failure was:  `not ok 6 parallelity factor is met exactly # timeout after 10s`

martin-schulze-vireso:
I am not sure if the network filesystem was the only reason but I can't remember anymore, only that I discarded `mkdir` for some reason.
I'd say let us try. I don't expect it to break catastrophically (like giving false passing tests), only to degrade performance if it does not work as intended.
I also like the code becomes much simpler.

martin-schulze-vireso (approved):

martin-schulze-vireso:
Thanks for this fix.

